PR #30, opened 2020-06-03
added initial files for member profile service

bhatiam12:
Added initial files for team member profile service. We will modify the objects as we move along but wanted to do initial PR so Pramukh and I can split work.

### DIFF
--- a/server/src/main/java/com/objectcomputing/member/MemberProfile.java
+++ b/server/src/main/java/com/objectcomputing/member/MemberProfile.java
@@ -1,0 +1,68 @@
+package com.objectcomputing.member;
+
+import javax.persistence.Column;
+import javax.persistence.Entity;
+import javax.persistence.GeneratedValue;
+import javax.persistence.GenerationType;
+import javax.persistence.Id;
+import javax.persistence.Table;
+import javax.validation.constraints.NotNull;
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+
+import java.sql.Date;
+
+
+@Entity
+@Getter
+@Setter
+@NoArgsConstructor
+@Table(name ="member_profile")
+public class MemberProfile {
+
+    public MemberProfile() {}
+
+    public MemberProfile(String name,String role,Long pdlId,String location
+                        ,String workEmail, String insperityId, Date startDate,
+                        String bioText) {
+                        this.name=name;
+                        this.role=role;
+                        this.pdlId=pdlId;
+                        this.location=location;
+                        this.workEmail=workEmail;
+                        this.insperityId=insperityId;
+                        this.startDate=startDate;
+                        this.bioText=bioText;
+                        }
+
+    @Id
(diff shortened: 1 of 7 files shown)


server/src/main/java/com/objectcomputing/member/MemberProfile.java
```diff
@@ -0,0 +1,68 @@
...
+                        String bioText) {
+                        this.name=name;
+                        this.role=role;
+                        this.pdlId=pdlId;
+                        this.location=location;
+                        this.workEmail=workEmail;
+                        this.insperityId=insperityId;
+                        this.startDate=startDate;
+                        this.bioText=bioText;
+                        }
```
mkimberlin, line 37: You have `@NoArgsConstructor`, but you have implemented that constructor...so it's not going to do anything.  If you want to use Lombok, I would kill line 24-37 and add `@AllArgsConstructor` to the class.

server/src/main/java/com/objectcomputing/member/MemberProfile.java
```diff
@@ -0,0 +1,68 @@
...
+
+    @Id
+    @GeneratedValue(strategy = GenerationType.AUTO)
+    private Long id;
```
mkimberlin, line 41: Might I suggest using a UUID?
```
    @Id
    @Column(name="uuid")
    @AutoPopulated
    private UUID uuid;
```
Then you just define it as a `varchar` in the database.

mkimberlin: By the way, `@AutoPopulated` is a Micronaut annotation: `import io.micronaut.data.annotation.AutoPopulated;`

server/src/main/java/com/objectcomputing/member/MemberProfile.java
bhatiam12: So i didn't remove it because lombok is actually not working. I was playing with dependencies and still can't make it to work so thought of checking the constructor in case i can't find a solution.

server/src/main/java/com/objectcomputing/member/MemberProfile.java
bhatiam12: Sure, I will change it to UUID.

mkimberlin:
With regard to the initial SQL file, you can keep it simple for now...something like:
https://github.com/objectcomputing/PulseSurvey/blob/develop/server/src/main/resources/db/2019.12.18.14.06.00_createInitialDb.sql

server/src/main/java/com/objectcomputing/member/MemberProfile.java
mkimberlin: In that case, I'd kill lombok completely and add your getters and setters.  We can try to revive lombok later.

bhatiam12: @mkimberlin If you are ok then can we merge this PR ? Pramukh can then pull these changes and do a PR to remove lombok and add controller logic.

PR #34, opened 2020-06-10
added files for check ins service and implement the post method

bhatiam12:
Adding initial files for check in service and implemented post method.


server/src/main/java/com/objectcomputing/checkins/CheckIns.java
```diff
@@ -0,0 +1,83 @@
...
+    public CheckIns(String teamMember, UUID pdlId, Date checkInDate, int targetQtr, int targetYear) {
+        this.teamMember= teamMember;
+        this.pdlId=pdlId;
+        this.checkInDate=checkInDate;
+        this.targetQtr=targetQtr;
+        this.targetYear=targetYear;
+    }
+    
+    @Column(name="teamMember")
+    private String teamMember;
```
bhatiam12, line 27: @mkimberlin remind me what this was for ? I couldn't remember so just added as a String.

server/src/main/java/com/objectcomputing/CheckInsRepository.java
```diff
@@ -0,0 +1,18 @@
...
+import java.util.UUID;
+
+import com.objectcomputing.checkins.CheckIns;
+
+public interface CheckInsRepository {
+
+    List<CheckIns> findByName(String teamMember);
+    List<CheckIns> findByTargetQtr(String targetQtr);
+    List<CheckIns> findByPdlId(UUID pdlId);
+    List<CheckIns> findByTargetYear(String targetYear);
```
mkimberlin, line 13: I'm think the target year and quarter have to be one call.  Perhaps this is just targetQuarter and that value should include year (i.e. "2020Q1")?

server/src/main/java/com/objectcomputing/checkins/CheckIns.java
mkimberlin: This is the reference to the team member ID. Similar to PDL ID below.  Team Member ID + PDL ID + Date = A Check-in

server/src/main/java/com/objectcomputing/CheckInsRepositoryImpl.java
```diff
@@ -0,0 +1,49 @@
+package com.objectcomputing;
```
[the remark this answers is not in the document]
bhatiam12: Makes sense. Talked to @Bagurp and we will change member profile as well to use CRUDRepository and then cleanup the code and so some refactoring. Will update the PR tomorrow after working with @Bagurp.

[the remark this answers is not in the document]
mkimberlin: PR #42 has the CRUDRepository code in it for the member profile.

PR #42, opened 2020-06-11
Fix email sending and some other refactorings

mkimberlin:


mkimberlin:
I will be the first to admit that this code doesn't stand up to my normal standards (needs tests and the interface for the EmailSender is uber simplistic). Just wanted to get this out to the group so that I didn't become a blocker.

bhatiam12:
I was thinking to move all controllers into /controllers folder, POJOs to /domain and Repository to /repository folder. @mkimberlin Let me know if you agree with this, if yes then I will do that refactoring with my checkin PR after this is merged.

mkimberlin:
@bhatiam12 We need to do something like that.  However, I wonder if we would be better off to slice those vertically instead and do something like 
```
com.objectcomputing.checkins.services.memberprofile
com.objectcomputing.checkins.services.checkins
...
```
And so on for each new REST service. That will position us to slice them into microservice deployments if we decide to. Thoughts?

bhatiam12:
If we decide to go microservice approach then yes, it makes sense to slice them vertically. Will refactor once this is merged!

PR #44, opened 2020-06-17
update micronaut version and small changes to checkin service

bhatiam12:
@mkimberlin I had to update the micronaut version in order to get past the compile issue after adding @Repository.
I also made some changes that I remember you talked about in the morning but I may have missed some so please feel to add or let me know and I will add it.


server/src/main/java/com/objectcomputing/checkins/services/checkins/CheckIn.java
```diff
@@ -32,7 +31,6 @@ public CheckIns(UUID teamMemberId, UUID pdlId, Date checkInDate, String targetQt
     @Id
     @Column(name="uuid")
     @NotNull
-    @GeneratedValue    
     private UUID pdlId;
```
mkimberlin, line 34: This column name should be "pdlId" and it shouldn't be the @Id.  This class needs a...
```
@Id
@Column(name="uuid")
@NotNull
@GeneratedValue
private UUID id;
```
...as well.

PR #46, opened 2020-06-18
Issue 17 unit tests

Bagurp:
* Adds Unit tests for MemberProfileController
* Upgrades Micronaut version to 2.0.0.M1
* Adds dependencies for Unit testing


server/build.gradle
```diff
@@ -54,17 +56,21 @@ dependencies {
     implementation "com.google.apis:google-api-services-gmail:v1-rev20200203-1.30.9"
     implementation "io.micronaut:micronaut-views"
     implementation "org.thymeleaf:thymeleaf:3.0.11.RELEASE"
+    implementation "io.micronaut.configuration:micronaut-hibernate-jpa"
+    implementation "io.micronaut.data:micronaut-data-jdbc:1.0.2"
```
mkimberlin, line 60: We should pull the µnaut data version into a gradle.properties file eventually so that these µnaut data module versions can be changed in one place. Not necessary to do in this in this PR.

server/src/main/java/com/objectcomputing/checkins/services/memberprofile/MemberProfile.java
```diff
@@ -28,9 +31,15 @@ public MemberProfile(String name, String role, UUID pdlId, String location,
...
-    @GeneratedValue
+    public MemberProfile() {
+	}
+
+	@Id
+    @Type(type = "uuid-char")
+    @Column(name="uuid", updatable = false, nullable = false, unique = true)
+    @GeneratedValue(generator = "UUID")
+    @GenericGenerator(name = "UUID", strategy = "org.hibernate.id.UUIDGenerator")
+    @ColumnDefault("random_uuid()")
```
mkimberlin, line 42: Interesting...does this impl work with both H2 and PostgreSQL?

server/src/main/java/com/objectcomputing/checkins/services/memberprofile/MemberProfile.java
```diff
@@ -28,9 +31,15 @@ public MemberProfile(String name, String role, UUID pdlId, String location,
                         this.bioText=bioText;
                         }
 
-    @Id
-    @Column(name="uuid")
-    @GeneratedValue
+    public MemberProfile() {
+	}
+
+	@Id
```
mkimberlin, line 37: Minor tabbing issue here.

server/src/main/java/com/objectcomputing/checkins/services/memberprofile/MemberProfileController.java
```diff
@@ -25,7 +24,7 @@ public MemberProfileController(MemberProfileRepository memberProfileRepository){
         this.memberProfileRepository = memberProfileRepository;
     }
 
-    @Get("/{?name,role,pdlId}")
+    @Get(uri = "/{?name,role,pdlId}", produces = MediaType.APPLICATION_JSON)
```
mkimberlin, line 27: You should be able to avoid repeating this on each method by annotating the class with `@Produces`, I believe:
https://docs.micronaut.io/latest/api/io/micronaut/http/annotation/Produces.html

server/src/test/java/com/objectcomputing/checkins/services/memberprofile/MemberProfileControllerTest.java
```diff
@@ -0,0 +1,274 @@
...
+        assertEquals(testUser, responseFindByRole.get(0).getName());
+        assertEquals(testRole, responseFindByRole.get(0).getRole());
+
+        // test Find By PdlId
+        HttpRequest requestFindByPdlId = HttpRequest.GET(String.format("/?pdlId=%s", testPdlId));
+        List<MemberProfile> responseFindByPdlId = client.toBlocking().retrieve(requestFindByPdlId, Argument.of(List.class, MemberProfile.class));  
+        assertEquals(1, responseFindByPdlId.size());
+        assertEquals(testUser, responseFindByPdlId.get(0).getName());
+        assertEquals(testRole, responseFindByPdlId.get(0).getRole());
+        assertEquals(testPdlId, responseFindByPdlId.get(0).getPdlId());
```
mkimberlin, line 163: I would tend to split these into separate tests.  If one fails at some point, it can be beneficial to know if only that case has regressed or multiple cases have regressed. Bundling then together will hide that, as the first failure will kill the test.

server/src/main/java/com/objectcomputing/checkins/services/memberprofile/MemberProfile.java
Bagurp: I am unsure since I don't have PostgreSQL setup on my local yet. Here's a blog that explains config with PostgreSQL - https://vladmihalcea.com/uuid-identifier-jpa-hibernate/

server/build.gradle
Bagurp: Agreed. I'll create a task in this sprint and make sure that it is changed.

PR #56, opened 2020-06-24
Ability to separate the build process for dev and prod

jorwalk:
I would like to separate the workflow process for the deployment of production and stage. Right now the script `gradle-master-develop.yml` explicitly declares the push commands on both branches. This is not ideal.
```
on:
  push:
    branches:
      - 'master'
      - 'develop'
```

Two additional files have been created to explicitly separate the build process for the two environments, with the addition of triggering a build on a pull request, not only the push.

```
on:
  push:
    branches: [develop]
  pull_request:
    branches: [develop]
``` 

- unblocks the Prod CI/CD issue: https://github.com/oci-labs/check-ins/issues/38
- unblocks the Dev CI/CD issue: https://github.com/oci-labs/check-ins/issues/29
- provides the ability to have separate credentials in Github secrets, when the `prod` environment is ready.


.github/workflows/gradle-build-feature.yml
```diff
@@ -0,0 +1,19 @@
+name: Gradle Build Feature Branch
+on:
+  push:
+    branches: [develop]
```
bhatiam12, line 4: this will run twice, correct ?I mean gradle-build-development workflow will also be triggered on push to develop and this one as well ?

.github/workflows/gradle-build-feature.yml
```diff
@@ -0,0 +1,19 @@
...
+    paths:
+      - 'feature/*'
```
mkimberlin, line 6: I'm confused by this...I think it belongs in branches, no?  This would trigger only if a file was modified in a directory called "feature".  I'm guessing that was not the intention.

https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestpaths

.github/workflows/gradle-build-feature.yml
jorwalk: Thank you for the review. It would run twice. My syntax was wrong. I have updated to be:
```
on:
  push:
    branches:
      - 'feature/**'
      - 'bug/**'
```

.github/workflows/gradle-build-feature.yml
jorwalk: I was confused also. Thank you for the link. I have updated the code to read:
```
on:
  push:
    branches:
      - 'feature/**'
      - 'bug/**'
```

.github/workflows/gradle-build-development.yml
```diff
@@ -0,0 +1,19 @@
+name: Gradle Build Production
```
mkimberlin, line 1: Crap...should have notice this before. The line above should say "Gradle Build Development". I only noticed it when I went into the actions to see how the build results report. :-/

mkimberlin: If you choose to change this in this PR, don't wait for additional approvals...just merge it.

bhatiam12: Good catch, i also missed it.

.github/workflows/gradle-build-feature.yml
```diff
@@ -0,0 +1,19 @@
+name: Gradle Build Feature Branch
```
hwilliams314, line 1: Should this (line 1) say something about bugs? Like Gradle Build Feature/Bug Branch?

Issue #19, opened 2020-05-20
Create RESTful service for recording check-ins


mkimberlin:
Record should initially hold:
team member,
pdl,
check-in date,
target quarter,
target year

All required fields

mkimberlin:
Base URL: /check-in

Create
Read
Find By:
- Team Member
- PDL
- Target Quarter
- Target Year
Update

PR #64, opened 2020-06-25
Issue #19 - Adds unit test for checkins service, adds checkins table in createInitialDb.sql

Bagurp:


server/src/main/resources/db/2020.06.03.16.00.00_createInitialDb.sql
```diff
@@ -3,12 +3,20 @@ CREATE TABLE memberProfile(
...
    insperityId varchar,
    startDate date,
    bioText varchar
 );
 
-
+drop table if exists checkins;
+CREATE TABLE checkins(
+   id varchar PRIMARY KEY,
+   teamMemberId varchar,
```
bhatiam12, line 17: are we not marking this as foreign key with this PR ?

Bagurp: Good catch. Updated PR.

PR #74, opened 2020-07-01
adding config for postgres running on gcp

bhatiam12:


bhatiam12:
@mkimberlin @jorwalk  I started this draft PR because I wanted to run past this approach I am planning to take to connect to postgres on GCP. 

Read a bit and went with using Google SQL socket factory so we don't need to deal with IP whitelisting. More info here : `https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory`

Also, Micronaut supports environment which we can use to run in gcp/local/test. I create two YAML files and they will be loaded based on the environments. For example: 
To run on local : ` MICRONAUT_ENVIRONMENTS=local ./gradlew run`
To run on GCP : ` MICRONAUT_ENVIRONMENTS=gcp ./gradlew run`

Lastly, if you see application-gcp.yaml then we are passing postgres credentials in the URL. i am thinking we can store these credentials on GITHUB secrets or if you guys have any other ideas to do this then please let me know.

bhatiam12:
If this approach is ok with you guys, then I will work with @jorwalk to get the credentials and try connecting from my machine and see everything works fine before storing credentials to GITHUB secrets.

jorwalk:
@bhatiam12 - makes sense to me

bhatiam12:
@mkimberlin  @jorwalk  Upon doing more research and then seeing peer-connection app which is in production and written using micronaut and using postgres, we can store credentials on google metadata . 
[`https://cloud.google.com/compute/docs/storing-retrieving-metadata`](url)

bhatiam12:
Deleting this PR because added commit as part of https://github.com/oci-labs/check-ins/pull/78

PR #78, opened 2020-07-02
Update gradle-build-development.yml

jorwalk:
Ability to build and deploy to Google Cloud Run.
- [x] Two secrets have been added to Github Secrets. `RUN_PROJECT`, `RUN_SA_KEY`


server/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM adoptopenjdk/openjdk11:latest
-COPY build/libs/*.jar check-ins-server.jar
+FROM openjdk:14-alpine
+COPY /build/libs/*-all.jar check-ins-server.jar
 EXPOSE 8080
-CMD java  -XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap -Dcom.sun.management.jmxremote -noverify ${JAVA_OPTS} -jar check-ins-server.jar
\ No newline at end of file
+CMD ["java", "-Dcom.sun.management.jmxremote", "-Xmx128m", "-jar", "check-ins-server.jar"]
```
mkimberlin, line 4: Would we be better off leaving the `${JAVA_OPTS}` instead of hardcoding `-Xmx128x`? That way the memory, etc. can be passed in via the environment...would prevent rebuilds for resource need changes, etc.

[See here.](https://docs.docker.com/engine/reference/commandline/run/#:~:text=Set%20environment%20variables%20(%2De,env%2C%20%2D%2Denv%2Dfile)&text=When%20running%20the%20command%2C%20the,be%20set%20in%20the%20container.)

jorwalk: You would know more than me, I have updated the CMD script.

bhatiam12:
@jorwalk do you know when you will be able to merge this ? I want to add postgres credentials to this workflow so once this is merged then i will do that because I see lot has changed in this workflow.

bhatiam12:
@jorwalk @mkimberlin I updated this PR and added environment variables. Also added yaml files to access those variables.

Issue #70, opened 2020-06-30
BE> Flyway 


sharmag04:
Definition of Done

- [ ] We have added flyway configuration to the project
- [ ] database migrations are handled by flyway

mkimberlin:
Fixed by PR #98

PR #98, opened 2020-07-10
Fixes #97 and #70 - fix gcp db connection and implement flyway

### LINKED ISSUE #97: Fix GCP DB connection
app won't start in cloud run due to some misconfiguration

mkimberlin:
Or at least it has a good chance to...so let's find out! :-)


mkimberlin:
Odd...not sure how that build error got in if it didn't happen on Jesse's PR. I'll look to fix...didn't touch that part of the app though. :-\

jorwalk:
```
Line 116:15:  React Hook useMemo does nothing when called with only one argument. Did you forget to pass an array of dependencies?  react-hooks/exhaustive-deps
```

server/src/main/resources/application.yml
```diff
@@ -64,4 +47,22 @@ micronaut:
       login-success-target-url: '/'
       unauthorized-target-url: '/unauthorized'
       forbidden-target-url: '/forbidden'
-      
\ No newline at end of file
+---      
+datasources:
+  default:
+    driver-class-name: org.postgresql.Driver
+    dialect: POSTGRES
```
bhatiam12, line 54: we don't need to pass username and password ?

server/src/main/resources/application-local.yml
```diff
@@ -1,66 +1,5 @@
...
-
-  server:
-    cors:
-      enabled: true
-    maxRequestSize: 100MB
-    multipart:
-      maxFileSize: 100MB
-
-  router:
-    static-resources:
```
bhatiam12, line 13: why did we delete all the config from application local ? is it for testing ?

server/src/main/resources/application.yml
mkimberlin: It should be picked up from the ENV, I believe. I may have to change the env var passed to DATASOURCES_DEFAULT_PASSWORD and DATASOURCES_DEFAULT_USERNAME, but I'll see on deploy.

server/src/main/resources/application-local.yml
mkimberlin: These files are cascading. You don't need to redefine all of the properties...only the ones that need to change. Check out the micronaut configuration documentation, it talks about order of precedence for properties.

server/src/main/resources/application.yml
mkimberlin: Just altered this...pretty positive it was going to cause an error.

server/src/main/resources/application-local.yml
bhatiam12: No, I meant swagger properties from local yaml. I don't see them defined anywhere.

bhatiam12: sorry, should have viewed the complete application.yaml and not just the changes :P

server/src/main/java/com/objectcomputing/checkins/gcp/postgres/GoogleCloudDatabaseSetup.java
```diff
@@ -0,0 +1,34 @@
...
+
+import io.micronaut.configuration.jdbc.hikari.DatasourceConfiguration;
+import io.micronaut.context.annotation.Property;
+import io.micronaut.context.annotation.Requires;
+import io.micronaut.context.env.Environment;
+import io.micronaut.context.event.BeanCreatedEvent;
+import io.micronaut.context.event.BeanCreatedEventListener;
+
+@Singleton
+@Requires(env = Environment.GOOGLE_COMPUTE)
```
mkimberlin, line 13: @Maroju100 Here's your answer to limiting the basic auth to tests: Environment.TEST

PR #101, opened 2020-07-13
Feature 58/skills rest service

hwilliams314:


server/src/main/java/com/objectcomputing/checkins/services/skills/SkillController.java
```diff
@@ -0,0 +1,106 @@
...
+    /**
+     * Find and read a skill or skills given its id, name, or pending status.
+     *
+     * @param skillid
+     * @param name
+     * @param pending
+     * @return
+     */
+
+    @Get("/{?skillid,name,pending}")
```
jahanner, line 75: Do we want to have to specify pending as part of the GET?

hwilliams314: Yes, they want to be able to search by pending status.

server/src/main/java/com/objectcomputing/checkins/services/skills/Skill.java
```diff
@@ -0,0 +1,77 @@
...
+
+    @Id
+    @Column(name="skillid")
+    @AutoPopulated
+    @TypeDef(type= DataType.STRING)
+    private UUID skillid;
+
+    @NotBlank
+    @NotNull
+    @Column(name="name", nullable = true, unique = true)
```
Bagurp, line 39: we're using @NotNull and nullable=true at the same time. Is this a good practice?

server/src/main/java/com/objectcomputing/checkins/services/skills/SkillController.java
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    /**
+     * Create and save a new skill.
+     *
+     * @param skill
+     * @return
+     */
+
+    @Post(value = "/")
```
Bagurp, line 45: @hwilliams314, could you add JavaDoc to this?

server/src/main/java/com/objectcomputing/checkins/services/skills/SkillController.java
```diff
@@ -0,0 +1,106 @@
...
+@Controller("/skill")
+@Secured(SecurityRule.IS_ANONYMOUS)
+@Produces(MediaType.APPLICATION_JSON)
+@Tag(name="skill")
+public class SkillController {
+
+    private static final Logger LOG = LoggerFactory.getLogger(SkillController.class);
+
+    @Inject
+    private SkillRepository skillRepo;
```
Bagurp, line 29: It looks like Logger and skillRepo are unused in this class. Do we still need it?

bhatiam12: @Bagurp skillRepo is being used on line 35. 
@hwilliams314 can we just rename it to skillRepository and i learned that we don't need @Inject annotation. Micronaut figures it out at compile time.

server/src/test/java/com/objectcomputing/checkins/services/skills/SkillControllerTest.java
```diff
@@ -0,0 +1,178 @@
...
+    @Client("/skill")
+    private HttpClient client;
+
+    SkillRepository mockSkillRepository = mock(SkillRepository.class);
+    Skill mockSkill = mock(Skill.class);
+
+    private static Date testDate = new Date(System.currentTimeMillis());
+    private static String testSkillName = "testName";
+    private static boolean pending = false;
+    private static boolean isDataSetupForGetTest = false;
```
Bagurp, line 35: It looks like isDataSetupForGetTest and testDate aren't being used in the test class. Could we remove these?

server/src/test/java/com/objectcomputing/checkins/services/skills/SkillControllerTest.java
```diff
@@ -0,0 +1,178 @@
...
+    public void testPutUpdateWithMissingField() {
+
+        HttpClientResponseException thrown = assertThrows(HttpClientResponseException.class, () -> {
+            client.toBlocking().exchange(HttpRequest.PUT("/updatePending", fakeBody));
+        });
+        assertNotNull(thrown);
+        assertEquals(HttpStatus.BAD_REQUEST, thrown.getStatus());
+    }
+
+    private void setupTestData() {
```
Bagurp, line 175: It looks like this method isn't being called. Do we need this?

server/src/main/java/com/objectcomputing/checkins/services/skills/SkillController.java
hwilliams314: @bhatiam12 I tried that. I took the SkillRepository @Inject out and renamed the skillRepo variable to skillRepository and it isn't compiling

server/src/main/java/com/objectcomputing/checkins/services/skills/SkillController.java
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    /**
+     * Load the current skills into checkinsdb.
+     *
+     * @param skillslist
+     * @return
+     */
+
+    @Post("/loadskills")
```
mkimberlin, line 62: Normally, I would say that we shouldn't have a verb in a URL. REST services as a rule are based around the entity and the "action" is driven by the HTTP command being used.  So, I would have probably made a single @Post service like the one below and not separated them out into two. However, since this is not a service that is intended for use by the frontend, but rather for convenience for now...I think it's ok.  Eventually we probably want to include the "load" of skills into a Flyway migration instead, at which point we can remove the loadSkills service.

server/src/main/java/com/objectcomputing/checkins/services/skills/SkillController.java
```diff
@@ -0,0 +1,110 @@
...
+        return found;
+
+    }
+
+    /**
+     * Update the pending status of a skill.
+     * @param skill
+     * @return
+     */
+    @Put("/updatePending")
```
mkimberlin, line 92: This is another violation of the whole "no verbs in the URL" rule.  This should just be `@Put("/")` and should be able to update anything about a skill...for example, why wouldn't I update the description with this service? There shouldn't be seperate services that are for updating different parts of an entity.  If you ever have a verb in your URL, you are probably doing something antithetical to the idea or RESTful services.

server/src/main/java/com/objectcomputing/checkins/services/skills/SkillServices.java
```diff
@@ -0,0 +1,76 @@
...
+
+    protected List<Skill> findByValue(UUID skillid, String name, Boolean pending) {
+        List<Skill> skillList = null;
+        if (skillid != null) {
+           skillList = Collections.singletonList(readSkill(skillid));
+        } else if(name != null) {
+            skillList = findByNameLike(name);
+        }  else if(pending != null) {
+            skillList = findByPending(pending);
+        }
```
mkimberlin, line 42: This is fine for now...however, I think there is a way to do a "findByILike" that takes an example entity and does case insensitive matching on *any and all* of the populated values.  I'd have to dig into that a little bit more...but as this stands, if I wanted to do a search for name="Java" and pending=true, I couldn't do that. it would only search by name and would return anything that matched regardless of pending.

server/src/main/java/com/objectcomputing/checkins/services/skills/SkillServices.java
```diff
@@ -0,0 +1,76 @@
...
+        return skillList;
+    }
+
+    protected List<Skill> findByPending(boolean pending) {
+        List<Skill> skillList = skillsRepo.findByPending(pending);
+
+        return skillList;
+    }
+
+    protected Skill updatePending(Skill skill) {
```
mkimberlin, line 60: Since this is just doing a simple update, that means your change to the "updatePending" service will only be a matter of renaming these and adjusting the url as I described.

PR #102, opened 2020-07-13
Feature/feeling component Closes #88

### LINKED ISSUE #88: FE> Feelings component
Definition of Done

- [x] display feelings prompt text
- [x] allow for the selection of likert scale of response icons
- [x] implement unit tests

jahanner:


web-ui/src/components/feelings/Feelings.jsx
```diff
@@ -0,0 +1,47 @@
+import React from "react";
+
+const Feelings = (props) => {
+  const { onSelect, message } = props;
```
mkimberlin, line 4: Any particular reason you do this props selector on a new line rather than inline in the method definition?  It's super minor, but whenever I see props passed in rather than a selector, it makes me feel like I need to look through the whole component to make sure there aren't other things that are passed in that I'm not noticing.

It might also help to define a propTypes.

web-ui/src/components/feelings/Feelings.test.js
```diff
@@ -0,0 +1,6 @@
+import React from "react";
+import Feelings from "./Feelings";
+
+it("renders correctly", () => {
+  snapshot(<Feelings />);
+});
```
mkimberlin, line 6: Wouldn't hurt to have a test to make sure that the onSelect gets called with the appropriate value.

web-ui/src/components/feelings/Feelings.jsx
mkimberlin: Also a displayName...

PR #108, opened 2020-07-16
Feature 86/rest service for pulse response

Bagurp:


server/src/main/java/com/objectcomputing/checkins/services/pulseresponse/PulseResponse.java
```diff
@@ -0,0 +1,113 @@
...
+@Table(name="pulse_response")
+public class PulseResponse {
+    
+    public PulseResponse() {}
+
+    public PulseResponse(LocalDate submissionDate, LocalDate updatedDate, UUID teamMemberId, String questionResponse, String internalFeelings, String externalFeelings) {
+        this.submissionDate = submissionDate;
+        this.updatedDate = updatedDate;
+        this.teamMemberId = teamMemberId;
+        this.questionResponse = questionResponse;
```
mkimberlin, line 26: This is actually not quite right. I'd remove questionResponse for now. There will be multiple questions...not just one, and as a result the answers will need to be stored in a separate table.

server/src/main/java/com/objectcomputing/checkins/services/pulseresponse/PulseResponseController.java
```diff
@@ -0,0 +1,98 @@
...
+
+    /**
+     * Update a Pulse Response.
+     * @param pulseResponse
+     * @return
+     */
+    @Put("/")
+    public HttpResponse<?> update(@Body @Valid PulseResponse pulseResponse) {
+
+        if(null != pulseResponse.getId()) {
```
mkimberlin, line 83: I have never seen anyone write null checks in this order...is this a thing now? Is there some reason for doing it this way?  Just curious more than anything...but it come across as much less readable to me.

server/src/main/resources/db/V3__create_pulse_response_table.sql
```diff
@@ -0,0 +1,10 @@
+drop table if exists pulse_response;
+CREATE TABLE pulse_response (
+   id varchar PRIMARY KEY,
+   submissionDate date,
+   updatedDate date,
+   teamMemberId varchar REFERENCES member_profile(uuid),
+   questionResponse varchar,
```
mkimberlin: Yank this column as well...and you're going to need to make this file V4, since Holly's PR went in first.

server/src/main/java/com/objectcomputing/checkins/services/pulseresponse/PulseResponseController.java
Bagurp: It's my negligence and not a thing! I've reversed it for now for better readability. If we want to it to be more formal, I could either use java.util.Objects.nonNull() or Apache.Commons.ObjectUtils.isNotEmpty()

server/src/test/java/com/objectcomputing/checkins/services/pulseResponse/PulseResponseControllerTest.java
```diff
@@ -0,0 +1,256 @@
...
+    
+    @Inject
+    @Client("/pulse-response")
+    private HttpClient client;
+
+    @Inject
+    MemberProfileController memberProfileController;
+    
+    PulseResponseRepository mockPulseResponseRepository = mock(PulseResponseRepository.class);
+    PulseResponse mockPulseResponse = mock(PulseResponse.class);
```
mkimberlin, line 52: This is unused.

server/src/test/java/com/objectcomputing/checkins/services/pulseResponse/PulseResponseControllerTest.java
```diff
@@ -0,0 +1,256 @@
...
+    
+    private static LocalDate testsubmissionDate = LocalDate.of(2020, 1, 01);
+    private static LocalDate testUpdatedDate = LocalDate.of(2020, 2, 01);
+    private static UUID testTeamMemberId;
+    private static boolean isDataSetupForTest = false;
+    
+    @Before
+    void setup() {
+        reset(mockPulseResponseRepository);
+        reset(mockPulseResponse);
```
mkimberlin, line 62: Except here...but obviously this isn't a reason to keep it.

server/src/test/java/com/objectcomputing/checkins/services/pulseResponse/PulseResponseControllerTest.java
```diff
@@ -0,0 +1,256 @@
...
+
+    @Test
+    public void testFindNonExistingEndpointReturnsNotFound() {
+        HttpClientResponseException thrown = assertThrows(HttpClientResponseException.class, () -> {
+            client.toBlocking().exchange(HttpRequest.GET("/bar?order=foo"));
+        });
+
+        assertNotNull(thrown.getResponse());
+        assertEquals(HttpStatus.NOT_FOUND, thrown.getStatus());
+    }
```
mkimberlin, line 104: How are the two tests above different?

server/src/test/java/com/objectcomputing/checkins/services/pulseResponse/PulseResponseControllerTest.java
```diff
@@ -0,0 +1,256 @@
...
+        PulseResponse testPulseResponse = new PulseResponse(testsubmissionDate, testUpdatedDate, testTeamMemberId, "Internal Feeling", "External Feeling");
+        testPulseResponse.setId(testId);
+
+        when(mockPulseResponseRepository.update(testPulseResponse)).thenReturn(testPulseResponse);
+
+        final HttpResponse<?> response = client.toBlocking().exchange(HttpRequest.PUT("", testBody));
+        assertEquals(HttpStatus.OK, response.getStatus());
+        assertNotNull(response.getContentLength());
+
+    }
```
mkimberlin, line 205: I notice there's not a test for updating a non-existent PulseResponse (one where the ID provided doesn't exist/isn't correct). Further, looking at your service layer code, I don't think that it is likely handled gracefully. I would consider adding a test for that and ensuring that your service responds accordingly.

server/src/test/java/com/objectcomputing/checkins/services/pulseResponse/PulseResponseControllerTest.java
Bagurp: Updated it in the latest commit

server/src/test/java/com/objectcomputing/checkins/services/pulseResponse/PulseResponseControllerTest.java
Bagurp: They weren't different from one another. I've removed one of them.

server/src/test/java/com/objectcomputing/checkins/services/pulseResponse/PulseResponseControllerTest.java
Bagurp: Good catch! I've added the validation for this and have moved all logic to a service class.

if there's no ID in HttpRequest -> throws Bad Request
if ID doesn't exist in DB -> throws Not Found
If ID is valid -> updates record in DB

server/src/main/resources/db/V4__create_pulse_response_table.sql
```diff
@@ -0,0 +1,9 @@
+drop table if exists pulse_response;
+CREATE TABLE pulse_response (
+   id varchar PRIMARY KEY,
+   submissionDate date,
+   updatedDate date,
+   teamMemberId varchar REFERENCES member_profile(uuid),
+   internalFeelings varchar,
+   externalFeelings varchar
+);
```
mkimberlin, line 9: I hate to be the bearer of bad news...but you're gonna have to change this to V5

mkimberlin: Do this before you merge...I'll approve anyway.

Bagurp: I've renamed the file to V5

PR #119, opened 2020-07-23
Feature 105/skill get by id closes issue 105

hwilliams314:


server/src/main/java/com/objectcomputing/checkins/services/skills/SkillServices.java
```diff
@@ -4,38 +4,43 @@
...
-        Skill returned = skillsRepo.findBySkillid(skillId);
+        Skill returned = skillRepository.findBySkillid(skillId);
 
         return returned;
 
     }
 
-    protected List<Skill> findByValue(UUID skillid, String name, Boolean pending) {
+    protected List<Skill> findByValue(String name, Boolean pending) {
+        LOG.info("findByValue");
```
chrismckiernan, line 40: This log statement still needed, or should it bet set to debug?

hwilliams314: not really needed. will set to debug.

hwilliams314:
missed a test case. will add and resubmit

PR #124, opened 2020-07-27
bugfix: Converts java.sql.date to localDate

Bagurp:
Looks like I missed this block of code when I converted java.sql.date to localDate. Thank you @hwilliams314 for catching this =)


bhatiam12:
@jahanner the builds are failing because of this : ```./src/context/SkillsContext.js
73
  Line 117:33:  'pdl' is assigned a value but never used  no-unused-vars```

PR #135, opened 2020-07-29
Feature 125/add fk to member profile, fix groovy tests

Bagurp:
* Adds foreign key to member_profile table, fix unit tests

Groovy - 
- Convert java.sql.Date to java.time.LocalDate
- fix tests after adding foreign key to MemberProfile and Checkin tables


server/src/main/java/com/objectcomputing/checkins/services/memberprofile/MemberProfile.java
```diff
@@ -48,6 +49,7 @@ public MemberProfile() {
     
     @Column(name="pdlId")
     @TypeDef(type=DataType.STRING)
+    @Nullable
```
Bagurp, line 52: Added nullable to pdlID column since an employee might not have a PDL at all times. Also, not all employees have an assigned PDL.

server/src/test/groovy/com/objectcomputing/checkins/controllers/CheckInControllerSpec.groovy
```diff
@@ -56,6 +57,7 @@ class CheckInControllerSpec extends EmbeddedServerSpecification implements Check
         cleanup:
         checkInRepository.deleteAll()
         memberProfileRepository.deleteAll()
+        skillRepository.deleteAll()
```
Bagurp, line 60: Need to delete all items in skillRepository to make sure 'assert !hasLeakage()' is satisfied. Records for Skills table is inserted from V4__load_initial_skills.sql that needs to be cleared.

mkimberlin:
@Bagurp There's a build failure related to the UI. Looks like an easy fix. Can you get that taken care of before merging, please?

PR #137, opened 2020-07-30
Selenium testing check in app

hwilliams314:


server/src/test/java/com/objectcomputing/checkins/endtoend/pages/HomePage.java
```diff
@@ -1,72 +1,79 @@
...
 
-	@FindBy(className="success")
-	WebElement success_msg;
-	
-	@FindBy(className="error")
-	WebElement error_msg;
-	
-	
+    WebDriver driver;
+    @FindBy(xpath = "/html[1]/body[1]/div[1]/form[1]/fieldset[1]/label[1]")
```
jorwalk, line 24: Why do all of these have `[1]`? instead of being `[0]`?

server/src/test/java/com/objectcomputing/checkins/endtoend/pages/HomePage.java
```diff
@@ -1,72 +1,79 @@
...
+		if (resource == null) {
+			throw new IllegalArgumentException("file is not found!");
+		} else {
+			return new File(resource.getFile());
+		}
+
 	}
+    public void uploadFileErrorMsg() throws InterruptedException {
+        uploadBtn.click();
+        Thread.sleep(5000);
```
jorwalk, line 71: Since the code define an implicit wait `driver.manage().timeouts().implicitlyWait(30, TimeUnit.SECONDS);`, could these `Thread.sleep()`'s be replaced with a wait for element to be visible or something similar?

server/src/test/java/com/objectcomputing/checkins/endtoend/pages/HomePage.java
mkimberlin: xpath indexes aren't 0 based...they start with 1 🤷‍♂️

server/src/test/java/com/objectcomputing/checkins/endtoend/pages/HomePage.java
mkimberlin: Probably...I think we have someone who is going to be diving into these tests more deeply. I am going to go ahead and merge this one as is and create an issue to remove `sleep` calls where possible.

PR #159, opened 2020-08-03
Feature 20/check in history Closes #20

### LINKED ISSUE #20: FE > Create check-in history component (M)
Depends on #19 (done).

- [x] Results should show team member avatar, PDL Name, and  check-in date.

- [x] Should be able to edit check-in date by clicking on calendar icon

jahanner:


web-ui/src/context/SkillsContext.js
```diff
@@ -7,6 +7,7 @@ export const MY_SKILL_TOGGLE = "toggle";
 export const MY_PROFILE_UPDATE = "update";
 export const UPDATE_PDL = "update_pdl";
```
chrismckiernan, line 8: SkillsContext is a placeholder right now that will be reworked and split up later? 

Holding default data so you don't have to mess around with inserting fake data?

I want to know your approach here, because I'm looking at working on a component for the FE.

Thanks

web-ui/src/context/SkillsContext.js
```diff
@@ -141,6 +180,22 @@ const reducer = (state, action) => {
       state.defaultTeamMembers = profiles;
       break;
     }
+    case UPDATE_CHECKIN:
+      const { date, index } = action.payload;
```
chrismckiernan, line 184: Will logic like this be reused for other components, could we make a function?

web-ui/src/pages/ResourcesPage.jsx
```diff
@@ -22,52 +13,65 @@ const ResourcesPage = () => {
   const ociLightBlue = "#72c7d5";
   const ociOrange = "#feb672";
 
+  const getPDF = async (name) => {
+    try {
+      const res = await fetch("http://localhost:3000/pdfs/" + name + ".pdf");
```
mkimberlin, line 18: why axios sometimes and fetch others? and this http://localhost:3000 stuff need to be removed, please.

PR #175, opened 2020-08-05
Update gradle-build-production.yml 

jorwalk:
Add steps to the workflow to build and deploy to a container optimized compute engine instance.

closes https://github.com/oci-labs/check-ins/issues/38

> We will need to update some variables for a true prod isolation: `DEV_DB_CONNECTION_NAME`, etc.

After this is merged, can we please submit a pull request from develop to master, in order to confirm and unblock any subsequent issues?


.github/workflows/gradle-build-production.yml
```diff
@@ -19,3 +22,25 @@ jobs:
...
+        uses: GoogleCloudPlatform/github-actions/setup-gcloud@master
+        with:
+          version: '290.0.1'
+          service_account_key: ${{ secrets.RUN_SA_KEY }}
+          project_id: ${{ secrets.RUN_PROJECT }}
+      - name: Build and push image to Google Container Registry
+        run: |-
+          gcloud builds submit \
+            --quiet \
+            --tag "gcr.io/$PROJECT_ID/$SERVICE_NAME:$GITHUB_SHA"
```
mkimberlin, line 35: The way this docker build seems to be happening is very different than dev. What is specifying the server directory for the docker build?

.github/workflows/gradle-build-production.yml
```diff
@@ -19,3 +22,25 @@ jobs:
...
+      - name: Deploy image to Compute Engine Instance
+        run: |-
+          gcloud compute instances update-container "$SERVICE_NAME" \
+            --zone "$GCE_INSTANCE_ZONE" \
+            --container-image "gcr.io/$PROJECT_ID/$SERVICE_NAME:$GITHUB_SHA" \
+            --container-env CLOUD_DB_CONNECTION_NAME=${{ secrets.DEV_DB_CONNECTION_NAME }}, \
+                            DB_NAME=${{ secrets.DEV_DB_NAME }}, \
+                            DATASOURCES_DEFAULT_PASSWORD=${{ secrets.DEV_DB_PASSWORD }}, \
+                            DATASOURCES_DEFAULT_USERNAME=${{ secrets.DEV_DB_USERNAME }}, \
```
mkimberlin, line 44: All of this is pointing to the DEV environment. Is this just a test? I'm concerned we will run into some flyway issues if we don't go ahead and split this out into its own DB fairly promptly.  I'm also not clear on what will separate the dev and prod compute engine instances. Could you point that out?

Would it be better to test out this method by creating a workflow for PRs that matches the feature branch? Then we could adjust the branch matching once we've proven out the process?

.github/workflows/gradle-build-production.yml
jorwalk: Good catch, I grabbed the build step from another project. This has been updated to have Docker build and push the container.

.github/workflows/gradle-build-production.yml
jorwalk: Thanks for the questions and logical thought process to find a common path. I've replied in kind below.

1. `All of this is pointing to the DEV environment.` Correct. We need to move away from Cloud Run.
1. ` Is this just a test?` This is a migration attempt using the credentials withheld in Github Secrets.
1. `I'm concerned we will run into some flyway issues if we don't go ahead and split this out into its own DB fairly promptly.` If the docker image running on GCE is the exact same as what is running on Cloud Run, will it trigger flyway to take any action?
1. `I'm also not clear on what will separate the dev and prod compute engine instances.` Naming convention and tags is what I was planning. Trying to keep complexity down, I was planning to name them `checkins-develop-dev` & `checkins-develop-prod`
1. `Would it be better to test out this method by creating a workflow for PRs that matches the feature branch?` Yes. This makes sense,  I will rename the file to `gradle-build-gce-dev.yaml`

PR #177, opened 2020-08-05
fix read all skills closes #176

### LINKED ISSUE #176: Fix Read All skills

hwilliams314:


server/src/main/java/com/objectcomputing/checkins/services/skills/SkillServices.java
```diff
@@ -36,6 +38,12 @@ protected Skill readSkill(UUID skillId) {
 
     }
 
+    public Set<Skill> readAll() {
+        Set<Skill> actionItems = new HashSet<>();
```
bhatiam12, line 42: Please can you update variable name because I think actionItems is for action item rest service ?

PR #178, opened 2020-08-05
Personnel component

chrismckiernan:


web-ui/src/App.js
```diff
@@ -52,9 +52,9 @@ function App() {
                 <Header title="Profile" />
                 <Profile />
               </Route>
-              <Route path="/checkins">
-                <Header />
-                <CheckinsPage />
+              <Route path="/personnel">
+                <Header title="My Personnel" />
+                <MyPersonnelPage />
```
mkimberlin, line 57: Instead of making a page for this, can it be integrated into the right hand side of the check-ins page?

It's eventually going to need to look something like:
Check-in

chrismckiernan: Done.
image

web-ui/src/components/personnel/Personnel.jsx
```diff
@@ -46,8 +49,9 @@ const PersonnelComponent = () => {
     updateCheckins();
   },[personnel]);
 
-  function createEntry(person, checkIn) {
-    let key = undefined;
+  // Create entry of member and their last checkin
+  function createEntry(person, checkIn, key) {
+    key = key ? key : undefined;
```
mkimberlin, line 54: Every time I see this (even when I've written it myself) it makes me tilt my head to the side a little.

chrismckiernan: Changed the key arg, so its more obvious.

PR #180, opened 2020-08-06
Update GoogleCloudDatabaseSetup.java

jorwalk:
Resolves question about, is it intentional that the defaultPassword is being logged.


mkimberlin:
Mine and yours are pretty similar, I cut out a bit more...I'm approving this. You can either merge this or mine, makes me no real difference.

PR #186, opened 2020-08-07
Feature 184/role service

chrismckiernan:


server/src/main/resources/db/V15__alter_create_roles_table.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE role
```
chrismckiernan, line 1: I may have to increment this if agenda gets merged first.

mkimberlin:
@chrismckiernan I think we need a delete method for this service as well.

chrismckiernan:
> @chrismckiernan I think we need a delete method for this service as well.

@mkimberlin Added.

PR #188, opened 2020-08-10
Feature 80/basicauth. Setup basic auth for local, else use OAuth2

chrismckiernan:
I'm not quite sure that the OAuth2 is setup correctly in the swagger yml file, this should be vetted later. It seems to be working, but we need flesh that side out more in another story. That story would include work on the google side to add roles and to add us developers to the project as members most likely. Need a spike to look in to how to do all that.

### DIFF
--- a/server/src/main/java/com/objectcomputing/checkins/security/UserPasswordAuthProvider.java
+++ b/server/src/main/java/com/objectcomputing/checkins/security/UserPasswordAuthProvider.java
@@ -1,0 +1,30 @@
+package com.objectcomputing.checkins.security;
+
+import io.micronaut.context.annotation.Requires;
+import io.micronaut.http.HttpRequest;
+import io.micronaut.security.authentication.*;
+import io.reactivex.Flowable;
+import org.reactivestreams.Publisher;
+
+import javax.inject.Inject;
+import javax.inject.Singleton;
+
+@Requires(env = {"local"})
+@Singleton
+public class UserPasswordAuthProvider implements AuthenticationProvider {
+
+    @Inject
+    UsersStore store;
+
+    @Override
+    public Publisher<AuthenticationResponse> authenticate(HttpRequest<?> httpRequest, AuthenticationRequest<?, ?> authReq) {
+        String username = authReq.getIdentity().toString();
+        String password = authReq.getSecret().toString();
+        if (password.equals(store.getUserPassword(username))) {
+            UserDetails details = new UserDetails(username, store.getUserRole(username));
+            return Flowable.just(details);
+        } else {
+            return Flowable.just(new AuthenticationFailed());
+        }
+    }
+}
--- a/server/src/main/java/com/objectcomputing/checkins/security/UserPasswordAuthProvider.java
+++ b/server/src/main/java/com/objectcomputing/checkins/security/UserPasswordAuthProvider.java
@@ -9,7 +9,7 @@ import org.reactivestreams.Publisher;
 import javax.inject.Inject;
 import javax.inject.Singleton;
 
-@Requires(env = {"local"})
+@Requires(env = {"local", "test"})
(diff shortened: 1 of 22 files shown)


server/src/main/resources/application-local.yml
```diff
@@ -1,5 +1,23 @@
...
+
 datasources:
   default:
     url: ${JDBC_URL:`jdbc:postgresql://localhost:5432/checkinsdb`}
     username: postgres
     password: 'postgres'
+
+credentials:
+  roles:
+    SUPER:
```
chrismckiernan, line 14: If you want to test an endpoint that is secured you will need to use one of these users. The password will be the same as the username.

In the swagger-ui for example you will want to click authorize and enter username: ADMIN, password: ADMIN

server/src/main/java/com/objectcomputing/checkins/security/UserPasswordAuthProvider.java
```diff
@@ -0,0 +1,30 @@
...
+import io.micronaut.context.annotation.Requires;
+import io.micronaut.http.HttpRequest;
+import io.micronaut.security.authentication.*;
+import io.reactivex.Flowable;
+import org.reactivestreams.Publisher;
+
+import javax.inject.Inject;
+import javax.inject.Singleton;
+
+@Requires(env = {"local"})
```
chrismckiernan, line 12: This means this bean will only exist when MICRONAUT_ENVIRONMENTS is set to local

server/src/main/java/com/objectcomputing/checkins/services/role/RoleController.java
```diff
@@ -23,7 +24,7 @@
 import java.util.UUID;
 
 @Controller("/services/role")
-@Secured(SecurityRule.IS_ANONYMOUS)
+@Secured(RoleType.Constants.ADMIN_ROLE)
```
chrismckiernan, line 27: Only authorized users with ADMIN role can reach any of these endpoints

server/src/main/resources/application-local.yml
```diff
@@ -1,5 +1,23 @@
+micronaut:
+  security:
+    enabled: true
+    reject-not-found: false
```
chrismckiernan, line 4: This is so we return 404 instead of 401/403 when using SECURED("SOME_ROLE") at the Controller level

server/src/main/java/com/objectcomputing/checkins/security/UserPasswordAuthProvider.java
mkimberlin: Doesn't this need to be local and test?

mkimberlin: And ditto for the UserStore?

chrismckiernan: @mkimberlin
Not necessarily, we could split it up that way, but then application-test.yml would be a subset of application-local.yml.

We could just set MICRONAUT_ENVIRONMENTS=local when running tests.

It depends on design questions like are we going to have tests that use OAuth to make sure it is working as expected?

Let me know the thoughts on what testing in the future we may want and I can make this a little more future proof/configurable. 

I can see pros and cons either way, but at the very least I agree "local" would be a lie if we were testing in GitHub. =)

chrismckiernan: I'll just do as you said for now, it makes sense enough. I think we may want to come back at this later in another story when we increase our testing footprint.

chrismckiernan: @mkimberlin Updated

PR #191, opened 2020-08-11
updated team service

Maroju100:
team & team member service implementation based on new definition


server/src/main/java/com/objectcomputing/checkins/services/team/member/TeamMemberServicesImpl.java
```diff
@@ -0,0 +1,86 @@
...
+    @Inject
+    private MemberProfileRepository memberRepo;
+
+    public TeamMember save(TeamMember teamMember) {
+        TeamMember teamMemberRet = null;
+        if (teamMember != null) {
+            final UUID teamId = teamMember.getTeamid();
+            final UUID memberId = teamMember.getMemberid();
+            if (teamId == null || memberId == null) {
+                throw new TeamBadArgException(String.format("Invalid teamMember %s", teamMember));
```
meyerinj, line 28: These null checks won't be hit in practice because the controller has @Valid annotation. Nothing wrong with two checks, but if we want this message format (rather than "teamMember.memberid: must not be null" which the ConstraintViolationException shows) we want the controller will need an error handler for ConstraintViolationException.

server/src/main/resources/db/V10__create_teams_table.sql
```diff
@@ -1,14 +1,15 @@
-DROP TABLE IF EXISTS team;
-CREATE TABLE team (
-  uuid varchar PRIMARY KEY,
-  name varchar,
-  description varchar
+drop table if exists teams;
+CREATE TABLE teams(
+   teamid varchar PRIMARY KEY,
+   name varchar UNIQUE,
+   description varchar
```
mkimberlin, line 5: This file can't be altered. You have to make a new migration file to make changes to a table. Also...rather than `teamid`, can we just go with `id`. I'd like to see this standardized across the project.

server/src/main/resources/db/V10__create_teams_table.sql
```diff
@@ -1,14 +1,15 @@
...
 );
 
-DROP TABLE IF EXISTS team_member;
-CREATE TABLE team_member (
-  uuid varchar PRIMARY KEY,
-  teamId varchar REFERENCES team(uuid),
-  memberId varchar REFERENCES member_profile(uuid),
-  isLead boolean default false
+drop table if exists teamMembers;
```
mkimberlin, line 8: also, not gonna lie, I'm not a big fan camel-cased table names.

Maroju100: updated the table name to team_member.

server/src/main/resources/db/V10__create_teams_table.sql
```diff
@@ -1,14 +1,14 @@
...
+CREATE TABLE team (
    teamid varchar PRIMARY KEY,
    name varchar UNIQUE,
    description varchar
 );
 
-drop table if exists teamMembers;
-CREATE TABLE teamMembers(
+drop table if exists team_member;
+CREATE TABLE team_member (
```
bhatiam12, line 9: as @mkimberlin said, this needs to be done in an another version. you will have to create another .sql file with V18 or whatever the next available version is and then do something like below. 
```
ALTER TABLE table_name
RENAME TO new_table_name;
```

server/src/main/java/com/objectcomputing/checkins/services/team/TeamController.java
```diff
@@ -1,94 +1,126 @@
...
+import javax.inject.Inject;
+import javax.validation.Valid;
+import javax.validation.constraints.NotNull;
+import java.net.URI;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Set;
+import java.util.UUID;
 
-@Controller(TeamController.TEAM_CONTROLLER_PATH)
```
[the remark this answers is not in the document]
mkimberlin: Ditching the path variable is fine now, I think...since the only other place it was used is now using the headers to get the location.  I agree on the `IS_ANONYMOUS` change though...that needs to go back to `IS_AUTHENTICATED`.  It means you will have to make use of the users/roles that @chrismckiernan created in your controller tests, but that is as simple as adding a `.basicAuth('MEMBER', 'MEMBER')` to your request creation in most cases.

server/src/main/resources/db/V10__create_teams_table.sql
Maroju100: added V18 alter script to update team/team_member table, updated teamid to uuid, some test issues if it was id.

server/src/main/java/com/objectcomputing/checkins/services/team/member/TeamMemberCreateDTO.java
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    public void setMemberid(UUID memberid) {
+        this.memberid = memberid;
+    }
+
+    public Boolean getLead() {
+        return lead;
+    }
+
```
hwilliams314, line 42: Shouldn't getLead be isLead? as it is a Boolean?

server/src/main/java/com/objectcomputing/checkins/services/team/member/TeamMemberServicesImpl.java
chrismckiernan: True, he copied what I originally wrote.

Since the service "could" be called from elsewhere I just threw that in there, I think mainly because Intellij was yelling at me.

I would argue just to leave it.

server/src/main/java/com/objectcomputing/checkins/services/team/TeamController.java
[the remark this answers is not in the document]
chrismckiernan: Just the constants instead of hardcoding 'MEMBER' =) but yes.

server/src/main/java/com/objectcomputing/checkins/services/team/member/TeamMemberCreateDTO.java
chrismckiernan: Probably should be for consistency. He copied this from my guild member dto most likely which also has getLead while the entity has isLead

server/src/test/java/com/objectcomputing/checkins/services/team/member/TeamMemberControllerTest.java
```diff
@@ -0,0 +1,416 @@
...
+        assertEquals(String.format("%s/%s", request.getPath(), g.getId()), response.getHeaders().get("location"));
+
+        verify(teamMemberServices, times(1)).save(any(TeamMember.class));
+    }
+
+    @Test
+    void testCreateAnInvalidTeamMember() {
+        TeamMemberCreateDTO teamMemberCreateDTO = new TeamMemberCreateDTO();
+
+        TeamMember g = new TeamMember(UUID.randomUUID(), UUID.randomUUID(), true);
```
chrismckiernan, line 67: replace these g's with tm or something.

server/src/main/java/com/objectcomputing/checkins/services/team/TeamController.java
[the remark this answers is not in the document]
chrismckiernan: I think in my role service change I made these IS_AUTHENTICATED, but those changes got wiped out when these changes were pushed. He didn't explicitly remove it, just was working of the code before I made my changes. That being said this should be changed as Michael said.

server/src/test/java/com/objectcomputing/checkins/services/team/member/TeamMemberControllerTest.java
Maroju100: updated to tm and tm2

server/src/main/java/com/objectcomputing/checkins/services/team/TeamController.java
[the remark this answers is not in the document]
mkimberlin: @Maroju100 I think getting this back to `IS_AUTHENTICATED` and the `isLead` change in the dto, and you'll be good to go!

server/src/main/java/com/objectcomputing/checkins/services/team/member/TeamMemberServicesImpl.java
chrismckiernan: If we decide to take a different approach, then lets do something similar to what I commented on Hollys PR:

> Replace this with, and make similar changes to update:
> 
> ```
>     public MemberSkill save(@NotNull MemberSkill memberSkill) {
>         MemberSkill memberSkillRet;
>         
>         @NotNull
>         final UUID memberId = memberSkill.getMemberid();
>         @NotNull
>         final UUID skillId = memberSkill.getSkillid();
> 
>         if (memberSkill.getId() != null) {
>             throw new MemberSkillBadArgException(String.format("Found unexpected id %s for member skill", memberSkill.getId()));
>         } else if (memberProfileRepository.findById(memberId).isEmpty())) {
>             throw new MemberSkillBadArgException(String.format("Member Profile %s doesn't exist", memberId));
>         } else if (skillRepository.findById(skillId).isEmpty()) {
>             throw new MemberSkillBadArgException(String.format("Skill %s doesn't exist", skillId));
>         } else if (memberSkillRepository.findByMemberidAndSkillid(memberSkill.getMemberid(),
>                 memberSkill.getSkillid()).isPresent()) {
>             throw new MemberSkillBadArgException(String.format("Member %s already has this skill %s", memberId, skillId));
>         }
> 
>         memberSkillRet = memberSkillRepository.save(memberSkill);
>         return memberSkillRet;
> 
>     }
> ```
> 
> Add `@Valdidated` to the class.
>

server/src/main/java/com/objectcomputing/checkins/services/team/TeamController.java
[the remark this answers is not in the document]
Maroju100: added IS_AUTHENTICATED and isLead change.

server/src/main/java/com/objectcomputing/checkins/services/team/member/TeamMemberCreateDTO.java
Maroju100: updated

server/src/main/resources/db/V10__create_teams_table.sql
Maroju100: updated

PR #195, opened 2020-08-12
questions updated to use service interface, errors, and assorted othe…

meyerinj:
…r changes


bhatiam12:
@meyerinj branch name should be in format `feature-[Issue Number]/* ` for workflows to kick in.

Issue #71, opened 2020-06-30
BE > Associate uploaded documents with a check-in (M)


sharmag04:
Definition of Done

- [x] Create a REST Service (checkin-document) that maps checkins table record with uploaded document ID. It should be noted that, a check-in record could contain multiple documents. 
- [x] This service is called after successfully uploading a file.
- [x] unit tests are created for each endpoint
- [x] code is submitted as a pull request
- [x] swagger configuration
- [x] split the logic into a service layer
- [x] define appropriate DTOs
- [x] to implement a error handler
- [x] change the implementation find by methods to `retainall`
- [x] make use of the headers for location
- [x] alter unit test
- [x] add schema definition to the entity 

# Initial entity should contain:
- Id
- CheckinsId
- UploadDocID

Base URL: /services/checkin-document 

- [x] Create
- [x] Read: Find By CheckinsId
- [x] Update
- [x] Delete: Only an admin can delete records

Bagurp:
Front-End:
- [ ] Add components to select team-member, target quarter, target year before uploading/checking in reviews. Components to be added in /upload page. Default to be current year and quarter, drop-downs to be added to select a different value.

Back-End:
- [ ] Create a REST Service that maps checkins table record with uploaded document ID. It should be noted that, a check-in record could contain multiple documents. This service is called by upload service after successfully uploading a file.

Entity should contain:
Id
CheckinsId
UploadDocID

Base URL: /checkin-document

- [ ] Create
- [ ] Read: Find By CheckinsId
- [ ] Update
- [ ] Delete: Only an admin can delete records

PR #196, opened 2020-08-12
Issue #71 - checkinDocument - secure delete endpoint for admin access

Bagurp:


server/src/main/java/com/objectcomputing/checkins/services/checkindocument/CheckinDocumentController.java
```diff
@@ -27,7 +28,7 @@
 import io.swagger.v3.oas.annotations.tags.Tag;
 
 @Controller("/services/checkin-document")
-@Secured(SecurityRule.IS_ANONYMOUS)
+@Secured({RoleType.Constants.ADMIN_ROLE, RoleType.Constants.PDL_ROLE})
```
Bagurp, line 31: Only Admin, PDLs are allowed to access the endpoints

PR #204, opened 2020-08-13
Feature 194/testcontainer role

chrismckiernan:


chrismckiernan:
Unwanted code got in.

server/src/test/java/com/objectcomputing/checkins/services/role/RoleControllerTest.java
```diff
@@ -2,6 +2,10 @@
```
chrismckiernan, line 2: I replaced RoleControllerTest with testcontainer. I could be talked in to keeping the old mockito code, and renaming this.

server/src/test/java/com/objectcomputing/checkins/services/TestContainersSuite.java
```diff
@@ -0,0 +1,41 @@
...
+import org.testcontainers.junit.jupiter.Container;
+import org.testcontainers.junit.jupiter.Testcontainers;
+import javax.inject.Inject;
+
+@MicronautTest(environments = "test", transactional = false)
+@Testcontainers
+public abstract class TestContainersSuite implements RepositoryFixture {
+
+    @Container
+    private static final PostgreSQLContainer<?> postgres = new PostgreSQLContainer<>("postgres:11.6");
```
chrismckiernan, line 19: I wanted to make this hardcoded postgres a property in the application.yml, but my `@Value("${tcimage:postgres:11.6}")` wasn't working. If anyone can figure it, please push to this branch =)

server/src/test/java/com/objectcomputing/checkins/services/role/RoleControllerTest.java
mkimberlin: @Bagurp @bhatiam12 @meyerinj I'd like to hear your input on this...I'm also on the fence about it. If I thought we would be able to better protect our quality by having both, I would get behind it. I'm not sure that I can see an argument for us being unable to test everything in the controller with these testcontainer / integration-style tests.

server/src/test/java/com/objectcomputing/checkins/services/TestContainersSuite.java
mkimberlin: Did you try doing something like: https://github.com/oci-labs/check-ins/blob/develop/server/src/main/java/com/objectcomputing/checkins/gcp/postgres/GoogleCloudDatabaseSetup.java#L14

chrismckiernan: I think it has something to do with `@Testcontainers` and `@Container` it looks like it tries to start and create that first before Micronaut has a chance to inject the value from the application.yml. I'll look in to it a little bit.

chrismckiernan: Well I didn't get to this today like I wanted. I'm going to punt on this one for now. I'll fix it later, won't be hard to retrofit in later.

server/src/test/java/com/objectcomputing/checkins/services/role/RoleControllerTest.java
bhatiam12: @mkimberlin @chrismckiernan tbh I am very new to test containers and this will be a learning for me. that said if we can test everything in controllers using the testcontainer approach then I am ok with getting rid of mockito way.

server/build.gradle
```diff
@@ -122,9 +120,11 @@ tasks.withType(JavaCompile) {
     }
 }
 
-processResources {
-    into('public') {
-        from configurations.publicResources
+if(System.getenv("SKIP_WEB_UI") == null) {
```
mkimberlin, line 123: might be worth defining a chore to get the build a bit more developer friendly.

mkimberlin: I'm not calling this out as anti-developer friendly, to be clear....I just know this was a bi-product of developer annoyance and wanted to call out that we need to try to fix it. ;-)

PR #205, opened 2020-08-14
Feature 168/member profile service update

### LINKED ISSUE #168: BE > Updating the member profile service (S)
Definition of Done

- [x] split the logic into a service layer
- [x] define appropriate DTOs
- [x] to implement a error handler
- [x] change the implementation find by methods to `retainall`
- [x] make use of the headers for location
- [x] alter unit test
- [x] add schema definition to the entity

meyerinj:
closes #168

### DIFF
--- a/server/src/main/java/com/objectcomputing/checkins/services/memberprofile/MemberProfile.java
+++ b/server/src/main/java/com/objectcomputing/checkins/services/memberprofile/MemberProfile.java
@@ -13,6 +13,7 @@ import javax.annotation.Nullable;
 import io.micronaut.data.annotation.AutoPopulated;
 import io.micronaut.data.annotation.TypeDef;
 import io.micronaut.data.model.DataType;
+import io.swagger.v3.oas.annotations.media.Schema;
 
 @Entity
 @Table(name ="member_profile")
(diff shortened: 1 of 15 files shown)


server/src/main/java/com/objectcomputing/checkins/services/memberprofile/MemberProfile.java
```diff
@@ -38,34 +39,43 @@ public MemberProfile() {
...
+    @Schema(description = "id of the member profile this entry is associated with", required = true)
     private UUID uuid;
 
     @NotNull
     @Column(name = "name", nullable = false)
+    @Schema(description = "full name of the employee", required = true)
     private String name;
 
     @Column(name="role")
+    @Schema(description = "employee's role at the company", required = true)
```
chrismckiernan, line 51: This might be a strange question but @mkimberlin will everyone with access to this site be an employee of OCI?

mkimberlin: This is not a strange question. Initially...yes...everyone will be an employee of OCI.

Eventually, this may be something that is worth making multi-tenant...but that wouldn't effect this

Issue #199, opened 2020-08-12
FE > Create check-in guides component 


mkimberlin:
Create a component(s) that allows for the download of the check-in guides.

Definition of Done

- [x] Component(s) are created similar to this mockup:
![Check-in Guides](https://user-images.githubusercontent.com/100510/90070345-561ca600-dcb9-11ea-846a-734abc98f819.png)

- [x] Component(s) shows static links and descriptions for each of the check-in guides based on role:
  - [x] PDLs see both PDL and Team Member guides
  - [x] Team Members see only Team Member guides
- [x] Guides are opened in a new window/tab
- [x] Snapshot test(s) have been created for the component
- [x] Applicable unit tests have been created for the component
- [x] Code has been submitted for PR and accepted

mkimberlin:
Closed by #215

PR #215, opened 2020-08-19
Added guides links. I hope y'all like it

meyerinj:
#199 

Material-UI doesn't seem to have an icon for Acrobat, so I used a PDF icon. The rest is probably equally wrong.


web-ui/src/components/guides/GuideLink.jsx
```diff
@@ -0,0 +1,18 @@
...
+import React from "react";
+
+const GuideLink = (props) => {
+    const path = "../../../pdfs/";
+    const fileName = props.name;
+    let fullPath = path + fileName + ".pdf";
+    return (
+        <li>
+            <PdfIcon />
+            <a href={fullPath} target="_blank">
```
chrismckiernan, line 11: Pipeline check fails, it treats warning as errors, this is what it said.  

`Line 11:32:  Using target="_blank" without rel="noopener noreferrer" is a security risk: see https://mathiasbynens.github.io/rel-noopener  react/jsx-no-target-blank`

meyerinj: This passes when I run it locally. Do you know what check is being done on the server side so I don't have to wait until I push to catch these?

chrismckiernan: Any warnings will be treated as errors in the pipeline, they are ignored in local runs. You can see warnings in the log on the console and you can set CI=true as an environment variable as well, this will cause warnings to fail the build.

web-ui/src/components/guides/GuideLink.jsx
```diff
@@ -0,0 +1,18 @@
+import PdfIcon from "@material-ui/icons/PictureAsPdf";
+import React from "react";
+
+const GuideLink = (props) => {
+    const path = "../../../pdfs/";
```
mkimberlin, line 5: I think this needs to be `/pdfs/`...that is where those files will resolve.

web-ui/src/components/guides/GuidesPanel.jsx
```diff
@@ -0,0 +1,57 @@
+import React, { useContext } from "react";
+import { AppContext } from "../../context/AppContext";
+import "./GuidesPanel.css";
+import GuideLink from "./GuideLink"
+
+const GuidesPanel = () => {
+    const { state } = useContext(AppContext);
+    const isPdl = state.user.role === "pdl";
```
jahanner, line 8: with what we have right now, though I think it may change, this would be state.userDate.role

meyerinj: I based this on the defaultProfile field on AppContext.js. I find it very plausible that this isn't how it's going to work in reality, bu the guidance I got is that this profile stuff isn't fully implemented at all so whatever I put here will probably have to change.

PR #218, opened 2020-08-20
Feature 211/security env

chrismckiernan:
This was an interesting one. I will have to write up wiki documentation for this. I will do that sometime this week.


server/src/main/java/com/objectcomputing/checkins/security/UserPasswordAuthProvider.java
```diff
@@ -20,11 +23,7 @@
     public Publisher<AuthenticationResponse> authenticate(HttpRequest<?> httpRequest, AuthenticationRequest<?, ?> authReq) {
         String username = authReq.getIdentity().toString();
         String password = authReq.getSecret().toString();
-        if (password.equals(store.getUserPassword(username))) {
-            UserDetails details = new UserDetails(username, store.getUserRole(username));
-            return Flowable.just(details);
-        } else {
-            return Flowable.just(new AuthenticationFailed());
-        }
+        UserDetails details = new UserDetails(username, store.getUserRole(password));
```
chrismckiernan, line 26: The way it will work is that, the "roleCred" passed in will set the roles, and the username is expected to be the email. 

I will talk more about this in the wiki.

PR #222, opened 2020-08-24
Query annotation spike

meyerinj:
Submitting for commentary on a possible way to do the queries.


server/src/main/java/com/objectcomputing/checkins/services/memberprofile/MemberProfileServices.java
```diff
@@ -1,12 +1,12 @@
...
 
-import java.util.Set;
+import java.util.List;
 import java.util.UUID;
 
 public interface MemberProfileServices {
     MemberProfile getById(UUID id);
 
-    Set<MemberProfile> findByValues(String name, String role, UUID pdlId);
+    List<MemberProfile> findByValues(String name, String role, UUID pdlId);
```
chrismckiernan, line 9: Why a list, wouldn't a set be more apropos since no duplicates? Doesn't matter either way we decide.

server/src/main/java/com/objectcomputing/checkins/services/memberprofile/MemberProfileRepository.java
```diff
@@ -6,6 +6,9 @@
 import javax.validation.constraints.NotBlank;
 
 import javax.annotation.Nullable;
+import javax.validation.constraints.Null;
```
chrismckiernan, line 9: Unused

server/src/main/java/com/objectcomputing/checkins/services/memberprofile/MemberProfileServices.java
chrismckiernan, line 9: Make UUID in this case just a String and pass it in.

server/src/main/java/com/objectcomputing/checkins/services/memberprofile/MemberProfileServicesImpl.java
```diff
@@ -23,18 +20,9 @@ public MemberProfile getById(UUID id) {
...
-            foundProfiles.retainAll(memberProfileRepository.findByName(name));
-        }
-        if (!StringUtils.isEmpty(role)) {
-            foundProfiles.retainAll(memberProfileRepository.findByRole(role));
-        }
-        if (pdlId != null) {
-            foundProfiles.retainAll(memberProfileRepository.findByPdlId(pdlId));
-        }
-        return foundProfiles;
+    public List<MemberProfile> findByValues(String name, String role, UUID pdlId) {
```
chrismckiernan, line 23: Switch pdlId to be a String, and pass that value right in.

chrismckiernan: This will result in some refactoring of the tests.

server/src/main/java/com/objectcomputing/checkins/services/memberprofile/MemberProfileRepository.java
```diff
@@ -19,4 +22,12 @@
     List<MemberProfile> findByRole(@NotBlank String name);
     List<MemberProfile> findByPdlId(@NotBlank UUID pdlId);
     List<MemberProfile> findAll();
+
+
+    @Query(" SELECT * " +
```
chrismckiernan, line 27: Looks good to me, nice formatting.

server/src/main/java/com/objectcomputing/checkins/services/memberprofile/MemberProfileServicesImpl.java
mkimberlin: +1

server/src/main/java/com/objectcomputing/checkins/services/memberprofile/MemberProfileRepository.java
```diff
@@ -23,4 +22,11 @@
     List<MemberProfile> findByPdlId(@NotBlank UUID pdlId);
     List<MemberProfile> findAll();
```
Bagurp, line 23: It looks like lines 20-24 are unused. Should we remove those?

meyerinj:
This is a draft PR intended only to demonstrate how the query annotation could be used. If we actually do this then test changes will almost certainly be required, but this PR shouldn't actually be merged.

server/src/main/java/com/objectcomputing/checkins/services/memberprofile/MemberProfileServices.java
```diff
@@ -6,7 +6,7 @@
 public interface MemberProfileServices {
     MemberProfile getById(UUID id);
 
-    Set<MemberProfile> findByValues(String name, String role, UUID pdlId, String workEmail);
+    Set<MemberProfile> findByValues(String name, String role, String pdlId, String workEmail);
```
bhatiam12, line 9: @meyerinj why was pdlId changed to string from UUID ? I ask because I started converting checkin notes and got hit with an error : 
```
Internal Server Error: SQL Error executing Query: ERROR: operator does not exist: character varying = uuid
  Hint: No operator matches the given name and argument type(s). You might need to add explicit type casts.
  Position: 67
```
I am assuming this error is because in our Postgres SQL script we have defined UUID as varchar because I think there some issues defining it as UUID, can't remember. @mkimberlin is that correct ?

I found a solution and it works, adding a cast to datatype resolves that error.

```

CREATE CAST (varchar AS uuid)
  WITH INOUT
  AS IMPLICIT;
```

My question is, do we create a new sql script to create the table again and add this cast or there is a way to alter the table and add the cast, can't find that on how to alter? @meyerinj @mkimberlin

hwilliams314: Postgres doesn't play nice with UUIDs that is why all the ids are stored as varchar. Then in the entity classes we have this line on the id field    @TypeDef(type = DataType.STRING) .  This has been working.

bhatiam12: @hwilliams314 it doesn't work when you convert findbyfields() method to use @Query annotation approach. You get an error that I posted.

meyerinj: @bhatiam12 To clarify, you converted the UUID param on findByCheckinid and findByCreatedbyid to a String and it caused an exception?

bhatiam12: @meyerinj I didn't convert it to string, it is UUID. here is the code : 
```
  @Query(" SELECT * " +
            "FROM checkin_notes cn " +
            "WHERE (:checkinid IS NULL OR cn.checkinid = :checkinid) " +
            "AND (:createbyid IS NULL OR cn.createdById = :createbyid) ")
    Set<CheckinNote> search(UUID checkinid, UUID createbyid);
```

meyerinj: Okay, I think I understand. Your suggestion is that rather than changing the repository to use a String we should leave it as a UUID and use the cast code that posted above?

bhatiam12: yup, correct. was this error the reason that you changed repository to use string or was it something else ?

meyerinj: It was. I'm not opposed to using this workaround instead of changing the data type in the Java code.

mkimberlin: I think another option would be to cast in the query: `OR cn.checkinid::text =`

bhatiam12: let me try casting in the query.

PR #228, opened 2020-08-26
Feature 227/security cleanup

chrismckiernan:
https://github.com/oci-labs/check-ins/wiki/Development

Added above wiki documentation which was written with this PR in mind.


server/src/main/java/com/objectcomputing/checkins/security/LocalOauthUserDetailMapper.java
```diff
@@ -80,7 +99,11 @@
         Optional<String> idToken = new JwtTokenGenerator(signatureGeneratorConfiguration, null, claimsGenerator)
                 .generateToken(Map.of("exp", System.currentTimeMillis() / 1000 + 60 * 60,
                         "sub", email,
-                        "roles", roles));
+                        "roles", roles,
+                        "email", email,
+                        "name", name,
+                        "picture", "https://upload.wikimedia.org/wikipedia/commons/7/74/SNL_MrBill_Doll.jpg"));
```
Bagurp, line 105: A+ for this image :)

server/src/main/java/com/objectcomputing/checkins/services/memberprofile/currentuser/CurrentUserDTO.java
```diff
@@ -1,132 +1,80 @@
...
 import javax.annotation.Nullable;
 import javax.validation.constraints.NotNull;
-import javax.validation.constraints.Past;
-import java.time.LocalDate;
-import java.util.UUID;
+import java.util.List;
+import java.util.Objects;
 
 @Introspected
 public class CurrentUserDTO {
```
Bagurp, line 13: Wouldn't we need these? Like, if an existing user logs in to the system (with a completed member-profile), shouldn't we return the whole member-profile?

server/src/main/resources/application.yml
```diff
@@ -44,8 +43,12 @@ micronaut:
           - isAuthenticated()
     reject-not-found: false
     redirect:
+      login-success: /
+      on-rejection: /oauth/login/google
       unauthorized:
         url: /oauth/login/google
+      forbidden:
+        url: /forbidden
     session:
```
Bagurp, line 52: We should remove lines 51-53 since it is deprecated.

https://micronaut-projects.github.io/micronaut-security/snapshot/guide/configurationreference.html#io.micronaut.security.token.config.TokenConfigurationProperties (end-of-file)

server/src/main/resources/application-local.yml
```diff
@@ -30,8 +30,15 @@ micronaut:
...
+          - isAuthenticated()
     reject-not-found: false
+    redirect:
+      login-success: /
+      on-rejection: /oauth/login/google
+      unauthorized:
+        url: /oauth/login/google
+      forbidden:
+        url: /forbidden
     session:
```
Bagurp, line 42: same with this

server/src/main/java/com/objectcomputing/checkins/services/memberprofile/currentuser/CurrentUserDTO.java
chrismckiernan: We do. That is what the member profile is stored for. It is serialized and deserialized correctly. If you run the app and do a console log on user profile. You will be able to see the json returned back which includes the whole profile.

server/src/main/java/com/objectcomputing/checkins/security/LocalOauthUserDetailMapper.java
chrismckiernan: Thanks =)

server/src/main/resources/application.yml
chrismckiernan: Done

server/src/main/resources/application-local.yml
chrismckiernan: Done

PR #229, opened 2020-08-26
Updated controller tests to use test containers

bhatiam12:


server/src/test/java/com/objectcomputing/checkins/services/checkinnotes/CheckinNoteControllerTest.java
```diff
@@ -1,81 +1,67 @@
 package com.objectcomputing.checkins.services.checkinnotes;
```
chrismckiernan, line 2: Can you run test coverage on this one file and check the code coverage.

Can you add more tests such as 

- creating an already existing checkin note (checkInId and memberProfileId match).
- trying to create a checkin note with a non existing checkin
- trying to create a checkin note with a non existing memberProfile
- A find with expected multiple results return along with a few notes created that won't be chosen, which should not be expected to be in the results.

chrismckiernan: Code looks good.

bhatiam12: Sure, I can add more tests. I just converted existing tests to use test containers as we have service tests(mock) as well which will take care of other scenarios.

chrismckiernan: Good point...

@mkimberlin Did we decide on whether to abandon the unit tests for impl, and focus all on testcontainers?

bhatiam12: done. added more tests and it should cover everything.

Issue #210, opened 2020-08-18
BE > Add a default mapping to allow all non-mapped requests to go to the UI


mkimberlin:
Implement a UI forwarding controller similar to:

https://github.com/jhipster/jhipster-sample-app-micronaut/blob/master/src/main/java/io/github/jhipster/sample/web/rest/ClientForwardController.java

This will allow us to deep link into the UI.

PR #250, opened 2020-08-26
Issue #210 - logic to forward any unmapped paths to the client

Bagurp:
- I have removed unauthorized.html since our latest configuration automatically sends user to /oauth/login/google


server/src/main/resources/views/forbidden.html
```diff
@@ -21,8 +20,7 @@
                </div>
                <div class="col-md-10">
                     <h1>Forbidden</h1>
-                    <p>Please Login to continue browsing.</p>
-                    <a class="btn btn-danger" href="/oauth/login/google">Login</a>
```
Bagurp, line 25: clean up and not part of this story. I removed the login button from Forbidden.html, since an unauth user cannot land on this page.

PR #261, opened 2020-08-28
Feature 211/security env

chrismckiernan:


server/src/test/java/com/objectcomputing/checkins/endtoend/testcases/TC01_Verify_Upload_File_Success_Msg.java
```diff
@@ -18,6 +18,7 @@
  */
```
bhatiam12, line 18: is there a reason these tests are commented out ?

chrismckiernan: That endpoint they are testing is not reachable `checkins.objectcomputing.com/`.

I think we are likely getting rid of these selenium tests in the future, but for now I don't want their test failures stopping builds.
These aren't good tests to decide whether the build is good or not in their current state.

Bagurp:
server/src/test/resources/application-local.yml vs server/src/test/resources/application-localtest.yml

Bagurp:
@chrismckiernan  \*asking for a friend\* - since we have multiple config files under /test/resources/ does micronaut -
* apply the config in server/src/test/resources/application-local.yml to the code where we have used @requires(env="local")
* apply the config in server/src/test/resources/application-localtest.yml to the methods/classes we have used @requires(env="test")

chrismckiernan:
> @chrismckiernan *asking for a friend* - since we have multiple config files under /test/resources/ does micronaut -
> 
>     * apply the config in server/src/test/resources/application-local.yml to the code where we have used @requires(env="local")
> 
>     * apply the config in server/src/test/resources/application-localtest.yml to the methods/classes we have used @requires(env="test")


The requires says to only allow this bean to be created when the environment is test. That way beans that are only meant for one environment aren't floating around in say production.

For the second;
```
@MicronautTest(environments = {"local", "localtest"}, transactional = false)
public class LocalLoginControllerTest extends TestContainersSuite implements MemberProfileFixture, RoleFixture {
```
The environments key in the MicronautTest annotation, tells the context to apply local env configurations, then apply on localtest.
I do that because in localtest I want to override the datasource to use test containers.

chrismckiernan:
> @chrismckiernan *asking for a friend* - since we have multiple config files under /test/resources/ does micronaut -
> 
>     * apply the config in server/src/test/resources/application-local.yml to the code where we have used @requires(env="local")
> 
>     * apply the config in server/src/test/resources/application-localtest.yml to the methods/classes we have used @requires(env="test")

@Bagurp 
The requires says to only allow this bean to be created when the environment is test. That way beans that are only meant for one environment aren't floating around in say production.

For the second;
```
@MicronautTest(environments = {"local", "localtest"}, transactional = false)
public class LocalLoginControllerTest extends TestContainersSuite implements MemberProfileFixture, RoleFixture {
```
The environments key in the MicronautTest annotation, tells the context to apply local env configurations, then apply on localtest.
I do that because in localtest I want to override the datasource to use test containers.

Bagurp:
@chrismckiernan Thank you!

PR #263, opened 2020-08-28
Fix upload tests

chrismckiernan:
This should fix the last tests as they are written now. These will be rewritten for the upload refactoring story.


bhatiam12:
One of them is still failing : 
```UploadControllerTest > testUploadFile() FAILED
70
    io.micronaut.http.client.exceptions.HttpClientResponseException at DefaultHttpClient.java:2046
71
```

PR #280, opened 2020-08-31
Convert member skill to test containers

bhatiam12:
Changed packaged name as well go be consistent with other services.


server/src/test/java/com/objectcomputing/checkins/services/member_skill/MemberSkillControllerTest.java
```diff
@@ -102,177 +86,190 @@ void testCreateANullMemberSkill() {
         assertEquals(request.getPath(), href.asText());
         assertEquals(HttpStatus.BAD_REQUEST, responseException.getStatus());
 
-        verify(memberSkillServices, never()).save(any(MemberSkill.class));
     }
 
+    @Test
+    void createAMemeberSkillForExistingMemberProfileO() {
+        MemberProfile memberProfile = createADefaultMemberProfile();
+        Skill skill = createADefaultSkill();
```
Maroju100, line 94: value of local variable skill not used.

bhatiam12: removed

Issue #220, opened 2020-08-21
FE > Load and/or create check -in on check-ins page (M)


mkimberlin:
If a user visits the check-ins page it should load the appropriate check-in.

Definition of done:
- [x] Appropriate check-in is loaded for the current team member
- [x] URL is updated to include the check-in id (i.e. `/checkins/B1G0LCH3CK1NID`)
- [x] If there is no open checkin (not submitted), then one should be created 3 months from the last checkin date OR one month from today if none exists
- [x] Snapshot test(s) have been updated (if needed)
- [ ] Applicable unit tests have been created for the component
- [ ] Code has been submitted for PR and accepted

PR #282, opened 2020-09-01
Feature 258/update checkin history && Feature 220/Load and/or create check -in on check-ins page Closes #258 && #220

### LINKED ISSUE #258: FE> Update Checkin History component (M)
- [x] If one of the arrows is clicked, Checkin History should update the checkin index and pass to the parent

- [x] If date is altered, Checkin History should do a PUT and update the date/time

- [x] Snapshot test(s) have been updated (if needed)

- [ ] Applicable unit tests have been created for the component

- [ ] Code has been submitted for PR and accepted

jahanner:


web-ui/src/components/notes/Note.jsx
```diff
@@ -31,6 +37,8 @@ const Notes = (props) => {
   const { checkin, memberName } = props;
   const [note, setNote] = useState({});
   const [isLoading, setIsLoading] = useState(true);
+  const [open, setOpen] = useState(false);
+  const [toast, setToast] = useState("");
```
jahanner, line 41: @mkimberlin ;)

web-ui/src/context/AppContext.js
```diff
@@ -55,26 +63,65 @@ const AppContextProvider = (props) => {
     updateUserProfile();
   }, []);
 
+  const date = (months, prevCheckinDate) => {
```
meyerinj, line 66: If I understand this right (and I very much might not) this will produce an array of numbers (e.g. "[2020, 9, 8]") but the service is expecting an ISO 8601 date (e.g. "2020-09-08").  Does React re-render this somewhere in the background or something?

jahanner: No, you're correct. The service was a bit odd because it said it expected the date in this format '2020-09-08' but then returned the date to me in this format [2020, 9, 8]. It actually accepts either format and since it returns an array, I wanted to try to keep it a bit more uniform.

web-ui/src/api/member.js
```diff
@@ -47,8 +47,8 @@ export const updateMember = async (member) => {
       method: "put",
       url: memberProfileUrl,
       responseType: "json",
-      data: member,
-      withCredentials: true
+      data: { ...member, id: member.id },
```
mkimberlin, line 50: am I just being thick here? what is the need for the `id: member.id` when you already have `...member`?

jahanner: Thanks for catching that, this was the result of a mixup from someone's PR that changed uuid to id

jahanner: I'll wait to push that up in case you find anything else

web-ui/src/components/checkin/__snapshots__/CheckinHistory.test.js.snap
```diff
@@ -0,0 +1,3 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`renders correctly 1`] = `<div />`;
```
mkimberlin, line 3: This doesn't look like a useful snapshot. I would think there has to be a way to mock that call and wait for the re-render, no? I'd do some digging on that. An empty div snapshot is kinda pointless.

web-ui/src/components/notes/Note.css
```diff
@@ -26,3 +27,20 @@
...
+    lightgray 25px
+  );
+  border: 2px solid black;
+  min-height: 150px;
+  padding: 5px;
+  width: 100%;
+  height: fit-content;
+  top: 0;
+  left: 0;
+}
```
mkimberlin, line 46: We should probably make use of `<Skeleton />`

web-ui/src/components/notes/Note.jsx
```diff
@@ -1,77 +1,90 @@
...
 import { AppContext } from "../../context/AppContext";
+import Snackbar from "@material-ui/core/Snackbar";
+import MuiAlert from "@material-ui/lab/Alert";
 
 import "./Note.css";
 
+async function realUpdate(note) {
+  let res = await updateCheckinNote(note);
+  if (res.error) {
+    console.error(res.error);
```
mkimberlin, line 19: so...if `resolve()` handles the errors...what good is this?

web-ui/src/components/notes/Note.jsx
mkimberlin: This makes me happy.

web-ui/src/components/notes/Note.jsx
```diff
@@ -1,77 +1,90 @@
...
+}
+
 const Notes = (props) => {
   const { state } = useContext(AppContext);
   const { userProfile } = state;
   const canViewPrivateNote =
-    userProfile.role.includes("PDL") || userProfile.role.includes("ADMIN");
+    userProfile.memberProfile &&
+    userProfile.memberProfile.role &&
+    (userProfile.memberProfile.role.includes("PDL") ||
```
mkimberlin, line 35: This is insufficient. Hopefully the backend will save us, but the private note section shouldn't show up if I'm a PDL and I'm looking at my own check-in. With just this check, it would.

web-ui/src/components/notes/Note.jsx
```diff
@@ -86,12 +99,25 @@ const Notes = (props) => {
           Notes for {memberName}
         </h1>
         <div className="container">
-          <textarea
-            onChange={handleNoteChange}
-            value={note.description}
-          ></textarea>
-          <canvas ref={canvasRef}></canvas>
+          {isLoading ? (
+            <div className="is-loading"></div>
```
mkimberlin, line 103:
```
import Skeleton from '@material-ui/lab/Skeleton';

<Skeleton />
```

web-ui/src/components/notes/Note.jsx
```diff
@@ -86,12 +99,25 @@ const Notes = (props) => {
...
+        <Snackbar
+          autoHideDuration={2500}
+          open={open}
+          onClose={handleClose}
+          style={{ left: "40%", bottom: "20%" }}
+        >
+          <Alert onClose={handleClose} severity="error">
+            {toast}
+          </Alert>
+        </Snackbar>
```
mkimberlin, line 120: Do we not already have a top level snackbar? If not, we should probably make one and have a means of enqueueing toast to it via the context.

web-ui/src/context/AppContext.js
mkimberlin: ok, wait...what? let's put in a ticket the service to not respond with that weirdness.

web-ui/src/context/AppContext.js
```diff
@@ -55,26 +63,65 @@ const AppContextProvider = (props) => {
...
+        if (data && data.length > 0) {
+          const allComplete = data.every(
+            (checkin) => checkin.completed === true
+          );
+          if (allComplete) {
+            const prevCheckinDate = data[data.length - 1].checkInDate;
+            const res = await createCheckin({
+              teamMemberId: id,
+              pdlId: pdlId,
+              checkInDate: date(3, prevCheckinDate),
```
mkimberlin, line 96: What if the previous checkin date was 6 months ago? I'd say if it's more than 3 months ago, schedule it for a month out like we do if there is no check-in.

web-ui/src/pages/CheckinsPage.jsx
```diff
@@ -1,22 +1,30 @@
...
   const { state } = useContext(AppContext);
-  const { checkins, userProfile } = state;
-  const [index, setIndex] = useState(0);
+  const { checkins, index, userProfile } = state;
   const checkin = checkins[index];
   const canSeePersonnel = userProfile && userProfile.role.includes("PDL");
 
+  useEffect(() => {
+    if (checkin && checkin.id) {
+      history.push(`/checkins/${checkin.id}`);
```
mkimberlin, line 24: Ok...I think this may be a little bit backward. I think we should be pulling the checkin id off of the URL when it's there and respecting that. The history should control what the current check-in is by changing the URL. Otherwise, if I hit refresh while looking at a given check-in, I will likely lose what I was looking at and be shown the default.

meyerinj (approved):

web-ui/src/components/avatar/Avatar.jsx
```diff
@@ -1,10 +1,9 @@
 import React from "react";
 import Avatar from "@material-ui/core/Avatar";
-import PersonIcon from "@material-ui/icons/Person";
 
 const AvatarComponent = ({ loggedIn = false, profile = {} }) => {
   const { image_url } = profile;
-  const src = !image_url || loggedIn === false ? <PersonIcon /> : image_url;
+  const src = !image_url || loggedIn === false ? "" : image_url;
```
mkimberlin, line 6: 👍

web-ui/src/components/menu/Menu.jsx
```diff
@@ -118,7 +113,7 @@ function Menu() {
       </Button>
       <br />
       <Button>
-        <Link style={linkStyle} to="/checkins">
+        <Link style={linkStyle} to={`/checkins`}>
```
mkimberlin, line 116: wha?

web-ui/src/components/notes/Note.jsx
```diff
@@ -86,11 +82,19 @@ const Notes = (props) => {
...
-            onChange={handleNoteChange}
-            value={note.description}
-          ></textarea>
-          <canvas ref={canvasRef}></canvas>
+          {isLoading ? (
+            <div className="skeleton">
+              <Skeleton variant="text" height={30} />
+              <Skeleton variant="text" height={30} />
+              <Skeleton variant="text" height={30} />
+              <Skeleton variant="text" height={30} />
```
mkimberlin, line 90: you might try a relative value like `1 rem` or whatever seems at the right scale. They tend to make for a better experience on a variety of devices.

mkimberlin (approved):

web-ui/src/components/menu/Menu.jsx
jahanner: Left over from when I was testing something lol

PR #283, opened 2020-09-02
Add validations to CheckinController

Bagurp:
* Changes 'completed' parameter in checkin from Nullable -> NotNull
* Update is only allowed if the check in is not completed unless made by admin
* Limit creation/read/updation to -
[] subject of the check in
[] the current PDL of the subject of the check in
[] admin
* Convert LocalDate to LocalDateTime for checkins service
* Add DB script to convert column type to time for Checkins table
* Issue - https://github.com/oci-labs/check-ins/issues/245


server/src/main/java/com/objectcomputing/checkins/services/checkins/CheckInController.java
```diff
@@ -47,17 +58,27 @@
...
      * Find Check-in details by Member Id or PDL Id. 
      * @param teamMemberId
      * @param pdlId
      * @return
      */
     @Get("/{?teamMemberId,pdlId,completed}")
-    public Set<CheckIn> findByValue(@Nullable UUID teamMemberId,
-                                     @Nullable UUID  pdlId, @Nullable Boolean completed) {          
-            return checkInservices.findByFields(teamMemberId, pdlId,completed);
+    public Set<CheckIn> findByValue(@Nullable UUID teamMemberId, @Nullable UUID  pdlId, @Nullable Boolean completed) {
```
Bagurp, line 68: These validations are expensive since DB call is made prior to validating the access of the current user. Could anyone think of a better way to do this?

server/src/main/java/com/objectcomputing/checkins/services/checkins/CheckInController.java
```diff
@@ -95,8 +131,15 @@
...
     @Get("/{id}")
     public CheckIn readCheckIn(@NotNull UUID id){
-        return checkInservices.read(id);
-    }
 
+        MemberProfile currentUser = currentUserServices.currentUserDetails();
+        Boolean isAdmin = securityService.hasRole(RoleType.Constants.ADMIN_ROLE);
+        CheckIn checkin = checkInservices.read(id);
 
+        if(currentUser.getUuid().equals(checkin.getTeamMemberId()) || currentUser.getUuid().equals(checkin.getPdlId()) || isAdmin) {
```
Bagurp, line 139: Same here. Expensive validation

Maroju100:
POST - Checkin creation works even if member and PDL are not associated. 
Right now multiple checkins seems to be created (for multiple POSTs ) without any warning, not sure if that is expected behavior.
PUT-checkins looks to be getting updated my member even when submitted is true.
GET-Multiple checkins entries are shows since multiple POSTs are possible.

Bagurp:
@Maroju100 Good catch. I've added validations to impl class to both save/update. I also updated formatting. (Hide whitespace on github for better PR experience)

server/src/main/java/com/objectcomputing/checkins/services/checkins/CheckInController.java
meyerinj: The @Secured annotation, when used on a method, can be used to specify a role.
https://micronaut-projects.github.io/micronaut-security/latest/guide/#secured

server/src/main/java/com/objectcomputing/checkins/services/checkins/CheckInController.java
```diff
@@ -38,6 +43,12 @@
     @Inject
     CheckInServices checkInservices ;
 
+    @Inject
+    CurrentUserServicesImpl currentUserServices;
```
meyerinj, line 47: Not specific to this story, but a better practice is to make these fields private and inject them via the constructor.

server/src/main/java/com/objectcomputing/checkins/services/checkins/CheckInServicesImpl.java
```diff
@@ -18,62 +20,80 @@
 
     @Inject
     private MemberProfileRepository memberRepo;
```
meyerinj, line 22: Private field injection is also bad. This can also be resolved by using constructor injection.

meyerinj: I should add that the class must be annotated with @Singleton to use constructor injection

server/src/main/java/com/objectcomputing/checkins/services/checkins/CheckInServicesImpl.java
```diff
@@ -18,62 +20,80 @@
...
+
+    @Inject
+    SecurityService securityService;
+
     @Override
     public CheckIn save(CheckIn checkIn) {
         CheckIn checkInRet = null;
-        if(checkIn!=null){
+
+        if(checkIn!=null) {
```
meyerinj, line 31: Is there a scenario where we want the client to be able to attempt to save nothing and simply allow that to pass? Would it make more sense to use the @NotNull annotation and simply not allow this field to be null?

server/src/main/java/com/objectcomputing/checkins/services/checkins/CheckInServicesImpl.java
```diff
@@ -18,62 +20,80 @@
...
-        if(id==null||!checkinRepo.findById(id).isPresent()){
-            throw new CheckInBadArgException(String.format("Unable to find checkin record with id %s", checkIn.getId()));
-        }else if(!memberRepo.findById(memberId).isPresent()){
-            throw new CheckInBadArgException(String.format("Member %s doesn't exist", memberId));
-        } else if(memberId==null) {
-            throw new CheckInBadArgException(String.format("Invalid checkin %s", checkIn));
-        } else if(chkInDate.isBefore(LocalDate.EPOCH) || chkInDate.isAfter(LocalDate.MAX)) {
-            throw new CheckInBadArgException(String.format("Invalid date for checkin %s",memberId));
+
+        if(checkIn!=null) {
```
meyerinj, line 65: As with the save method would it make sense to simply not allow this to be null?

server/src/main/java/com/objectcomputing/checkins/services/memberprofile/currentuser/CurrentUserServicesImpl.java
```diff
@@ -24,4 +30,15 @@ public MemberProfile findOrSaveUser(@NotNull String name, @NotNull String workEm
                 "", workEmail, "", null, "")));
 
     }
+
+    public MemberProfile currentUserDetails() {
```
meyerinj, line 34: When doing authentication it is probably easier to include a Principal argument in the controller than to use the security service.
https://micronaut-projects.github.io/micronaut-security/latest/guide/#retrievingAuthenticatedUser

meyerinj (changes requested):

server/src/main/java/com/objectcomputing/checkins/services/checkins/CheckInController.java
bhatiam12: that will not help in this situation, right ? Because we are validating if a current user is a PDL of the subject that he is doing a check in and not if he has a PDL role because many people will have PDL role but need to make sure they do the checkin for their team member.

server/src/main/java/com/objectcomputing/checkins/services/checkins/CheckInServicesImpl.java
Bagurp: Good catch. I'll make the required changes.

server/src/main/java/com/objectcomputing/checkins/services/checkins/CheckInServicesImpl.java
Bagurp: I'll make the required changes.

server/src/main/java/com/objectcomputing/checkins/services/checkins/CheckInController.java
meyerinj: It will replace the securityService call.

As for the rest there is another option but whether it is better or not is debatable.

We can add claims to our user details mapper that will contain the user's PDL's ID and (if the user is a PDL) any team members for whom the user is a PDL. You would be able to check that against the request body instead of having to go through the user service.

server/src/main/java/com/objectcomputing/checkins/services/checkins/CheckInBadArgException.java
```diff
@@ -1,11 +1,8 @@
 package com.objectcomputing.checkins.services.checkins;
 
 public class CheckInBadArgException extends RuntimeException{
-     /**
-     *
-     */
-    private static final long serialVersionUID = 1L;
 
+    private static final long serialVersionUID = 1L;
```
mkimberlin, line 5: I don't believe in this. You have offended my religion and we must now duel to the death. If you're not willing to face me, you may substitute a champion or generate a proper serialVersionUID.

server/src/main/java/com/objectcomputing/checkins/services/checkins/CheckInController.java
mkimberlin: No, it won't help us here. This validation has to be done by hand.

server/src/main/java/com/objectcomputing/checkins/services/checkins/CheckInController.java
mkimberlin: We need to allow the current PDL as well. There needs to be a case added for this.

server/src/main/java/com/objectcomputing/checkins/services/checkins/CheckInController.java
mkimberlin: Also...why are you injecting the implementation rather than the interface here?

server/src/main/java/com/objectcomputing/checkins/services/checkins/CheckInController.java
Bagurp: currentUser.getUuid().equals(checkin.getPdlId()) refers to current PDL.

server/src/main/java/com/objectcomputing/checkins/services/checkins/CheckInServicesImpl.java
mkimberlin: Yeah, I would change all your dependency fields to private and inject them via constructor.

server/src/main/java/com/objectcomputing/checkins/services/checkins/CheckInController.java
Bagurp: good catch. I'll update it.

server/src/main/java/com/objectcomputing/checkins/services/checkins/CheckInBadArgException.java
Bagurp: pardon, my lord - for I am just a humble developer cleaning unused comments. I choose to clean the code over facing your wrath ;)

server/src/main/java/com/objectcomputing/checkins/services/checkins/CheckInController.java
meyerinj: I made a branch, [here](https://github.com/oci-labs/check-ins/tree/feature-245/also-with-claim-validation), which has a really hacky way to do that. If I understand what y'all are trying to do we should be able to use something like that.

Bagurp:
* If you are trying to run tests on IDE - run it with `MICRONAUT ENVIRONMENTS = test` so that it uses the right test config file.
* Please hide Whitespace on github for better PR review experience.

bhatiam12 (approved):

server/src/main/java/com/objectcomputing/checkins/services/checkins/CheckInServicesImpl.java
```diff
@@ -1,86 +1,167 @@
...
+        Boolean isAdmin = securityService!=null ? securityService.hasRole(RoleType.Constants.ADMIN_ROLE) : false;
+
+        CheckIn result = checkinRepo.findById(id).orElse(null);
+
+        if (result == null) {
+            throw new CheckInBadArgException(String.format("Invalid checkin id %s", id));
+        } else if(!isAdmin) {
+            if(!currentUser.getId().equals(result.getTeamMemberId()) && !currentUser.getId().equals(result.getPdlId())) {
+                // Limit read to Subject of check-in, PDL of subject and Admin
+                throw new CheckInBadArgException(String.format("Member %s is unauthorized to do this operation", currentUser.getId()));
```
mkimberlin, line 79: The ID isn't very meaningful to a user. I would nix the id here and just say something like "You are not authorized to perform this operation". It probably wouldn't hurt for us to have some AOP based logging that logs any thrown exceptions and records things like who the user was the method name and what the inputs to the method were.

mkimberlin: To be clear, that wasn't intended to indicate you should implement AOP based logging in this PR.  Just a thought about something we might want to consider adding.

server/src/main/java/com/objectcomputing/checkins/services/checkins/CheckInServicesImpl.java
```diff
@@ -1,86 +1,167 @@
...
+        } else if(chkInDate.isBefore(Util.MIN) || chkInDate.isAfter(Util.MAX)) {
             throw new CheckInBadArgException(String.format("Invalid date for checkin %s",memberId));
+        } else if(!isAdmin) {
+            if(!currentUser.getId().equals(checkIn.getTeamMemberId()) && !currentUser.getId().equals(checkIn.getPdlId())) {
+                // Limit update to subject of check-in, PDL of subject and Admin
+                throw new CheckInBadArgException(String.format("Member %s is unauthorized to do this operation", currentUser.getId()));
+            } else if(checkinRepo.findById(id).get().isCompleted()) {
+                // Update is only allowed if the check in is not completed unless made by admin
+                throw new CheckInBadArgException(String.format("Checkin with id %s is complete and cannot be updated", checkIn.getId()));
+            }
```
mkimberlin, line 115: I feel like this could be made more concise by defining a method that would take care of the `if...throw` bit maybe?

It could take a lambda, a message, and an optional set of replacement values...or even a boolean instead of a lambda.  There's no need for the elses since an exception being thrown...so it could just be a sequence of validation calls. Just a thought.

mkimberlin (approved):
couple of minor comments...nothing that I would block this proceeding though.

server/src/main/java/com/objectcomputing/checkins/services/checkins/CheckInServicesImpl.java
Bagurp: Sounds good, I've created a new story to implement AOP based logging. For the current PR, I've updated error messages.

New Issue - https://github.com/oci-labs/check-ins/issues/309

server/src/main/java/com/objectcomputing/checkins/services/checkins/CheckInController.java
```diff
@@ -33,10 +33,14 @@
 @Produces(MediaType.APPLICATION_JSON)
 @Consumes(MediaType.APPLICATION_JSON)
 @Tag(name="check-ins")
+@Singleton
```
meyerinj, line 36: Anything annotated with @Controller can also be treated as a Singleton so using both annotations is unnecessary.

meyerinj, line 36: Anything annotated with @controller can also be treated as a Singleton so using both annotations is unnecessary.

meyerinj (changes requested):

server/src/main/java/com/objectcomputing/checkins/services/checkins/CheckInController.java
Bagurp: Updated. I've removed @Singleton annotation from controller.

server/src/main/java/com/objectcomputing/checkins/services/checkins/CheckInController.java
mkimberlin: I'm actually not against the idea of keeping a user's assigned team members in their user details...but I think we should make a separate story to investigate that.

server/src/main/java/com/objectcomputing/checkins/services/checkins/CheckInServicesImpl.java
Bagurp: I've added a private method called validate in checkinServiceImpl in the following commit - https://github.com/oci-labs/check-ins/pull/283/commits/7437ac3545ee2026d01ec738582b3e7ee7c20a08

meyerinj (approved):

Issue #268, opened 2020-08-28
Convert Action Items service and repository to use query annotation


meyerinj:
Definition of Done

- [ ] ActionItemServicesImpl.findByFields should be able to make only one repository call to find any records that match the parameters. See #222 for examples

PR #287, opened 2020-09-03
Action item query annotation

meyerinj:
#268 

Implementing @bhatiam12 's suggested casting. It seems to require the null handling casts you see in the repository

### DIFF
--- a/server/src/main/java/com/objectcomputing/checkins/services/action_item/ActionItemRepository.java
+++ b/server/src/main/java/com/objectcomputing/checkins/services/action_item/ActionItemRepository.java
@@ -1,12 +1,12 @@
 package com.objectcomputing.checkins.services.action_item;
 
-import edu.umd.cs.findbugs.annotations.NonNull;
-import edu.umd.cs.findbugs.annotations.Nullable;
 import io.micronaut.data.annotation.Query;
 import io.micronaut.data.jdbc.annotation.JdbcRepository;
 import io.micronaut.data.model.query.builder.sql.Dialect;
 import io.micronaut.data.repository.CrudRepository;
+import io.reactivex.annotations.NonNull;
 
+import javax.annotation.Nullable;
 import javax.validation.Valid;
 import javax.validation.constraints.NotNull;
 import java.util.List;
--- a/server/src/main/java/com/objectcomputing/checkins/services/action_item/ActionItemRepository.java
+++ b/server/src/main/java/com/objectcomputing/checkins/services/action_item/ActionItemRepository.java
@@ -9,8 +9,8 @@ import io.micronaut.data.repository.CrudRepository;
 
 import javax.validation.Valid;
 import javax.validation.constraints.NotNull;
-import javax.validation.constraints.Null;
 import java.util.List;
+import java.util.Set;
 import java.util.UUID;
 
 @JdbcRepository(dialect = Dialect.POSTGRES)
@@ -24,7 +24,7 @@ public interface ActionItemRepository extends CrudRepository<ActionItem, UUID> {
             "FROM action_items item " +
             "WHERE (cast(:checkinId as uuid) IS NULL OR item.checkinid = :checkinId) " +
             "AND (cast(:createdById as uuid) IS NULL OR item.createdbyid = :createdById)")
-    List<ActionItem> search(@Nullable UUID checkinId, @Nullable UUID createdById);
+    Set<ActionItem> search(@Nullable UUID checkinId, @Nullable UUID createdById);
 
     @Override
     <S extends ActionItem> List<S> saveAll(@Valid @NotNull Iterable<S> entities);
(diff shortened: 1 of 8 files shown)


bhatiam12:
@meyerinj all the tests are failing.

server/src/main/java/com/objectcomputing/checkins/services/action_item/ActionItemRepository.java
```diff
@@ -17,6 +20,12 @@
 
     List<ActionItem> findByCreatedbyid(UUID uuid);
 
+    @Query("SELECT * " +
+            "FROM action_items item " +
+            "WHERE (cast(:checkinId as uuid) IS NULL OR item.checkinid = :checkinId) " +
+            "AND (cast(:createdById as uuid) IS NULL OR item.createdbyid = :createdById)")
```
mkimberlin, line 26: Are you casting the right thing here? This looks backward to me. Don't you need to coerce the parameter to text (or the column to uuid) on the comparison? This looks like you are casting the already a UUID thing to a UUID on the null check.

server/src/main/resources/db/V26__add_uuid_cast.sql
```diff
@@ -0,0 +1,5 @@
+DROP CAST IF EXISTS (varchar AS uuid);
+
+CREATE CAST (varchar AS uuid)
+  WITH INOUT
+  AS IMPLICIT;
```
mkimberlin, line 5: And if you are casting on the query...you don't need this migration, right?

I have to admit...while this migration might make some things easier, It feels like we are hiding an issue from future contributors to this project.

bhatiam12: I tried casting on the query and it throws an error that explicit cast is needed. This is the only way I was able to make it work.

server/src/main/java/com/objectcomputing/checkins/services/action_item/ActionItemRepository.java
meyerinj: We have the implicit cast of uuid to varchar that is added in the flyway script, but the SQL doesn't know what to do with `NULL` when we have that. The explicit cast is needed for the null case. If you take those away then the query will work as expected when you supply a UUID (because we have an implicit cast) but blow up if you leave them null.

bhatiam12: actually I don't think so that we need this cast if we are using explicit cast in the migration.

bhatiam12: > We have the implicit cast of uuid to varchar that is added in the flyway script, but the SQL doesn't know what to do with `NULL` when we have that. The explicit cast is needed for the null case

Ahh then that make sense and that was my issue. I was not able to run a test for NULL case.

server/src/main/resources/db/V26__add_uuid_cast.sql
mkimberlin: how did you try casting on the query? what did that query look like exactly?

bhatiam12: I think I tried something like `WHERE (:checkinId as text IS NULL OR item.checkinid = :checkinId)`

meyerinj: The way I got it to work on the draft PR was to simply use a String when making the database call. Open debate to be had on if using the String is better or worse than having to explicitly cast the null case in the query.

bhatiam12 (approved):

server/src/main/java/com/objectcomputing/checkins/services/action_item/ActionItemRepository.java
```diff
@@ -1,13 +1,16 @@
 package com.objectcomputing.checkins.services.action_item;
 
 import edu.umd.cs.findbugs.annotations.NonNull;
+import edu.umd.cs.findbugs.annotations.Nullable;
```
Bagurp, line 4: on lines 3-4, can we use javax.validation.constraints.* instead of edu.umd.cs?

server/src/main/java/com/objectcomputing/checkins/services/action_item/ActionItemRepository.java
```diff
@@ -17,6 +20,12 @@
 
     List<ActionItem> findByCreatedbyid(UUID uuid);
```
Bagurp, line 21: is it safe to remove findByCheckinid, findByCreatedbyid?

Bagurp (approved):

server/src/main/java/com/objectcomputing/checkins/services/action_item/ActionItemRepository.java
meyerinj: Ooh, good catch, thanks!

server/src/main/java/com/objectcomputing/checkins/services/action_item/ActionItemRepository.java
meyerinj: I was going to keep these because the search is going to get ordering in my next story, but on further thought I can't come up with a use case for keeping these non-ordered methods.

More importantly, removing these exposed that some of the unit tests needed further alteration. They were passing fine, they were just kidna funky.

PR #295, opened 2020-09-09
Can only see personnel if user is pdl Closes #292

### LINKED ISSUE #292: FE > Only display My Personnel panel if the user is a PDL
On the check-ins page, this panel should only show if the current user is a PDL.

jahanner:


web-ui/src/pages/CheckinsPage.jsx
```diff
@@ -13,8 +13,10 @@ const CheckinsPage = () => {
   const [show, setShow] = useState(false);
   const { state } = useContext(AppContext);
   const { checkins, userProfile } = state;
+  console.log({ userProfile });
   const [index, setIndex] = useState(0);
   const checkin = checkins[index];
+  const canSeePersonnel = userProfile && userProfile.role.includes("PDL");
```
bhatiam12, line 19: can ADMIN also view it ? do we need another check for ADMIN role ?

jahanner: The story said only if the user is a PDL, but maybe we do want admin to see it too? @mkimberlin What do you think?

mkimberlin: Only PDLs. If an ADMIN is also a PDL, yes... If not, then it doesn't make sense for them to see it.

web-ui/src/pages/CheckinsPage.jsx
```diff
@@ -13,8 +13,10 @@ const CheckinsPage = () => {
   const [show, setShow] = useState(false);
   const { state } = useContext(AppContext);
   const { checkins, userProfile } = state;
+  console.log({ userProfile });
```
mkimberlin, line 16: I'd remove this line.

mkimberlin (approved):
Kill that `console.log` and then you are good.

web-ui/src/pages/CheckinsPage.jsx
jahanner: Mille gratzie

Issue #197, opened 2020-08-12
FE > Create action items component (L)


mkimberlin:
Create a component that allows for the management of a Check-In's action items.

Definition of Done

- [ ] Component is created similar to this mockup but with (-) icon on the right for deletion and vertical ellipsis on the left for reordering on each action item:
<img width="700" alt="Screen Shot 2020-08-12 at 12 05 14 PM" src="https://user-images.githubusercontent.com/100510/90065470-f2db4580-dcb1-11ea-88cb-b336a95ac81a.png">

- [ ] Component accepts a check-in id and retrieves the action items for the check-in for display
- [ ] Component displays each action item once loaded 
- [ ] Component appears similar to this while loading:
![Action Items - Loading](https://user-images.githubusercontent.com/100510/90066867-ee179100-dcb3-11ea-976a-fd832d6e2759.png)

- [ ] Component updates action items on change (via REST service)
- [ ] Component allows for deletion of action items (via REST service)
- [ ] Component allows for the addition of action items (via REST service)
- [ ] Component allows for the reordering of action items (via REST service)
- [ ] Snapshot test(s) have been created for the component
- [ ] Applicable unit tests have been created for the component
- [ ] Code has been submitted for PR and accepted

meyerinj:
Couple Qs:

Is this component intended to go on the side bar of the check-ins page, similar to the guides component Zack and I worked on? Or its own page?

There is a requirement to be able to reorder the list. What should they be ordered by by default?

meyerinj:
Additional question: It's not clear what data is supposed to be displayed for each record. There's a description field obvi, but there are also UUIDs for what I assume are a CheckIn and TeamMember object. What data from them is needed?

mkimberlin:
> Is this component intended to go on the side bar of the check-ins page, similar to the guides component Zack and I worked on? Or its own page?

It is a section in the main content of the check-ins page.

> There is a requirement to be able to reorder the list. What should they be ordered by by default? 

Reorder is part of the DoD above. Natural ordering (by creation time).

meyerinj:
> Reorder is part of the DoD above. Natural ordering (by creation time).

The requirement also says the order should be persisted via a REST service. Obviously we can't change the creation time when we reorder the records.

mkimberlin:
Only the initial ordering should be natural...that's really a matter of just putting new ones at the bottom. We will need to persist an order. Sorry...didn't mean to suggest that we would control it by creation time.

[...]

PR #296, opened 2020-09-09
Feature 197/action items component

meyerinj:
#197 

Not sure I understood the requirements 100%, for example it's unclear where/how this component should be used, but the component itself seems to work.

Since I know someone will ask the return type of the repository search was changed because Micronaut Data uses HashSet, which is not ordered.


web-ui/demo/.gitignore
```diff
@@ -0,0 +1,23 @@
+# See https://help.github.com/articles/ignoring-files/ for more about ignoring files.
```
jahanner, line 1: I'm not sure I understand why we have this new demo section?

web-ui/demo/src/index.js
```diff
@@ -0,0 +1,96 @@
...
+    ...draggableStyle
+});
+
+const getListStyle = isDraggingOver => ({
+  background: isDraggingOver ? "lightblue" : "lightgrey",
+  padding: grid,
+  width: 250
+});
+
+class App extends Component {
```
jahanner, line 36: This probably isn't a big deal but the code to be more uniform, we may want to use functional components and hooks here

web-ui/src/components/action_item/ActionItemsPanel.jsx
```diff
@@ -0,0 +1,237 @@
...
+    if (actionItem) {
+      await updateActionItem(actionItem);
+    }
+  }
+
+  let [prevActionItems, setPrevActionItems] = useState();
+
+  const actionItemsCompare = currItems => {
+    if (!prevActionItems) {
+      console.log('no previous');
```
jahanner, line 30: We should probably get rid of these console.logs

web-ui/src/components/action_item/ActionItemsPanel.jsx
```diff
@@ -0,0 +1,237 @@
...
+    display: 'flex',
+    flexDirection: 'row',
+
+    background: isDragging ? 'lightgreen' : '#fafafa',
+
+    ...draggableStyle
+  });
+
+  const onDragEnd = result => {
+    console.log('onDragEnd: result =', result);
```
jahanner, line 111: and here

web-ui/src/components/action_item/ActionItemsPanel.jsx
```diff
@@ -0,0 +1,237 @@
...
+
+    actionItems[result.destination.index].priority = newPriority;
+
+    doUpdate(actionItems[result.destination.index]);
+    getActionItems();
+  };
+
+  const killActionItem = (id, event) => {
+    doDelete(id);
+    console.log('BEFORE');
```
jahanner, line 142: and these

Checkins-setup.postman_collection.json
```diff
@@ -0,0 +1,502 @@
+{
```
mkimberlin, line 1: Was this file an intentional addition?

package.json
```diff
@@ -0,0 +1,7 @@
+{
```
mkimberlin, line 1: This shouldn't exist at the top level. These dependency additions belong in the package.json that exists in the web-ui project.

package-lock.json
```diff
@@ -0,0 +1,200 @@
+{
```
mkimberlin, line 1: This file also shouldn't exist at the top level and is a result of the top level package.json.  Also, we use yarn not NPM on this project. So package-lock files should never be created.

server/src/main/java/com/objectcomputing/checkins/services/action_item/ActionItem.java
```diff
@@ -42,15 +42,29 @@
     @Schema(description = "description of the action item")
     private String description;
 
+    @Column(name = "priority")
+    @TypeDef(type = DataType.DOUBLE)
```
mkimberlin, line 46: This was necessary? It didn't default to this type?

server/src/main/resources/secrets/directory.json
```diff
@@ -0,0 +1,3 @@
+{
```
mkimberlin, line 1: Just an FYI, this is changing to be a property that's passed in...so this file will no longer be necessary.

web-ui/demo/.gitignore
mkimberlin: I'm with Jesse on this. It seems like it's here to show how to use the DnD stuff. I think that would be better put into the Storybook stories (just make a .stories.jsx file for it...see: https://storybook.js.org/docs/react/writing-stories/introduction)

yarn.lock
```diff
@@ -0,0 +1,107 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
mkimberlin, line 1: This also shouldn't exist at the root directory level

server/src/main/resources/secrets/directory.json
meyerinj: Should that change as part of this story or just leave this to be changed when that happens?

web-ui/package-lock.json
```diff
@@ -0,0 +1,16581 @@
+{
```
jahanner, line 1: I think this file may need to be removed

meyerinj: Should it be added to the .gitignore?

jahanner: Just deleted actually, we haven't had a package-lock thus far since we've been using yarn and I think Michael mentioned using yarn and npm like this could cause issues. Is that right @mkimberlin ?

server/src/main/resources/secrets/directory.json
mkimberlin: not this story

web-ui/package-lock.json
mkimberlin: yep...should just be using yarn and this won't ever get generated

web-ui/src/pages/CheckinsPage.jsx
```diff
@@ -46,6 +47,7 @@ const CheckinsPage = ({ history }) => {
               </div>
               <CheckinDocs />
               <div className="modal-container">
+                <ActionItemsPanel checkinId="9636fdaa-75cd-430e-84d8-1efea999682a" />
```
mkimberlin, line 50: I think we want this above checkindocs and not in this modal-container

mkimberlin (approved):
I'd move the location of the component on the checkins page...approving, but I haven't exercised this code locally yet, so I would encourage someone to do that (sorry to bum that off on someone else)

jahanner (approved):

Issue #225, opened 2020-08-24
SPIKE > BE > Look into using flyway.locations to allow for loading test data for our local and test environments (M)


mkimberlin:
See:  https://flywaydb.org/documentation/faq#db-specific-sql

I believe we would want to use a repeatable migration script so that we could reset our dev database regularly:

https://flywaydb.org/getstarted/repeatable

Definition of Done:
- [ ] Determine the feasibility of such an approach
- [ ] Define a story to implement test data refresh process

PR #297, opened 2020-09-09
Repeatable scripts spike

meyerinj:
#225 

Here's the findings:

I have added three files

`R__repeatable_script_creates_table.sql` creates a table and has it depend on member_profile to demonstrate execution order
`R_ZZZrepeatable_script_another_table.sql` will not execute without adding a second _. It will cause an error as written. See below for more info.
`V_27__Dependent_on_repeatable` will not execute without removing the first _. If it executes it will cause an error. This is to demonstrate that versioned files always run before repeatables. The error in execution is expected.

Repeatable scripts allow us to change the CREATE script for a table (or other relation) rather than adding a versioned script with an ALTER command.

All repeatable scripts run after all versioned scripts, so a versioned script cannot reference something in a repeatable script.

According to the documentation if multiple repeatable scripts exist they will run in the order implied by their description (the part after `R__`) however my experimentation with it shows this is unreliable and scripts that should execute later are unable to find a dependency in a script that should be executed earlier. You can test this out by enabling `R_ZZZ_repeatable_script_another_table.sql`. Although repeatables go in the order of their description, and the "ZZZ" means this should have a later description, in my testing it does not.

My takeaway:

I do not believe repeatable scripts are useful as a replacement for the versioned scripts. Creating multiple scripts (e.g. one per table) would cause unreliable behavior among tables that reference each other. This could be resolved by putting the entire migration in a single script, but this would not be very readable.

Repeatable scripts would be useful for scripts that will never be interdependent. If we end up implementing some DML, for example, a repeatable script would be a useful way to do it.


meyerinj:
@mkimberlin 

Split scripts between "db\common" and "db/local" directories. The application yaml field `flyway.datasources.default.locations` can then take a collection of all relevant directories.

It appears that in the background Flyway treats all these directories as one. This means if we are using both "db\common" and "db\local" they cannot both contain a versioned script with the same version number but if "db\local" and, say, "db\prod" will never be used together they can both contain a V2__ script.

As with a single directory all versioned scripts from all locations will run before all repeatable scripts from all locations.

This will serve our needs of having DML scripts that run only locally but not in prod.

Issue #203, opened 2020-08-13
CHORE > Convert to OCI SonarQube installation (M)


mkimberlin:
See Adam Rider for additional information. Sonar scanning should be happening as a part of the build (and used to, I thought). This needs be put in place and integrated into our PR checks as well.

meyerinj:
Looks like there's an issue in Sonar config and the Sonar guy is out until Friday so this is blocked for a few days.

PR #300, opened 2020-09-11
Chore 203/sonarqube integration

meyerinj:
#203 
SonarQube will now be run on push to a new branch. Findings can be viewed here: https://sonarqube.cloudops.objectcomputing.com/dashboard?id=oci-checkins

Developers can see these findings as a change to correct at time of push.

Alternatively this can be done on merge to develop in which case BAs would make stories to address findings when appropriate.


server/src/main/resources/application.yml
```diff
@@ -70,4 +70,4 @@ flyway:
   enabled: enabled
   datasources:
     default:
-      locations: classpath:db
+      locations: classpath:db/common
```
bhatiam12, line 73: why did we change the class path ?

meyerinj: This isn't a change to the classpath; it's a change to where on the classpath we look for scripts. This is done so we can split scripts into different directories under db to control which get run under what circumstance.

meyerinj: ...But should not be part of this story. A different branch leaked onto this one.

bhatiam12: I guess that was more of my concern that why it was changed as part of this story :)

Maroju100 (approved):

.github/workflows/gradle-build-feature.yml
```diff
@@ -19,5 +19,10 @@ jobs:
...
-      - name: Gradle runs tests
-        run: ./gradlew check
\ No newline at end of file
+      - name: Do SonarQube checks
+        run: ./gradlew check jacocoTestReport sonarqube -Dsonar.projectKey=$SONAR_PROJECT_KEY -Dsonar.host.url=$SONAR_URL -Dsonar.login=$TOKEN
+        env:
+          SONAR_PROJECT_KEY: oci-labs_oci-checkins
+          SONAR_ORG: oci-labs
+          SONAR_URL: https://sonarcloud.io
+          TOKEN: 76d54e72c4fbab6a5d7928cc95e2e9248986b1c5
```
bhatiam12, line 28: should this token be stored in secrets ?

web-ui/src/pages/CheckinsPage.test.js
```diff
@@ -1,11 +1,3 @@
...
 import CheckinsPage from "./CheckinsPage";
 import { AppContextProvider } from "../context/AppContext";
-
-it("renders correctly", () => {
-    <AppContextProvider>
-        snapshot(
-        <CheckinsPage />
-        );
-    </AppContextProvider>;
-});
```
mkimberlin, line 11: How'd this get in here?

meyerinj: Sonar was complaining about it and I was hoping to force the workflow to succeed by commenting it out. Then I forgot to undo it before I pushed.

mkimberlin (approved):

PR #303, opened 2020-09-14
Feature 269/agenda query annotation

Maroju100:
Converted agenda items service and repository to use query annotation


server/src/main/java/com/objectcomputing/checkins/services/agenda_item/AgendaItemServicesImpl.java
```diff
@@ -65,15 +67,8 @@ public AgendaItem update(AgendaItem agendaItem) {
     }
 
     public Set<AgendaItem> findByFields(UUID checkinid, UUID createdbyid) {
```
meyerinj, line 69: Not caused by this story specifically, but it would be worth putting a nullable annotation on these.

meyerinj (approved):

server/src/main/java/com/objectcomputing/checkins/services/agenda_item/AgendaItemServicesImpl.java
Maroju100: updated, thanks

hwilliams314 (approved):

PR #304, opened 2020-09-14
Feature 265/convert guild member to query closes #265

### LINKED ISSUE #265: Convert guild member service and repository to use query annotation
Definition of Done

- [x] GuildMemberServicesImpl.findByFields should be able to make only one repository call to find any records that match the parameters. See #222 for examples

hwilliams314:


server/src/main/java/com/objectcomputing/checkins/services/guild/member/GuildMemberRepository.java
```diff
@@ -22,6 +25,13 @@
 
     Optional<GuildMember> findByGuildidAndMemberid(UUID guildMemberid, UUID memberId);
```
bhatiam12, line 26: is this needed or query can be used for this as well ?

server/src/main/java/com/objectcomputing/checkins/services/guild/member/GuildMemberController.java
```diff
@@ -26,11 +26,15 @@
 @Secured(SecurityRule.IS_AUTHENTICATED)
 @Produces(MediaType.APPLICATION_JSON)
 @Tag(name = "guild-member")
+@Singleton
```
meyerinj, line 29: Anything annotated with @Controller also can be injected as a Singleton, so this line is unnecessary.

server/src/main/java/com/objectcomputing/checkins/services/guild/member/GuildMemberRepository.java
meyerinj: I think this serves a different role than the search and should not be rolled up into it, but many of the others on this class safely can, I think.

server/src/test/java/com/objectcomputing/checkins/services/guild/member/GuildServicesImplTest.java
```diff
@@ -15,7 +15,10 @@
 import org.mockito.Mockito;
```
meyerinj, line 15: Not a product of this story specifically, but is there any particular reason the class name doesn't match the file name?

server/src/test/java/com/objectcomputing/checkins/services/guild/member/GuildServicesImplTest.java
```diff
@@ -313,104 +316,80 @@ void testFindByFieldsNullParams() {
                 new GuildMember(UUID.randomUUID(), UUID.randomUUID(), false)
         );
 
-        when(guildMemberRepository.findAll()).thenReturn(guildMemberSet);
+        when(guildMemberRepository.search(null, null, null)).thenReturn(guildMemberSet);
 
         assertEquals(guildMemberSet, services.findByFields(null, null, null));
 
-        verify(guildMemberRepository, times(1)).findAll();
+        verify(guildMemberRepository, times(1)).search(null, null, null);
```
meyerinj, line 323: Because we aren't building the query based on which methods get called these verifies aren't useful anymore. We know the bottom three never get called because they don't exist in the module we're testing and we know the first DOES get called because if it doesn't the assert would fail.

meyerinj: Probably applicable to all the references to `verify` in the other test for `findByFields`.

meyerinj (changes requested):

server/src/test/java/com/objectcomputing/checkins/services/guild/member/GuildServicesImplTest.java
hwilliams314: not sure. fixed though

server/src/main/java/com/objectcomputing/checkins/services/guild/member/GuildMemberController.java
hwilliams314: so, if I am understanding this correctly, @Controller sort of implies (or maybe encompasses) @Singleton and thus makes it unnecessary?

meyerinj: Correct. If you look at the actual class `io.micronaut.http.annotation.Controller` you'll see the line `@DefaultScope(Singleton.class)` which explicitly gives this annotation the scope of a Singleton.

bhatiam12 (approved):

server/src/main/java/com/objectcomputing/checkins/services/guild/member/GuildMemberController.java
hwilliams314: done

hwilliams314: done

meyerinj (approved):

PR #310, opened 2020-09-18
Feature 270/stop blocking

meyerinj:
Implementation of rxJava that will prevent blocking on I/O operations.

Submitting as draft PR as this contains a great deal of unnecessary logging and commenting intended to demonstrate what is happening behind the scenes. I will be returning to this later to remove them.

To explain what is happening this is an example log from running the findActionItems operation:

```
09:24:15.336 [nioEventLoopGroup-1-3] INFO  c.o.c.s.a.ActionItemController - Entering controller on main event loop
09:24:15.417 [pool-1-thread-2] DEBUG io.micronaut.data.query - Executing Query: SELECT * FROM action_items item WHERE (? IS NULL OR item.checkinid = ?) AND (? IS NULL OR item.createdbyid = ?)
09:24:15.417 [pool-1-thread-2] TRACE io.micronaut.data.query - Binding parameter at position 1 to value null
09:24:15.418 [pool-1-thread-2] TRACE io.micronaut.data.query - Binding parameter at position 2 to value null
09:24:15.418 [pool-1-thread-2] TRACE io.micronaut.data.query - Binding parameter at position 3 to value null
09:24:15.418 [pool-1-thread-2] TRACE io.micronaut.data.query - Binding parameter at position 4 to value null
09:24:15.435 [nioEventLoopGroup-1-4] INFO  c.o.c.s.a.ActionItemController - Mapping on main event loop
```

We see that we entered the action item controller on the nioEventLoopGroup, the main event loop, on pool 1, thread 3.
As operation is turned over to the I/O task executor we see that the thread reads `pool-1-thread-2`. Although the naming is unclear (and will be changed to be more descriptive in Micronaut 2.1) this is the I/O task executor's 1st pool and 2nd thread. The previous thread has been released and is no longer blocked by what could be a costly I/O operation.

In the last line we see that The I/O operation has ended and operation has returned to the main event loop, which assigned a different thread to this resumed operation than it was using before.


mkimberlin:
I asked James to take a look at this one and he sent me the following:

```
I don't understand this comment https://github.com/oci-labs/check-ins/pull/310/files#diff-8154f07ebb856d55e5a9db120182e39cR209
I assume actionItemServices.save(actionItem); is blocking
In general I think this is a valid approach
Without doing performance testing, I don't know if it would be better to context switch
or just leave the rest of the logic on the io pool
if I had to guess, I would think leaving it on the io pool would be faster
One thing missing is the configuration of the IO pool
The default configuration is an unbounded pool, so under heavy load that can create many threads. If the IO pool is mostly used for database access, that would likely be a waste because the number of connections to the DB is limited
```

meyerinj:
> I asked James to take a look at this one and he sent me the following:

I asked James for guidance on this before I started and his guidance was that context switching was generally more costly than it was worth so I went into this with the assumption that putting the entire service call onto the I/O threads. This had the benefit of allowing the services and repos to be completely unchanged with the non-blocking logic isolated entirely in the Controller. One of the big delays in getting this done originally was that I tried to split it and the result was confusing to read (you can see it in the first commit on this branch) and was almost impossible to get to work on the bulk load operation.

Reactive is a weak spot of mine even after I read some documentation on it during my OPES days so if there's a better implementation I'm very interested. 

I have pushed a change that (if I understood the documentation correctly) has a limited I/O thread pool.

meyerinj:
@mkimberlin I have made stories to actually implement this and they reference this PR to look to it as an example. Should I just close the PR? Will those other stories keep their reference to this?

PR #311, opened 2020-09-18
fixed failed checkin post

jahanner:


bhatiam12 (approved):

meyerinj:
I don't think I'm quite grocking this change. The bug it references was to change a hard coded URL, but this seems to be blocking the creation of a plugin for cases where a member does not have a PDL.

bhatiam12:
> I don't think I'm quite grocking this change. The bug it references was to change a hard coded URL, but this seems to be blocking the creation of a plugin for cases where a member does not have a PDL.

The original bug was fixed and we found an additional bug where it throws 404 on page reload because a member doesn't have a pdl id assigned and it tries to create a checkin. This check  makes sure that if pdlid exists then do it.

So we just reused the same branch for original bug to fix this.

mkimberlin (approved):

PR #326, opened 2020-09-21
refactored checkin history

jahanner:


web-ui/src/components/checkin/CheckinHistory.jsx
```diff
@@ -1,120 +1,120 @@
...
-      if (pdlId) {
-        let res = await getMember(pdlId);
-        let pdlProfile =
-          res.payload.data && !res.error ? res.payload.data : undefined;
-        setPDL(pdlProfile ? pdlProfile.name : "");
-      }
+  useEffect(() => {
+    if (checkins.length) {
+      const { pathname } = document.location;
+      const [, , checkinid] = pathname.split("/");
```
mkimberlin, line 26: This seems error prone...What happens if someone deploys this under a different root path? I would use react-router to get this value: https://reactrouter.com/web/example/url-params

web-ui/src/components/checkin/CheckinHistory.jsx
```diff
@@ -1,120 +1,120 @@
...
-        <div className="info">
-          <p>{name}</p>
-          <p>{role}</p>
-          <p>PDL: {pdl}</p>
-          <p>Company Email: {workEmail}</p>
+      {getCheckinDate() && (
+        <div className="date-picker">
+          <ArrowBackIcon
+            className={leftArrowClass}
+            onClick={previousCheckin}
```
mkimberlin, line 99: My gut on these would be to simply make them links to the appropriate `/checkins/{id}` path rather than using onClick handlers for them...it reduces the number of possible paths for data loading if you are only worried about the current state originating from the URL. That's just my $0.02.

web-ui/src/components/notes/__snapshots__/Note.test.js.snap
```diff
@@ -27,10 +27,9 @@ exports[`renders correctly 1`] = `
     <div
       className="container"
     >
-      <textarea
-        onChange={[Function]}
+      <div
+        className="is-loading"
```
mkimberlin, line 31: This looks out of date, no?

web-ui/src/components/snackbar/SnackBarWithContext.jsx
```diff
@@ -0,0 +1,28 @@
...
+  return (
+    <SnackBar
+      handleClose={closeToast}
+      open={toast !== "" && severity !== ""}
+      toast={toast}
+    />
+  );
+};
+
+export default SnackBarWithContext;
```
mkimberlin, line 28: Any reason not to combine this file and the one above? I can't see a situation where you would use the inner `SnackBar` by itself, right?

web-ui/src/index.js
```diff
@@ -5,12 +5,7 @@ import App from "./App";
 import * as serviceWorker from "./serviceWorker";
 import "font-awesome/css/font-awesome.min.css";
 
-ReactDOM.render(
-  <React.StrictMode>
-    <App />
-  </React.StrictMode>,
-  document.getElementById("root")
-);
+ReactDOM.render(<App />, document.getElementById("root"));
```
mkimberlin, line 8: Curious as to what lead to this change....

jahanner: I remember having a error specifically related to this so I removed it, but I can't seem to replicate it anymore so I just added it back in.

jahanner: Found it, here is the error if left in StrictMode
index.js:1 Warning: findDOMNode is deprecated in StrictMode. findDOMNode was passed an instance of Transition which is inside StrictMode.

mkimberlin: do you know what is utilizing that api?

mkimberlin (approved):

bhatiam12 (approved):

PR #329, opened 2020-09-22
Feature 248/validations for checkin note

bhatiam12:


server/src/main/java/com/objectcomputing/checkins/services/checkin_notes/CheckinNoteServicesImpl.java
```diff
@@ -73,12 +113,33 @@ public CheckinNote update(CheckinNote checkinNote) {
...
+            } else {
+                validate(!isAdmin,"User is unauthorized to do this operation");
+            }
         return checkinNote;
     }
+
+    private void validate(@NotNull boolean isError, @NotNull String message, Object... args) {
+        if(isError) {
+            throw new CheckinNotesBadArgException(String.format(message, args));
+        }
```
hwilliams314, line 143: It appears that we have a validate method in each of these validation stories. Is this something that we can pull out and just use everywhere?

bhatiam12: each validation method is throwing a different exception so I didn't pull it out.

server/src/main/java/com/objectcomputing/checkins/services/checkin_notes/CheckinNoteServicesImpl.java
```diff
@@ -43,25 +66,42 @@ public CheckinNote save(CheckinNote checkinNote) {
 
     @Override
     public CheckinNote read(@NotNull UUID id) {
-        return checkinNoteRepository.findById(id).orElse(null);
+        String workEmail = securityService!=null ? securityService.getAuthentication().get().getAttributes().get("email").toString() : null;
+        MemberProfile currentUser = workEmail!=null? currentUserServices.findOrSaveUser(null, workEmail) : null;
+        Boolean isAdmin = securityService!=null ? securityService.hasRole(RoleType.Constants.ADMIN_ROLE) : false;
+        CheckinNote checkInNoteResult =  checkinNoteRepository.findById(id).orElse(null);
+        validate(checkInNoteResult == null, "Invalid checkin note id %s", id);
+        if(!isAdmin) {
```
meyerinj, line 74: This can be accomplished with the @Secured annotation as is done on the create and update operations

server/src/test/java/com/objectcomputing/checkins/services/checkin_notes/CheckinNoteControllerTest.java
```diff
@@ -13,15 +13,16 @@
  import io.micronaut.http.HttpStatus;		
  import io.micronaut.http.client.HttpClient;		
  import io.micronaut.http.client.annotation.Client;
```
meyerinj, line 15: There's something wonky going on with the whitespace in this file.

server/src/test/java/com/objectcomputing/checkins/services/checkin_notes/CheckinNoteControllerTest.java
```diff
@@ -95,7 +120,7 @@ void testCreateNullCheckinNote(){
       }		
 
       @Test		
-     void testCreateACheckInNoteForExistingCheckInId() {		
+     void testCreateACheckInNoteForNonExistingCheckInId() {
          MemberProfile memberProfile = createADefaultMemberProfile();		
          MemberProfile memberProfileForPDL = createADefaultMemberProfileForPdl(memberProfile);
```
meyerinj, line 126: Some cleanup can be done here. Unused variable.

meyerinj (changes requested):

server/src/main/java/com/objectcomputing/checkins/services/checkin_notes/CheckinNoteServicesImpl.java
```diff
@@ -1,39 +1,62 @@
...
+        this.securityService = securityService;
+        this.currentUserServices = currentUserServices;
+    }
+
 
     @Override
     public CheckinNote save(CheckinNote checkinNote) {
         CheckinNote checkinNoteRet = null;
+        String workEmail = securityService!=null ? securityService.getAuthentication().get().getAttributes().get("email").toString() : null;
+        MemberProfile currentUser = workEmail!=null? currentUserServices.findOrSaveUser(null, workEmail) : null;
```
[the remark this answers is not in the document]
bhatiam12: I thought that was intentional to create a user and not use findByWorkEmail. That what I understood when I was working with @Bagurp on validations. @Bagurp ?

server/src/test/java/com/objectcomputing/checkins/services/checkin_notes/CheckinNoteControllerTest.java
bhatiam12: there is, switched from vs code to IntelliJ so it messed it up.

server/src/main/java/com/objectcomputing/checkins/services/checkin_notes/CheckinNoteServicesImpl.java
bhatiam12: I didn't do that because on the controller I used IS_AUTHENTICATED annotation and then on create/update I did use admin and pdl because only they can create a note. But on /read end point even a member can read a checkin note and that's why this validation is there to check if its an admin then don't verify anything else and let them read it or else do other validations to make sure only pdl of the subject or the subject can read the notes.

server/src/main/java/com/objectcomputing/checkins/services/checkin_notes/CheckinNoteServicesImpl.java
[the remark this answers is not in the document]
Bagurp: That's probably right @meyerinj. I can't think of a scenario where we'd need to create a new user at this point. @bhatiam12, I remember creating a getCurrentUser() in CurrentUserServiceImpl (but removed it for some reason).

What are your thoughts on injecting @SecurityService only on currentUserServiceImpl? i.e, getCurrentUser() will get email address from security service, validate user with memberProfileRepo.findByWorkEmail, getRoles and add all data to CurrentUserDTO and returns it. This way, we've to inject security service in only one service.

I feel like I've been down this road and chose to inject it in every service.

server/src/main/java/com/objectcomputing/checkins/services/checkin_notes/CheckinNoteServicesImpl.java
meyerinj: Oh, I see. I read the code too quickly and missed what was going on.

I'm going to say in that case that this should throw a different exception that will be caught by the controller and return a 401, rather than 400.

server/src/main/java/com/objectcomputing/checkins/services/checkin_notes/CheckinNoteServicesImpl.java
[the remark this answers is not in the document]
meyerinj: I think it's worth us all having a sit down at some point to talk how best to use the SecurityService.

I'm leaning towards thinking we should only use it on the Controller and it will parse out the relevant information and either reject the user there or pass relevant information into the service. I feel like the nature of the server and how authentication is done should be abstracted away from the services. That said I'm not exactly prepared to commit that to an RFC.

server/src/main/java/com/objectcomputing/checkins/services/checkin_notes/CheckinNoteServicesImpl.java
meyerinj: I was reading about this this morning and it turns out I was wrong. This should return a 403, not a 401.
https://tools.ietf.org/html/rfc7231#section-6.5.3

meyerinj (approved):

mkimberlin (approved):

Bagurp (approved):

PR #346, opened 2020-09-25
changed notes to use query

bhatiam12:


meyerinj:
Not really related to this story but I'm just now noticing that checkin_notes only has end-to-end tests. Are we collectively cool with that?

meyerinj (approved):

bhatiam12:
> Not really related to this story but I'm just now noticing that checkin_notes only has end-to-end tests. Are we collectively cool with that?

I think so. If you look at checkins, that also have end to end tests and I believe we had this discussion when we converted to use end to end tests as every scenario is covered by these tests so didn't see a need to write tests for service layer separately.

Bagurp (approved):

PR #361, opened 2020-09-28
example of jooq operational for action items

meyerinj:
#275 

Jooq relies on code generation so the first thing you'll need to do will be to run build (or simply jooqGenerate) in order for you IDE to understand what's going on. I'm not anticipating this being the way we go so I didn't bother refactoring the tests.

Pros:
-Moving to generated pojos should help to sharpen the currently too-blurry divide between entity and DTO (many controllers are returning entities currently)
-The search methods become a bit clearer to read
- Allows for stronger separation of concerns between service and repository
- When you do your first query on startup the console prints a cute "JOOQ" ascii art

Cons:
- Significant development time investment
- Subjective opinion: most queries are less readable than the jdbc repositories
- Code that could be trusted would now require its own tests
- Duplication of datasource info in application yml and build.gradle
- Reliance on generated code can lead to confusion on pulling down changes

My recommendation would be to stick with the jdbc repositories.

### DIFF
--- a/server/build.gradle
+++ b/server/build.gradle
@@ -12,6 +12,7 @@ plugins {
     id "application"
     id "idea"
     id "groovy"
+    id 'nu.studer.jooq' version '5.1.1'
 }
 
 targetCompatibility = "11"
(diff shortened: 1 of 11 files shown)


meyerinj:
Adding further info as requested by @mkimberlin 

Further work was done on this branch: https://github.com/oci-labs/check-ins/tree/feature-342/jooq_for_team

We found an issue with the jOOQ autogenerated classes not having any apparent way to recognize entity relationships. Although the relationships could be forced by the jOOQ queries themselves representing them would require either not using or extending the autogenerated classes. Team decided this made too much effort to justify transitioning away from Micronaut JDBC.

Concerns were also raised about how jOOQ's autogenerated classes would work in CI. This would require CI to run flyway against some container and generate classes from the result. What level of effort this would involve is unknown.

There was general agreement that transition to jOOQ may be necessary eventually, so these two branches are being kept for posterity, but at this time there is no indication that the effort of such a transition is justified.

Issue #209, opened 2020-08-18
BE > Alter document upload service to handle associations with check-in (M)


mkimberlin:
The `/upload` service needs to be altered to handle document associations with a check-in. It might be worth renaming to something like `/file` as well.

Definition of Done:
- [ ] an endpoint is created for `find by` (check-in id) and/or `find all`
  - [ ] find all should only be allowed by an ADMIN
  - [ ] find by id should only be allowed by:
    - [ ] ADMIN
    - [ ] the current PDL for the user whose check-in it is
    - [ ] PDL for that check-in 
- [ ] an endpoint is created for retrieving the document information (by id)
- [ ] an endpoint is created for retrieving the document contents (maybe `/upload/{id}/download` 😂)
- [ ] an endpoint is created for `create`
  - [ ] endpoint creates a folder within the configured root location named `lastname, firstname - YYYY.MM.DD` if it does not already exist
    - [ ] if it already exists, it that folder is used.
  - [ ] endpoint uploads the file to the created or found named directory
  - [ ] endpoint creates an association to the uploaded document id via the `checkin_document` table
- [ ] an endpoint is created for `delete`
  - [ ] the delete endpoint removes the uploaded document id entry for the document
  - [ ] the delete endpoint removes the document from Google Drive
- [ ] unit tests are created for each endpoint
- [ ] end-to-end test container tests are created for each endpoint
- [ ] code is submitted as a pull request
- [ ] swagger configuration
  - [ ] add schema definition to the entity
- [ ] controller logic and service logic are split into different layers
- [ ] define appropriate DTOs
- [ ] implement appropriate error handlers
- [ ] make use of the headers for location on create method

Bagurp:
- [ ] update GoogleAuthenticator and remove deprecated functions
- [ ] set up property 'google credentials' to be passed in as env variable
- [ ] update github project workflow to add the property 'google credentials'

PR #379, opened 2020-10-06
Issue #209 - rewrite document upload service to handle associations with check-in

Bagurp:
* As part of this story, I also rewrote GoogleAuthenticator.java (since the method it was using was depricated)
* Create a helper method in CheckinDocument service to deleteByUploadDocId.
* Deleted uploadController, tests.
* Setup configuration to get service-account-credentials via env variables.
* Update development build to include env vars of dev service account
* Github Issue @ https://github.com/oci-labs/check-ins/issues/209

end-to-end tests will fail for the PR since GitHub secrets do not have service-account-credentials config.


server/src/main/java/com/objectcomputing/checkins/services/file/FileServices.java
```diff
@@ -0,0 +1,14 @@
+package com.objectcomputing.checkins.services.file;
+
+import io.micronaut.http.HttpResponse;
+import io.micronaut.http.multipart.CompletedFileUpload;
+
+import java.io.File;
```
Maroju100, line 6: unused import

Maroju100 (approved):

server/src/main/java/com/objectcomputing/checkins/services/file/FileServicesImpl.java
```diff
@@ -0,0 +1,226 @@
...
+        this.checkInServices = checkInServices;
+        this.checkinDocumentServices = checkinDocumentServices;
+        this.memberProfileServices = memberProfileServices;
+    }
+
+    @Override
+    public HttpResponse<?> findFiles(@Nullable UUID checkInID) {
+
+        Boolean isAdmin = securityService!=null ? securityService.hasRole(RoleType.Constants.ADMIN_ROLE) : false;
+        validate(checkInID == null && !isAdmin, "You are not authorized to perform this operation");
```
mkimberlin, line 62: From the DoD:

- [ ] find by id should only be allowed by:
  - [ ] ADMIN
  - [ ] the current PDL for the user whose check-in it is
  - [ ]  PDL for that check-in

I think you are missing those last two cases.

server/src/main/java/com/objectcomputing/checkins/services/file/FileServicesImpl.java
```diff
@@ -0,0 +1,226 @@
...
+                    .status(HttpStatus.valueOf(e.getStatusCode()))
+                    .body(e.getContent());
+        } catch (IOException e) {
+            LOG.error("Error occurred while retrieving files from Google Drive.", e);
+            return HttpResponse.status(HttpStatus.INTERNAL_SERVER_ERROR);
+        }
+    }
+
+    @Override
+    public HttpResponse<?> downloadFiles(@NotNull String uploadDocId) {
```
mkimberlin, line 100: It's missing from the DoD, but this endpoint should follow the same validation as findFiles.

server/src/main/java/com/objectcomputing/checkins/services/file/FileServicesImpl.java
```diff
@@ -0,0 +1,226 @@
...
+                    .status(HttpStatus.valueOf(e.getStatusCode()))
+                    .body(e.getContent());
+        } catch (IOException e) {
+            LOG.error("Error occurred while retrieving files from Google Drive.", e);
+            return HttpResponse.status(HttpStatus.INTERNAL_SERVER_ERROR);
+        }
+    }
+
+    @Override
+    public HttpResponse<FileInfoDTO> uploadFile(@NotNull UUID checkInID, @NotNull CompletedFileUpload file) {
```
mkimberlin, line 128: Also missing from the DoD...upload should only be possible by the same folk as above (current PDL, checkin PDL, or ADMIN). sorry. :-( Those missing items can be made into a new issue if you'd prefer. Just please do so before merging/closing this one.

Bagurp: Not a problem, it shouldn't be a big change.

server/src/main/java/com/objectcomputing/checkins/services/file/FileServicesImpl.java
```diff
@@ -0,0 +1,226 @@
...
+
+        CheckIn checkIn = checkInServices.read(checkInID);
+        validate(checkIn == null, "Unable to find checkin record with id %s", checkInID);
+        validate((file.getFilename() == null || file.getFilename().equals("")), "Please select a file before uploading.");
+
+        // create folder name in the format name-date
+        String subjectName = memberProfileServices.getById(checkIn.getTeamMemberId()).getName();
+        String sb = subjectName.concat(LocalDate.now().toString());
+        sb = sb.replaceAll("\\s", "");
+        final String directoryName = sb;
```
mkimberlin, line 138: Can we make this name match the format in the DoD? it's ok to include time too...but it should at least start: `lastname, firstname`. I know it's kind of a pain.

server/src/main/java/com/objectcomputing/checkins/services/file/FileServicesImpl.java
```diff
@@ -0,0 +1,226 @@
...
+                    .body(setFileInfo(uploadedFile, cd));
+        } catch (IOException e) {
+            LOG.error("Unexpected error processing file upload.", e);
+            return HttpResponse.status(HttpStatus.INTERNAL_SERVER_ERROR);
+        }
+    }
+
+    @Override
+    public HttpResponse<?> deleteFile(@NotNull String uploadDocId) {
+
```
mkimberlin, line 183: Should also only be able to be done by the folk above...again...can be part of a new story.

server/src/main/resources/application.yml
```diff
@@ -71,3 +71,16 @@ flyway:
...
+  type: ${ TYPE }
+  project_id: ${ PROJECT_ID }
+  private_key_id: ${ PRIVATE_KEY_ID }
+  private_key: ${ PRIVATE_KEY }
+  client_email: ${ CLIENT_EMAIL }
+  client_id: ${ CLIENT_ID }
+  auth_uri: ${ AUTH_URI }
+  token_uri: ${ TOKEN_URI }
+  auth_provider_x509_cert_url: ${ AUTH_PROVIDER_X509_CERT_URL }
+  client_x509_cert_url: ${ CLIENT_X509_CERT_URL }
```
mkimberlin, line 86: I wonder if we shouldn't have just automagically mapped these from env variables like `SERVICE_ACCOUNT_DIRECTORY_ID`...those can be automagically brought in as params out of the environment without the need to add them to the yaml. Just a thought, we can discuss.

mkimberlin (approved):
None of my comments are blockers, so long as we address them either in either this PR or with the creation of new issues.

server/src/main/java/com/objectcomputing/checkins/services/file/FileServicesImpl.java
Bagurp: We don't have an option to fetch last and first names from memberprofile. When creating a new user, we are getting 'name' attribute from auth object and saving it in memberprofile. This attribute includes first and last names of the user.

In the above code, directoryName will be something like - PramukhBagur2020-10-08
If a user has a middle name / middle initial, it will be something like - PramukhDayanandaBagur2020-10-08

We can skip time since only 1 directory will be created for the user on a day. DOD -
- [ ] endpoint creates a folder within the configured root location named lastname, firstname - YYYY.MM.DD if it does not already exist

When the user tries to upload second document on the same day, new directory won't be created.

Could you suggest a workaround for doing `lastname, firstname`?

server/src/main/java/com/objectcomputing/checkins/services/file/FileServicesImpl.java
Bagurp: @mkimberlin - Should we block user from uploading a new document if checkin is completed?

server/src/main/java/com/objectcomputing/checkins/services/file/FileServices.java
Bagurp: removed it :)

server/src/main/java/com/objectcomputing/checkins/services/file/FileServicesImpl.java
Bagurp: updated in the latest commit.

server/src/main/java/com/objectcomputing/checkins/services/file/FileServicesImpl.java
Bagurp: updated in the latest commit.

server/src/main/java/com/objectcomputing/checkins/services/file/FileServicesImpl.java
Bagurp: updated in the latest commit.

server/src/main/java/com/objectcomputing/checkins/services/file/FileServicesImpl.java
Bagurp: updated in the latest commit.

server/src/main/resources/application.yml
Bagurp: I will try to refactor this as part of the bugfix issue.

PR #380, opened 2020-10-06
add priority to agenda item

bhatiam12:


server/src/main/java/com/objectcomputing/checkins/services/agenda_item/AgendaItemRepository.java
```diff
@@ -27,5 +27,8 @@
     <S extends AgendaItem> List<S> saveAll(@Valid @NotNull Iterable<S> entities);
 
     @Override
-    <S extends AgendaItem> S save(@Valid @NotNull @Nonnull S entity);
-}
\ No newline at end of file
+    <S extends AgendaItem> S save(@Valid @NotNull @NonNull S entity);
+
+    Optional<Double> findMaxPriorityByCheckinid(UUID checkinid);
```
Maroju100, line 32: do we need a unit test for this?

Maroju100 (approved):

Bagurp (approved):

server/src/main/java/com/objectcomputing/checkins/services/agenda_item/AgendaItemRepository.java
meyerinj: This raises a not unreasonable question  about all of our repositories. When we make a method such as `findById` or `findMaxPriorityByCheckinId` or using explicit queries we are writing code and it is currently only getting tested when doing end-to-end tests.

Does Micronaut have a best practice on this?

meyerinj (approved):

server/src/main/java/com/objectcomputing/checkins/services/agenda_item/AgendaItemRepository.java
bhatiam12: I will add service level test for this in another PR as another PR #391 is depending on this one.

PR #397, opened 2020-10-13
Feature 247/validation action item closes #247

### LINKED ISSUE #247: BE> Add validation to the update to action items (M)
Definition of Done
- [x] Create, update or delete is only allowed if the associated check in is not completed unless made by admin
- [x] Limit creation to 
     - [x] subject of the check in
     - [x] the current PDL of the subject of the check in
     - [x] the admin
- [x] Limit read 
     - [x] subject of the check in 
     - [x] the current PDL of the subject of the check in
     - [x] the admin
- [x] Limit update 
     - [x] subject of the check in if created by 
     - [x] the current PDL of the subject of the check in
     - [x] the admin
- [x] Limit delete
     - [x] subject of the check in if created by 
     - [x] the current PDL of the subject of the check in
     - [x] the admin
- [x] add these to end-to-end tests

hwilliams314:


server/src/main/java/com/objectcomputing/checkins/services/action_item/ActionItemServicesImpl.java
```diff
@@ -82,14 +124,70 @@ public ActionItem update(ActionItem actionItem) {
...
+
+        validate(actionItem == null, "invalid action item %s", id);
+
+        final UUID checkinId = actionItem.getCheckinid();
+        final UUID createById = actionItem.getCreatedbyid();
+
+        CheckIn checkinRecord = checkinRepo.findById(checkinId).orElse(null);
+        Boolean isCompleted = checkinRecord != null ? checkinRecord.isCompleted() : null;
+        final UUID pdlId = checkinRecord != null ? checkinRecord.getPdlId() : null;
+        final UUID teamMemberId = checkinRecord != null ? checkinRecord.getTeamMemberId() : null;
```
Maroju100, line 168: variable teamMemberId could be deleted it is not used

server/src/main/java/com/objectcomputing/checkins/services/action_item/ActionItemServices.java
```diff
@@ -9,7 +9,7 @@
 
     ActionItem read(UUID id);
 
-    Set<ActionItem> readAll();
+//    Set<ActionItem> readAll();
```
Maroju100, line 12: could be deleted

Maroju100 (approved):

server/src/main/java/com/objectcomputing/checkins/services/action_item/ActionItemServicesImpl.java
hwilliams314: good catch - updated

server/src/main/java/com/objectcomputing/checkins/services/action_item/ActionItemServices.java
hwilliams314: done, thanks!

jahanner (commented):
You'll want to get rid of that package-lock.json file before you merge

hwilliams314:
> You'll want to get rid of that package-lock.json file before you merge

done, thanks!

server/src/main/java/com/objectcomputing/checkins/services/action_item/ActionItemServicesImpl.java
```diff
@@ -1,79 +1,121 @@
...
+            CheckIn checkinRecord = checkinRepo.findById(checkinId).orElse(null);
+            Boolean isCompleted = checkinRecord != null ? checkinRecord.isCompleted() : null;
+            final UUID pdlId = checkinRecord != null ? checkinRecord.getPdlId() : null;
+            final UUID teamMemberId = checkinRecord != null ? checkinRecord.getTeamMemberId() : null;
+            validate(checkinId == null || createdById == null, "Invalid action item %s", actionItem);
+            validate(actionItem.getId() != null, "Found unexpected id %s for action item", actionItem.getId());
+            validate(checkinRepo.findById(checkinId).isEmpty(), "CheckIn %s doesn't exist", checkinId);
+            validate(memberRepo.findById(createdById).isEmpty(), "Member %s doesn't exist", createdById);
+            if (!isAdmin && isCompleted) {
+                validate(!currentUser.getId().equals(pdlId) || !currentUser.getId().equals(teamMemberId), "User is unauthorized to do this operation");
```
Bagurp, line 57: if the user is not admin and checkin is completed, shouldn't we throw ActionItemBadArgException directly?

if the user is not admin and checkin is not completed, then we should validate if user belongs to the associated checkin.

server/src/main/java/com/objectcomputing/checkins/services/action_item/ActionItemServicesImpl.java
```diff
@@ -1,79 +1,121 @@
...
-            if (guildId == null || createById == null) {
-                throw new ActionItemBadArgException(String.format("Invalid actionItem %s", actionItem));
-            } else if (id == null || !actionItemRepo.findById(id).isPresent()) {
-                throw new ActionItemBadArgException(String.format("Unable to locate actionItem to update with id %s", id));
-            } else if (!checkinRepo.findById(guildId).isPresent()) {
-                throw new ActionItemBadArgException(String.format("CheckIn %s doesn't exist", guildId));
-            } else if (!memberRepo.findById(createById).isPresent()) {
-                throw new ActionItemBadArgException(String.format("Member %s doesn't exist", createById));
+
+            CheckIn checkinRecord = checkinRepo.findById(checkinId).orElse(null);
```
Bagurp, line 108: * Asking for a friend

Not to throw Holly under the bus, but what's the best practice when we call one service from another? Is is okay to call database layer of another service directly? or is it good practice to call service (interface)? or it doesn't matter?

@meyerinj @mkimberlin @bhatiam12

Maroju100 (approved):

server/src/main/java/com/objectcomputing/checkins/services/action_item/ActionItemServicesImpl.java
meyerinj: This should not be throwing that exception at all. If we are kicking the user out for bad permissions it should be returning a 403. That exception will resolve as a 400.

server/src/main/java/com/objectcomputing/checkins/services/action_item/ActionItemServicesImpl.java
meyerinj: The purpose of the service layer is to abstract some back end process (in this case the repositories) from some calling process. There's no universal answer on whether separate services should interact with separate repositories, but the structure of check-ins being as highly delineated as it is I would generally lean toward going through the services rather than directly to the repositories. This will serve to prevent duplication of input validation and force exposure of any issues of permissions being changed in one area but not another.

meyerinj (changes requested):
In addition to the notes about the response types I recall a while back we were talking about implementing a permissions service. What ever happened with that?

server/src/main/java/com/objectcomputing/checkins/services/action_item/ActionItemServicesImpl.java
```diff
@@ -1,79 +1,131 @@
...
+            Boolean isCompleted = checkinRecord != null ? checkinRecord.isCompleted() : null;
+            final UUID pdlId = checkinRecord != null ? checkinRecord.getPdlId() : null;
+            final UUID teamMemberId = checkinRecord != null ? checkinRecord.getTeamMemberId() : null;
+
+            validation.validateArguments(checkinId == null || createdById == null, "Invalid action item %s", actionItem);
+            validation.validateArguments(actionItem.getId() != null, "Found unexpected id %s for action item", actionItem.getId());
+            validation.validateArguments(checkInServices.read(checkinId) == null, "CheckIn %s doesn't exist", checkinId);
+            validation.validateArguments(memberServices.getById(createdById) == null, "Member %s doesn't exist", createdById);
+
+            if (!isAdmin && isCompleted) {
```
meyerinj, line 61: Let's use `validatePermissions` to do this validation, rather than doing the validation here and just having that method throw an exception. For example a `validateActionItemsPermissions` method that does all this.

meyerinj (changes requested):

server/src/main/java/com/objectcomputing/checkins/services/action_item/ActionItemController.java
```diff
@@ -25,20 +28,41 @@
 @Secured(SecurityRule.IS_AUTHENTICATED)
 @Produces(MediaType.APPLICATION_JSON)
 @Tag(name = "action-item")
+@Singleton
```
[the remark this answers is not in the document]
hwilliams314: that's weird - I have the @Controller in my local

server/src/main/java/com/objectcomputing/checkins/services/action_item/ActionItemServicesImpl.java
```diff
@@ -1,79 +1,131 @@
...
-            final UUID createById = actionItem.getCreatedbyid();
+            final UUID checkinId = actionItem.getCheckinid();
+            final UUID createdById = actionItem.getCreatedbyid();
+
+            CheckIn checkinRecord = checkInServices.read(checkinId);
+            Boolean isCompleted = checkinRecord != null ? checkinRecord.isCompleted() : null;
+            final UUID pdlId = checkinRecord != null ? checkinRecord.getPdlId() : null;
+            final UUID teamMemberId = checkinRecord != null ? checkinRecord.getTeamMemberId() : null;
+
+            validation.validateArguments(checkinId == null || createdById == null, "Invalid action item %s", actionItem);
```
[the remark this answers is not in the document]
hwilliams314: done

meyerinj (approved):

meyerinj (approved):

Bagurp (approved):

mkimberlin (approved):
It's definitely an improvement. I'm a little iffy about keeping the business logic validations for everything in one big class. I feel pretty certain that will eventually proliferate back out to live with the services themselves. I like the reduction of boilerplate exceptions.  We'll ride with this and see at what point we need to consider how we house the validations.

Issue #389, opened 2020-10-08
BE> Create delete endpoint for team service (M)


sharmag04:
Definition of Done

- [x] end point exists for delete team by id
- [x] swagger
- [x] unit tests
- [x] end-to-end tests
- [x] only an admin/ one of the team leaders can delete a team

PR #402, opened 2020-10-15
Feature 389/create delete for team #389

Maroju100:
-delete end point for team and controller tests
-V32 alter sql script for "on delete cascade" to delete team member references when a team is deleted.


server/src/main/java/com/objectcomputing/checkins/services/team/TeamServicesImpl.java
```diff
@@ -63,4 +82,23 @@ public Team update(Team team) {
         }
         return teams;
     }
+
+    // public void delete(@NotNull UUID id) {
+    //     teamsRepo.deleteById(id);
+    // }
+
```
hwilliams314, line 89: Might want to remove old, commented out method =)

Maroju100: updated, thanks =)

hwilliams314 (approved):

meyerinj (approved):

PR #403, opened 2020-10-15
Bugfix 384/file service unit testing

Bagurp:
* Adds util methods to CheckinDocumentService which is used in FileServiceImpl.deleteFile
* Convert end-to-end tests for file service into unit tests
* Deleted secrets under /resources since we are injecting secrets as env variables
* PR for original service - https://github.com/oci-labs/check-ins/pull/379


server/src/main/java/com/objectcomputing/checkins/services/file/FileServicesImpl.java
```diff
@@ -197,11 +203,16 @@ public FileServicesImpl(GoogleDriveAccessor googleDriveAccessor, EmailSender ema
             CheckinDocument cd = new CheckinDocument(checkInID, uploadedFile.getId());
             checkinDocumentServices.save(cd);
 
-            emailSender.sendEmail("New Check-in Notes", "New check-in notes have been uploaded. Please check the Google Drive folder.");
+//            emailSender.sendEmail("New Check-in Notes", "New check-in notes have been uploaded. Please check the Google Drive folder.");
```
hwilliams314, line 206: can this commented out code be removed?

Bagurp: This was part of the old uploadController. It was sending email to hr@objectcomputing.com from kimberlinm@objectcomputing.com after every successful file upload. Since the new file service did not have an AC for sending email after file upload, I've commented it out. (I am assuming that we will add this back at some point)

The logic for this module is available at java/com/objectcomputing/checkins/notifications/email

hwilliams314 (approved):

server/src/test/java/com/objectcomputing/checkins/services/file/FileControllerTest.java
```diff
@@ -55,237 +60,117 @@ void deleteTestFile() {
...
-        assertEquals(resultFromPost.getFileId(), result.getFileId());
-        assertEquals(resultFromPost.getCheckInId(), result.getCheckInId());
-        assertEquals(resultFromPost.getName(), result.getName());
+        assertNotNull(response);
+        verify(fileServices, times(1)).uploadFile(any(UUID.class), any(CompletedFileUpload.class));
     }
 
     @Test
-    public void testFindByCheckInIdFailsForUnrelatedUser() {
+    public void testUploadEndpointFailsForInvalidFile() {
```
meyerinj, line 119: This test failed for me.  `(No such file or directory)` vs `(The system cannot find the path specified)`

server/src/test/java/com/objectcomputing/checkins/services/file/FileServicesImplTest.java
```diff
@@ -0,0 +1,1019 @@
...
+        Mockito.reset(checkInServices);
+        Mockito.reset(checkinDocumentServices);
+        Mockito.reset(googleDriveAccessor);
+        Mockito.reset(securityService);
+        Mockito.reset(currentUserServices);
+        Mockito.reset(memberProfileServices);
+        Mockito.reset(completedFileUpload);
+    }
+
+    private void setupMocksForAuth() {
```
meyerinj, line 151: This is in every test. Would it make sense to put it in the BeforeEach?

meyerinj (changes requested):

server/src/test/java/com/objectcomputing/checkins/services/file/FileServicesImplTest.java
Bagurp: Good catch, not sure how I missed it. I've moved it to beforeEach now.

server/src/main/java/com/objectcomputing/checkins/services/file/FileServices.java
```diff
@@ -8,6 +8,6 @@
 public interface FileServices {
     HttpResponse<?> findFiles(UUID checkInId);
```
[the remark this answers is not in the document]
Bagurp: Thanks for bringing this up. I've updated methods to reflect this.

server/src/test/java/com/objectcomputing/checkins/services/file/FileControllerTest.java
Bagurp: Not sure why you are seeing this. Could you try once after the latest commit? I am not seeing this on my local and looks like tests pass on Github actions as well.

meyerinj (approved):
Approving, but with the caveat that this should either have the non-blocking processes added to the controllers before merge or a story needs to be written up to get that done at a later date. There are probably no services in check-ins that would more benefit from it.

Bagurp:
I have added story to convert fileServiceController to non-blocking - 
https://github.com/oci-labs/check-ins/issues/439

PR #435, opened 2020-11-02
Feature #419 - Allow name to be null on MemberProfile

Bagurp:


Maroju100 (approved):

meyerinj (approved):
Since the alter statement is on its way

bhatiam12 (approved):

Bagurp:
Merging this since Alter script is not required. In postgresql, columns are nullable by default. To make a column not null, we should add NOT NULL constraint. At the moment, `name` does not have NOT NULL constraint.

PR #443, opened 2020-11-06
Feature 409/alter team member skill

warnerj-oci:
Add fields to member skill for skill level and last used date.


meyerinj (approved):

server/src/main/java/com/objectcomputing/checkins/services/member_skill/MemberSkillCreateDTO.java
```diff
@@ -17,6 +18,12 @@
     @Schema(description = "the id of the skill", required = true)
     private UUID skillid;
 
+    @Schema(description = "the level of the skill")
```
hwilliams314, line 21: For the description here could we say something like "member's expertise level for this skill"?

warnerj-oci:
Certainly we could. I was keeping the description similar to the
descriptions for the other fields. I will change it.

server/src/main/java/com/objectcomputing/checkins/services/member_skill/MemberSkillCreateDTO.java
warnerj-oci: This is done.

hwilliams314 (approved):

PR #446, opened 2020-11-10
Feature 273/csrf token

bhatiam12:


server/src/main/java/com/objectcomputing/checkins/security/OAuthCsrfFilter.java
```diff
@@ -0,0 +1,30 @@
...
+
+    @Override
+    protected Publisher<MutableHttpResponse<?>> doFilterOnce(HttpRequest<?> request, ServerFilterChain chain) {
+        HttpHeaders headers = request.getHeaders();
+
+        String headerValue = headers.get("X-CSRF-Header", String.class).orElse("");
+        String cookieValue = request.getCookies().findCookie("_csrf").map(Cookie::getValue).orElse(null);
+
+
+        if (!cookieValue.equals(headerValue)) {
```
meyerinj, line 24: This seems to have broken all the unit tests.

meyerinj (changes requested):

server/src/main/java/com/objectcomputing/checkins/security/OAuthCsrfFilter.java
bhatiam12: yaa, now I need to find a way to ignore this filter on tests otherwise I will need to modify tests. I think I will try adding @env=GCP and see if that take cares of it.

meyerinj: If it's not intended to run locally then that would be a good fix. If it is you have the option of injecting this filter (or a mock of it) in the TestContainerSuite.

bhatiam12: I was able to add environment as GCP because it is not needed locally.

server/src/main/java/com/objectcomputing/checkins/security/CsrfModelProcessor.java
```diff
@@ -0,0 +1,34 @@
...
+    @Get("/cookie")
+    public HttpResponse <?> getCsrfToken()  {
+        SecureRandom random = new SecureRandom();
+        byte[] randomBytes = new byte[24];
+        random.nextBytes(randomBytes);
+        String cookieValue = base64Encoder.encodeToString(randomBytes);
+
+        return HttpResponse.ok()
+        // set cookie
+                .cookie(new NettyCookie("_csrf", cookieValue).path("/")).body(cookieValue);
```
mkimberlin, line 30: I think you also need to set samesite = strict with this cookie. not sure offhand how to do that with this API, so you might need to dig a bit.

server/src/main/java/com/objectcomputing/checkins/security/OAuthCsrfFilter.java
```diff
@@ -0,0 +1,33 @@
...
+import io.micronaut.http.*;
+import io.micronaut.http.annotation.Filter;
+import io.micronaut.http.cookie.Cookie;
+import io.micronaut.http.filter.OncePerRequestHttpServerFilter;
+import io.micronaut.http.filter.ServerFilterChain;
+import org.reactivestreams.Publisher;
+
+
+
+@Filter(value = {"/services/*", "/services/*"})
```
mkimberlin, line 15: Am I missing something here? aren't these two the same?

server/src/main/java/com/objectcomputing/checkins/security/CsrfModelProcessor.java
bhatiam12: done

server/src/main/java/com/objectcomputing/checkins/security/OAuthCsrfFilter.java
bhatiam12: fixed it.

jahanner (approved):

mkimberlin (approved):

PR #450, opened 2020-11-10
Feature 408/alter skill entity

warnerj-oci:
Add fields for description and to mark skill as extraneous in skill entity

### DIFF
--- a/server/src/main/java/com/objectcomputing/checkins/services/skills/Skill.java
+++ b/server/src/main/java/com/objectcomputing/checkins/services/skills/Skill.java
@@ -111,19 +111,23 @@ public class Skill {
         Skill skill = (Skill) o;
         return pending == skill.pending &&
                 Objects.equals(id, skill.id) &&
-                Objects.equals(name, skill.name);
+                Objects.equals(name, skill.name) &&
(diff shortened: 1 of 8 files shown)


meyerinj (approved):

server/src/main/java/com/objectcomputing/checkins/services/skills/Skill.java
```diff
@@ -74,6 +88,22 @@ public void setPending(boolean pending) {
...
+    public boolean isExtraneous() {
+        return extraneous;
+    }
+
+    public void setExtraneous(boolean extraneous) {
+        this.extraneous = extraneous;
+    }
+
     @Override
     public boolean equals(Object o) {
```
Bagurp, line 108: Should we add `description` and `extraneous` fields to `equals`, `hashcode` and `toString` methods?

Bagurp (approved):

server/src/main/java/com/objectcomputing/checkins/services/skills/Skill.java
warnerj-oci: Of course, good catch!  I will make the changes.

PR #460, opened 2020-11-17
guild not blocking

Maroju100:


warnerj-oci (approved):

server/src/main/java/com/objectcomputing/checkins/services/guild/GuildController.java
```diff
@@ -41,59 +56,95 @@ public GuildController(GuildServices guildService) {
...
+         return Single.fromCallable(() -> {
+             List<String> errors = new ArrayList<>();
+             List<Guild> guildCreated = new ArrayList<>();
+             for (GuildCreateDTO guildDTO : guildsList) {
+                 Guild guild = new Guild(guildDTO.getName(),
+                         guildDTO.getDescription());
+                 try {
+                     guildService.save(guild);
+                     guildCreated.add(guild);
+                 } catch (CompositeException e) {
```
bhatiam12, line 115: what's the advantage of throwing composite exception here ? when we call save only GuildBadArg exception is thrown in service class. I am new to RxJava so trying to understand why it is done this way ? @meyerinj 
@Maroju100 also there are no test cases to test this exception ?  how are we going to test this ? because right now it just throws BadArg exception from service layer as per the test cases we have.

meyerinj: There are scenarios where a reactivex process, because it can encompass multiple threads, may need to throw multiple exceptions. This gets bundled up in a composite exception. This try/catch occurs within a single reactivex process, so it shouldn't ever have the opportunity catch a composite exception. Rather, it's something that could come up if the entire `Single.fromCallable` was wrapped in a try/catch. This is why many of the other controllers have an error handler for composite exception

server/src/main/java/com/objectcomputing/checkins/services/guild/GuildController.java
```diff
@@ -41,59 +56,95 @@ public GuildController(GuildServices guildService) {
...
+                     errors.add(String.format("Guild %s was not added because: %s", guild.getName(),
+                             guild.getDescription(), e.getMessage()));
+                 }
+             }
+             if (errors.isEmpty()) {
+                 return guildCreated;
+             }
+             throw new GuildBulkLoadException(errors);
+         }).map(guildCreated -> HttpResponse.created(guildCreated)
```
meyerinj, line 124: The single needs to observe on the main event loop and then the map needs to observe on the io loop. Otherwise we aren't necessarily using the right threads for this process.

meyerinj (changes requested):

server/src/main/java/com/objectcomputing/checkins/services/guild/GuildController.java
Maroju100: Based on chats with John and Michael we have decided to remove the bulk load end points and the associated tests, created an issue for it.

server/src/main/java/com/objectcomputing/checkins/services/guild/GuildController.java
Maroju100: Based on chats with John and Michael we have decided to remove the bulk load end points and the associated tests, created an issue for it.

bhatiam12: Good this answer during sprint planning, will be removing this on team.

server/src/main/java/com/objectcomputing/checkins/services/guild/GuildController.java
```diff
@@ -41,47 +52,34 @@ public GuildController(GuildServices guildService) {
...
+     * @param guild, {@link GuildCreateDTO}
+     * @return {@link HttpResponse<Guild>}
      */
+    @Post("/")
+    public Single<HttpResponse<Guild>> createAGuild(@Body @Valid GuildCreateDTO guild,
+                                                             HttpRequest<GuildCreateDTO> request) {
+        return Single.fromCallable(() -> guildService.save(new Guild(guild.getName(), guild.getDescription())))
+                .observeOn(Schedulers.from(eventLoopGroup))
+                .map(createdGuild -> {
+                    //Using code block rather than lambda so we can log what thread we're in
```
meyerinj, line 76: This comment can be removed and this block can be replaced with a lambda.

meyerinj (approved):

Issue #407, opened 2020-10-21
BE> Create service for tagging entities (M)


sharmag04:
- [x] Add a UUID for the tag itself
- [x] Add a name for the tag
- [x] Add a type for the tag
- [x] Update tests as needed

meyerinj:
Are there any acceptance criteria for this? From the name I'm gathering we want there to be tags which can have a many-to-many relationship with skills? Is that right?

mkimberlin:
Correct.

PR #466, opened 2020-11-19
Feature 407/skill tags

meyerinj:
#407


server/src/main/java/com/objectcomputing/checkins/services/skills/tags/SkillTagController.java
```diff
@@ -0,0 +1,88 @@
...
+                        .ok(skillTagResponseDTOs))
+                .subscribeOn(Schedulers.from(ioExecutorService));
+    }
+
+    @Post("/")
+    public Single<HttpResponse<SkillTagResponseDTO>> save(@Valid @NotNull SkillTagCreateDTO dto) {
+        return Single.fromCallable(() -> skillTagService.save(dto))
+                .observeOn(Schedulers.from(eventLoopGroup))
+                .map(skillTagResponseDTO -> (HttpResponse<SkillTagResponseDTO>)HttpResponse
+                        .ok(skillTagResponseDTO))
```
Bagurp, line 76: Should we return 201 instead of 200?

Bagurp (approved):

server/src/main/java/com/objectcomputing/checkins/services/skills/tags/SkillTagController.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    @Error(exception = SkillTagNotFoundException.class)
+    public HttpResponse<?> handleNotFound(HttpRequest<?> request, SkillTagNotFoundException stnfe) {
+        JsonError error = new JsonError(stnfe.getMessage())
+                .link(Link.SELF, Link.of(request.getUri()));
+
+        return HttpResponse.<JsonError>status(HttpStatus.NOT_FOUND).body(error);
+    }
+
```
Maroju100, line 52: We can add comments over each service end point as we do for other controllers.

server/src/main/java/com/objectcomputing/checkins/services/skills/SkillRepository.java
```diff
@@ -17,4 +20,11 @@
 
     List<Skill> findByPending(boolean pending);
 
+    @Override
+    @Join(value = "tags", type = Join.Type.LEFT_FETCH)
```
mkimberlin, line 24: Does this have an effect with JDBC repos?

server/src/main/java/com/objectcomputing/checkins/services/skills/Skill.java
```diff
@@ -43,6 +42,11 @@
     @Schema(description = "the skill is extraneous (or not)", required = true)
     private boolean extraneous = false;
 
+    @ManyToMany(mappedBy = "skills",
+            fetch = FetchType.EAGER,
+            cascade = CascadeType.ALL)
+    private List<SkillTag> tags;
```
mkimberlin, line 48: Should tags be exclusive to skills? Perhaps they should be able to tag other things? Also does the `@ManyToMany` have an effect?

server/src/main/java/com/objectcomputing/checkins/services/skills/tags/SkillTag.java
```diff
@@ -0,0 +1,107 @@
...
+
+    @Column(name = "name")
+    @TypeDef(type = DataType.STRING)
+    @Schema(description = "the name of the skill tag", required = true)
+    private String name;
+
+    @Column(name = "description")
+    @TypeDef(type = DataType.STRING)
+    @Schema(description = "the description of the skill tag", required = false)
+    private String description;
```
mkimberlin, line 36: Let's remove description...

server/src/main/java/com/objectcomputing/checkins/services/skills/SkillRepository.java
meyerinj: Yes. If this isn't here JDBC won't do a fetch on the joined records and the skill will be returned without its tags.

server/src/main/java/com/objectcomputing/checkins/services/skills/Skill.java
meyerinj: Rereading the story it says "create service for tagging skills" and I took that to mean the tags were for skills but that's definitely not what it says. This would be a very do-able refactor but it would be more involved than I'd want to do since I'm not on this project anymore. My vote is we should leave it as skill tags for now and make a story to refactor it to be more generalized.

As for the many-to-many JDBC can use it to auto-populate based on the join using the @join annotation in the repository and the matching annotations on the SkillTag entity.

PR #482, opened 2020-11-25
remove unnecessary bulkload endpoints

Maroju100:
bulk load end points and tests are removed as they were not needed.


Maroju100:
Closing the pull request since it was created with branch name BE and not going through the checks. Will recreate a new branch and do a pull request.

PR #498, opened 2020-12-08
Shuffled some dependencies and made some adjustments to tests

mkimberlin:


jahanner (approved):
I assume all those yarn changes are to fix the test issues caused by babel?

mkimberlin:
> I assume all those yarn changes are to fix the test issues caused by babel?

Are you talking about the changes in yarn.lock? If so, those are a result of my dependency changes and upgrades that occurred when I regenerated that file.

PR #536, opened 2020-12-28
Feature 489/crud validator action item #closes 489

hwilliams314:


server/src/main/java/com/objectcomputing/checkins/services/validate/crud/CRUDValidator.java
```diff
@@ -0,0 +1,19 @@
...
+
+   public void validateArgumentsCreate(T entity);
+   public void validateArgumentsRead(T entity);
+   public void validateArgumentsUpdate(T entity);
+   public void validateArgumentsDelete(T entity);
+
+   public void validatePermissionsCreate(T entity);
+   public void validatePermissionsRead(T entity);
+   public void validatePermissionsUpdate(T entity);
+   public void validatePermissionsFindByFields(UUID entity, UUID secondEntity);
```
hwilliams314, line 15: This(validatePermissionsFindByFields) could be an issue with later services as most of the services use different entities and number of entities to do this search. A few even call this method by a different name (FindByValues).

warnerj-oci (approved):

Bagurp (approved):

PR #577, opened 2021-01-06
Feature 462/add supervisorid to member directory

warnerj-oci:
Add supervisor name to member directory card and add member


web-ui/src/components/member-directory/MemberModal.jsx
```diff
@@ -105,47 +112,49 @@ const MemberModal = ({ member = {}, open, onSave, onClose }) => {
...
-          <Button
-            onClick={async () => {
-              onSave(editedMember);
-              await updateMember(editedMember);
-            }}
-            color="primary"
-          >
-            Save Member
-          </Button>
+         <Autocomplete
```
meyerinj, line 115: Tabbing appears to be a bit off here.

web-ui/src/components/member-directory/MemberSummaryCard.jsx
```diff
@@ -12,17 +14,33 @@ import SplitButton from "../split-button/SplitButton";
...
-  const [open, setOpen] = useState(false);
+  const [supervisorName, setSupervisorName] = useState();
 
+  const [open, setOpen] = useState(false);
   const handleOpen = () => setOpen(true);
 
   const handleClose = () => setOpen(false);
 
+  // Get Supervisor's name
+  useEffect(() => {
```
meyerinj, line 30: Are we worried about doing an extra request for each record slowing down page load? This being the member directory page we already have every member loaded into client memory or the supervisor name can be made a field on the response DTO.

meyerinj (approved):

web-ui/src/components/member-directory/MemberModal.jsx
warnerj-oci: oops! Thanks.

web-ui/src/components/member-directory/MemberSummaryCard.jsx
warnerj-oci: That's a good point. I was having trouble accessing the member profile on that page, the call was a fallback.  Let me take another look at that.

zsbrown97 (approved):

PR #584, opened 2021-01-07
Feature 493/add delete team member closes #493

### LINKED ISSUE #493: Add delete endpoint for teamMembers (S)
Definition of Done - 
 - [x] write the end point for deleting a team member
 - [x] write tests for delete

hwilliams314:


server/src/test/java/com/objectcomputing/checkins/services/team/member/TeamMemberControllerTest.java
```diff
@@ -453,8 +454,61 @@ void testUpdateTeamMemberThrowExceptionWithInvalidId() {
...
+        final HttpRequest<Object> request = HttpRequest.
+                DELETE(String.format("/%s", teamMember.getId())).basicAuth(ADMIN_ROLE, ADMIN_ROLE);
+
+        HttpClientResponseException responseException = assertThrows(HttpClientResponseException.class,
+                () -> client.toBlocking().exchange(request, Map.class));
+
+        assertNotNull(responseException.getResponse());
+        assertEquals(HttpStatus.NOT_FOUND, responseException.getStatus());
+    }
+
```
Bagurp, line 513: Can we add a test case to make sure that team lead can remove a member?

server/src/main/java/com/objectcomputing/checkins/services/team/member/TeamMemberServicesImpl.java
```diff
@@ -108,4 +110,26 @@ public TeamMember update(@NotNull @Valid TeamMember teamMember) {
...
+    public void delete(@NotNull UUID id) {
+        MemberProfile currentUser = currentUserServices.getCurrentUser();
+        boolean isAdmin = currentUserServices.isAdmin();
+
+        TeamMember teamMember = teamMemberRepo.findById(id).orElse(null);
+        if(teamMember != null) {
+            Set<TeamMember> teamLeads = this.findByFields(teamMember.getTeamid(), null, true);
+
+            if (teamMemberRepo.findById(id).isEmpty()) {
+                throw new NotFoundException(String.format("Unable to locate teamMember to delete with id %s", id));
```
Bagurp, line 123: Looks like this statement unreachable. Since `teamMemberRepo.findById(id).isEmpty()` is always `false`

server/src/main/java/com/objectcomputing/checkins/services/team/member/TeamMemberServicesImpl.java
```diff
@@ -108,4 +110,26 @@ public TeamMember update(@NotNull @Valid TeamMember teamMember) {
...
+            } else if (!isAdmin && teamLeads.stream().noneMatch(o -> o.getMemberid().equals(currentUser.getId()))) {
```
meyerinj, line 124: Since checking if the authenticated user is a lead for a given user  comes up a lot this is something worth adding to the currentUserServices.

meyerinj (changes requested):

server/src/main/java/com/objectcomputing/checkins/services/team/member/TeamMemberServicesImpl.java
mkimberlin: I don't think this requested change is in the scope of this story, given that this logic is spread throughout the app currently and is fairly straight forward.  Further, I think I would be more prone to make it a method in the team or team member service, rather than current user and make it something that you could ask about any user id/team id combination...but that's certainly up for discussion/debate.

Regardless, I don't think we should hold this PR for that discussion and task. Thoughts, @meyerinj ?

hwilliams314: Probably, but let's do it in another story =)

server/src/main/java/com/objectcomputing/checkins/services/team/member/TeamMemberServicesImpl.java
mkimberlin: Agreed... @hwilliams314 could you remove the check above?

server/src/test/java/com/objectcomputing/checkins/services/team/member/TeamMemberControllerTest.java
hwilliams314: done

server/src/main/java/com/objectcomputing/checkins/services/team/member/TeamMemberServicesImpl.java
hwilliams314: done

meyerinj (approved):

Bagurp (approved):

PR #602, opened 2021-01-14
Feature 557/delete guild closes #557

### LINKED ISSUE #557: BE> Add delete functionality to Guild (M)
Definition of Done

 - [x] create delete end point
 - [x] create delete method in services impl 
 - [x] validation - delete by admin only
 - [x] add on delete cascade to db
 - [x] add tests

hwilliams314:


server/src/main/java/com/objectcomputing/checkins/services/guild/GuildRepository.java
```diff
@@ -23,13 +23,16 @@
             "FROM guilds guild " +
             "WHERE (:name IS NULL OR guild.name LIKE :name) " +
             "AND (:memberid IS NULL OR guild.id IN " +
-                "(SELECT guildid FROM guildmembers gm " +
```
meyerinj, line 26: I would argue the tabbing here is useful as it shows that the subquery being at the SQL version of a lower scope than the rest of the query.

meyerinj (approved):

server/src/main/java/com/objectcomputing/checkins/services/guild/GuildRepository.java
hwilliams314: Intellij must have changed that

zsbrown97 (approved):

mkimberlin (approved):

PR #627, opened 2021-01-20
Resolve bug 527 which caused a 403 on editing a user

meyerinj:
#527

Replacing the old (and misspelled) privileges exception was an early attempted fix and since we don't need it anyway I left it deleted even though we don't actually need that change for this bugfix.


hwilliams314 (approved):

Maroju100 (approved):

Bagurp (approved):

server/src/main/java/com/objectcomputing/checkins/services/memberprofile/MemberProfileServicesImpl.java
```diff
@@ -1,7 +1,7 @@
 package com.objectcomputing.checkins.services.memberprofile;
 
 
-import com.objectcomputing.checkins.security.InsufficientPrivelegesException;
+import com.objectcomputing.checkins.services.exceptions.PermissionException;
```
Bagurp, line 4: This has been moved to com.objectcomputing.checkins.exceptions.PermissionException. Try pulling the latest.

PR #628, opened 2021-01-21
Honor the configured upload directory…

mkimberlin:
Properly support shared drives
Adjust the list all files capability
Enable downloads in the UI

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -1,5 +1,13 @@
 version: '3'
 services:
+  pgadmin:
+    image: dpage/pgadmin4
+    environment:
+      PGADMIN_DEFAULT_EMAIL: checkins@objectcomputing.com
+      PGADMIN_DEFAULT_PASSWORD: admin
+    ports:
+      - "80:80"
+    container_name: pgadmin
   postgresql:
     image: postgres:11.6
     environment:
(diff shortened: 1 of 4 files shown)


docker-compose.yaml
```diff
@@ -1,5 +1,13 @@
 version: '3'
 services:
+  pgadmin:
+    image: dpage/pgadmin4
+    environment:
+      PGADMIN_DEFAULT_EMAIL: checkins@objectcomputing.com
+      PGADMIN_DEFAULT_PASSWORD: admin
+    ports:
+      - "80:80"
+    container_name: pgadmin
```
mkimberlin, line 10: In the future, it might be worth splitting sonarqube and pgadmin into a separate docker-compose file. Maybe a `docker-compose-tools.yaml` or some such...

warnerj-oci (approved):

bhatiam12 (approved):

docker-compose.yaml
```diff
@@ -1,5 +1,13 @@
 version: '3'
 services:
+  pgadmin:
```
hwilliams314, line 3: so pgadmin here is not the postgres gui?

mkimberlin: it is the postgres gui

Issue #598, opened 2021-01-13
BUG> Allow PDLs to view checkin documents for their assigned team members (S)


mkimberlin:
Current behavior: 400 is thrown with an unauthorized error message when requested by the team member's current PDL

Expected behavior: The document information is returned when requested by the team member's current PDL

- [ ] Alter validation to allow the current PDL to retrieve documents for their team members
- [ ] Add or alter tests to account for this change

PR #655, opened 2021-02-04
removed unneeded validation as checkins now does it itself.

meyerinj:
#598


Maroju100 (approved):

server/src/main/java/com/objectcomputing/checkins/services/memberprofile/currentuser/CurrentUserServicesImpl.java
```diff
@@ -63,6 +76,35 @@ public MemberProfile getCurrentUser() {
...
+        MemberProfile currentUser = getCurrentUser();
+        MemberProfile checkUser = memberProfileRepo.findById(memberId).orElseThrow(() -> {
+            throw new NotFoundException("No such member exists");
+        });
+
+        if (checkUser.getPdlId().equals(currentUser.getId())) {
+            return true;
+        }
+
+        List<Team> memberTeams = teamRepository.matchMemberAndlead(memberId.toString(), currentUser.getId().toString());
```
Bagurp, line 93: Per my understanding, this piece of code is mentioning that if teamMember and TeamLead (Current user) are part of the same team, and if teamLead is the currentUser then return true (that currentUser is PDL for memberID). Is this correct?

could you elaborate on why this validation is required?

server/src/main/java/com/objectcomputing/checkins/services/file/FileServicesImpl.java
```diff
@@ -93,10 +93,6 @@ public FileServicesImpl(GoogleApiAccess googleApiAccess,
                 CheckIn checkIn = checkInServices.read(checkInID);
                 validate(checkIn == null, String.format("Unable to find checkin record with id %s", checkInID));
 
-                if(!isAdmin) {
-                    validate((!currentUser.getId().equals(checkIn.getTeamMemberId()) && !currentUser.getId().equals(checkIn.getPdlId())), "You are not authorized to perform this operation");
```
Bagurp, line 97: Should we update this to call `access granted` method instead of removing validation? That way, only the users with right permission will be able to access the resource.

mkimberlin: the checkinservice is doing this

mkimberlin (approved):

Issue #675, opened 2021-02-15
BE > Create skills report service (M)


mkimberlin:
# Definition of Done:

- [x] a web service endpoint is created to accept a POST skill report request
- [x] base URL for the service is `/reports/skills`
- [x] if "inclusive" is not included or "false", then any members included in the results must have *all* of the requested skills
- [x] if a set of member IDs is provided, then the results should be limited to only those members
- [x] if a skillLevel is specified, then the member skill must be of that level or higher in order for the member to be included in the results
- [x] unit tests are created for endpoint
- [x] swagger configuration has been updated
- [x] code is submitted as a pull request

# SkillReportRequestDTO should contain:
- requestedSkills - a list of SkillLevelDTOs that contain the following:
  - skillId - the UUID of the skill to include (not null)
  - skillLevel - the required proficiency for this skill (nullable)
- members - a set of UUIDs for the member profiles of the members to be included (nullable)
- inclusive - Boolean value indicating if team members that only have *some* of the requested skill requirements

# SkillReportResponseDTO should contain:
- teamMembers - a list of TeamMemberSkillDTOs that contain the following:
  - memberId - the UUID of the member profile for the team member
  - name - the name of the team member
  - skills - a list of SkillLevelDTOs (see above) for this member representing the member's skill level in each of the requested skills

PR #708, opened 2021-02-20
feature-675/Add skills report service

sonndinh:


mkimberlin:
Hey, @sonndinh...in order to get your CI build to run, you will have to rename your branch (or make a new one, whatever) with the one of the formats listed in the CONTRIBUTING.md doc. Basically, this one would be: `feature-675/some-description`.

server/src/main/java/com/objectcomputing/checkins/services/member_skill/skillsreport/SkillsReportServicesImpl.java
```diff
@@ -0,0 +1,169 @@
...
+    private final MemberSkillRepository memberSkillRepo;
+    private final MemberProfileRepository memberProfileRepo;
+
+    public SkillsReportServicesImpl(MemberSkillRepository memberSkillRepo,
+                                    MemberProfileRepository memberProfileRepo) {
+        this.memberSkillRepo = memberSkillRepo;
+        this.memberProfileRepo = memberProfileRepo;
+    }
+
+    public SkillsReportResponseDTO report(SkillsReportRequestDTO request) {
```
Bagurp, line 26: we could use a @notnull here instead of manual check on line 28

server/src/main/java/com/objectcomputing/checkins/services/member_skill/skillsreport/SkillLevelDTO.java
```diff
@@ -0,0 +1,35 @@
...
+
+@Introspected
+public class SkillLevelDTO {
+    @NotNull
+    @Schema(required = true, description = "UUID of the skill")
+    private UUID id;
+
+    @Nullable
+    @Schema(description = "Level of the skill")
+    private String level;
```
Bagurp, line 17: should we make this an `enum` instead of `string`?

Bagurp:
Just confirming - We do not need a `GET` method for `SkillsReportController` since POST method creates a report and responds with SkillReportResponseDTO. Is this assumption correct?

Do we need validations for access control? (Not sure if these part of the story)

@sonndinh @mkimberlin

server/src/main/java/com/objectcomputing/checkins/services/member_skill/skillsreport/SkillsReportServicesImpl.java
sonndinh: Does `@NotNull` enforce that the parameter cannot be null? Or is it only for documentation purposes?

server/src/main/java/com/objectcomputing/checkins/services/member_skill/skillsreport/SkillLevelDTO.java
sonndinh: I agree that `enum` should be used. This is just so that it's compatible with the existing definition of skill level in member_skill service, which uses `string`. If we decide to change to `string`, I can do that in this PR.

sonndinh:
> Just confirming - We do not need a `GET` method for `SkillsReportController` since POST method creates a report and responds with SkillReportResponseDTO. Is this assumption correct?
> 
Yes, this is what I understand. Michael can confirm it.

> Do we need validations for access control? (Not sure if these part of the story)
> 
> @sonndinh @mkimberlin
Do you mean controlling who can get a skills report?

server/src/main/java/com/objectcomputing/checkins/services/member_skill/skillsreport/SkillLevelDTO.java
mkimberlin: I would like to change the database and the related entities/dtos to use an Integer instead of a string.

mkimberlin:
> > Just confirming - We do not need a `GET` method for `SkillsReportController` since POST method creates a report and responds with SkillReportResponseDTO. Is this assumption correct?
> 
> Yes, this is what I understand. Michael can confirm it.

Consider it confirmed. :) This service implements a command pattern, where a command entity is posted...no need to do a GET in this situation unless it were a long lived request that we wanted to be able to offload the loading of the results until later. This shouldn't need that.

> > Do we need validations for access control? (Not sure if these part of the story)
>
> Do you mean controlling who can get a skills report?

Member skills are available to anyone in the system...no need for access control beyond ensuring they are authenticated.

server/src/main/java/com/objectcomputing/checkins/services/member_skill/skillsreport/SkillsReportServicesImpl.java
Bagurp: Yes, it enforces that parameter should not be null.

https://docs.oracle.com/javaee/7/api/javax/validation/constraints/NotNull.html

example in repo - https://github.com/oci-labs/check-ins/blob/develop/server/src/main/java/com/objectcomputing/checkins/services/checkin_notes/CheckinNoteServicesImpl.java

sonndinh: Okay, thanks.

sonndinh: I added `@NotNull` and removed the manual check, but one of the tests shows that the method is still called when the input parameter is `null`. So I'm moving back to use manual check until we figure that out.

server/src/main/java/com/objectcomputing/checkins/services/member_skill/skillsreport/SkillLevelDTO.java
```diff
@@ -0,0 +1,36 @@
...
+
```
mkimberlin, line 19: Did you opt not to do this under this story? That's fine...just confirming.

mkimberlin: (by this I mean the resolved comment above)

server/src/main/java/com/objectcomputing/checkins/services/member_skill/skillsreport/SkillsReportController.java
```diff
@@ -0,0 +1,55 @@
...
+import javax.validation.Valid;
+import javax.validation.constraints.NotNull;
+import java.net.URI;
+import java.util.concurrent.ExecutorService;
+
+@Controller("/reports/skills")
+@Secured(SecurityRule.IS_AUTHENTICATED)
+@Produces(MediaType.APPLICATION_JSON)
+@Consumes(MediaType.APPLICATION_JSON)
+@Tag(name = "skills-report")
```
mkimberlin, line 25: Should be `/reports/skills`, per issue #675

server/src/main/java/com/objectcomputing/checkins/services/member_skill/skillsreport/SkillsReportServices.java
```diff
@@ -0,0 +1,5 @@
+package com.objectcomputing.checkins.services.member_skill.skillsreport;
+
+public interface SkillsReportServices {
+    SkillsReportResponseDTO report(SkillsReportRequestDTO request);
```
mkimberlin, line 4: I don't think I like using the DTO directly here. DTO is really a transport thing and this is the service API. I think I would pass in the individual parameters needed by the method. If we are doing this elsewhere in the app, we should likely create stories to clean that up.

mkimberlin: I could be convinced otherwise on this...it's a grey area. Especially given that I would be inclined to share the response object in this case. It can always be refactored. 🤷

server/src/main/java/com/objectcomputing/checkins/services/member_skill/skillsreport/SkillsReportServicesImpl.java
```diff
@@ -0,0 +1,188 @@
...
+        this.memberSkillRepo = memberSkillRepo;
+        this.memberProfileRepo = memberProfileRepo;
+        this.skillRepo = skillRepo;
+
+        skillLevels = new HashMap<>();
+        skillLevels.put("interested", 0);
+        skillLevels.put("novice", 1);
+        skillLevels.put("intermediate", 2);
+        skillLevels.put("advanced", 3);
+        skillLevels.put("expert", 4);
```
mkimberlin, line 37: You've already got this bit...why not go ahead and go full enum now?

sonndinh:
Replaced by PR #725. Thanks @mkimberlin for moving the comments over to the new PR.

PR #709, opened 2021-02-22
Deleted agenda item not found exception

moshirim:


mkimberlin (approved):
Don't worry about the sonar issue. We have dropped below testing thresholds somehow. I'll take a look and maybe we'll get a couple of stories added.

hwilliams314 (approved):

PR #710, opened 2021-02-22
Refactored member skills to use global exception

moshirim:


server/src/main/java/com/objectcomputing/checkins/services/member_skill/MemberSkillController.java
```diff
@@ -56,25 +58,16 @@ public MemberSkillController(MemberSkillServices memberSkillServices,
...
-                .link(Link.SELF, Link.of(request.getUri()));
-
-        return HttpResponse.<JsonError>badRequest()
-                .body(error);
-    }
-
-    @Error(exception = MemberSkillAlreadyExistsException.class)
-    public HttpResponse<?> handleAlreadyExists(HttpRequest<?> request, MemberSkillAlreadyExistsException e) {
+    @Error(exception = AlreadyExistsException.class)
+    public HttpResponse<?> handleAlreadyExists(HttpRequest<?> request, AlreadyExistsException e) {
```
hwilliams314, line 62: Don't think we need this. Should be handled by the global handler AlreadyExistsHandler in the exceptions/handlers folder

server/src/main/java/com/objectcomputing/checkins/services/member_skill/MemberSkillController.java
```diff
@@ -56,25 +58,16 @@ public MemberSkillController(MemberSkillServices memberSkillServices,
...
         JsonError error = new JsonError(e.getMessage())
                 .link(Link.SELF, Link.of(request.getUri()));
 
         return HttpResponse.<JsonError>status(HttpStatus.CONFLICT).body(error);
     }
 
-    @Error(exception = MemberSkillNotFoundException.class)
-    public HttpResponse<?> handleNotFound(HttpRequest<?> request, MemberSkillNotFoundException e) {
+    @Error(exception = NotFoundException.class)
+    public HttpResponse<?> handleNotFound(HttpRequest<?> request, NotFoundException e) {
```
hwilliams314, line 70: Don't think we need this. Should be handled by the global handler NotFoundHandler in the exceptions/handlers folder

mkimberlin: @moshirim Holly is correct. This handler can be removed.

mkimberlin: The same is true for the AlreadyExistsException handler, as Holly mentioned.

server/testFile.txt
```diff
@@ -0,0 +1 @@
+This.Is.A.Test.File
```
mkimberlin: Deleted this...it doesn't belong.

hwilliams314 (approved):

Bagurp (approved):

PR #724, opened 2021-02-24
Feature 700/encrypt member profile #closes 700

hwilliams314:


zsbrown97 (approved):

moshirim (approved):

server/src/test/resources/logback-test.xml
```diff
@@ -13,4 +13,5 @@
     </root>
 
     <logger name="org.flywaydb" level="ERROR"/>
+<!--    <logger name="io.micronaut.data" level="DEBUG"/>-->
```
Maroju100, line 16: Commented out line may be removed if not needed.

Maroju100 (approved):

server/src/test/resources/logback-test.xml
hwilliams314: Normally I would agree with you, but Michael likes to leave this one as it is often, but not always, needed and this way it is there when we do need it. =)

mkimberlin (approved):

PR #725, opened 2021-02-25
Feature 675: Add skills report service

sonndinh:
Issue #675


server/src/main/java/com/objectcomputing/checkins/services/member_skill/skillsreport/SkillsReportController.java
```diff
@@ -0,0 +1,55 @@
...
+import javax.validation.Valid;
+import javax.validation.constraints.NotNull;
+import java.net.URI;
+import java.util.concurrent.ExecutorService;
+
+@Controller("/reports/skills")
+@Secured(SecurityRule.IS_AUTHENTICATED)
+@Produces(MediaType.APPLICATION_JSON)
+@Consumes(MediaType.APPLICATION_JSON)
+@Tag(name = "skills-report")
```
mkimberlin, line 25: Should be /reports/skills, per issue #675 (bringing comment over from other PR)

server/src/main/java/com/objectcomputing/checkins/services/member_skill/skillsreport/SkillsReportServicesImpl.java
```diff
@@ -0,0 +1,191 @@
...
+        this.memberSkillRepo = memberSkillRepo;
+        this.memberProfileRepo = memberProfileRepo;
+        this.skillRepo = skillRepo;
+
+        skillLevels = new HashMap<>();
+        skillLevels.put("interested", 0);
+        skillLevels.put("novice", 1);
+        skillLevels.put("intermediate", 2);
+        skillLevels.put("advanced", 3);
+        skillLevels.put("expert", 4);
```
mkimberlin, line 37: You've already got this bit...why not go ahead and go full enum now? (bringing comment over from previous PR)

server/src/main/java/com/objectcomputing/checkins/services/member_skill/skillsreport/SkillsReportServices.java
```diff
@@ -0,0 +1,5 @@
+package com.objectcomputing.checkins.services.member_skill.skillsreport;
+
+public interface SkillsReportServices {
+    SkillsReportResponseDTO report(SkillsReportRequestDTO request);
```
mkimberlin, line 4: I don't think I like using the DTO directly here. DTO is really a transport thing and this is the service API. I think I would pass in the individual parameters needed by the method. If we are doing this elsewhere in the app, we should likely create stories to clean that up.

I could be convinced otherwise on this...it's a grey area. Especially given that I would be inclined to share the response object in this case. It can always be refactored. 🤷 (bringing comment over from previous PR)

server/src/main/java/com/objectcomputing/checkins/services/member_skill/skillsreport/SkillsReportServicesImpl.java
```diff
@@ -0,0 +1,191 @@
...
+                } else {
+                    final List<TeamMemberSkillDTO> qualifiedMembers = getMembersSatisfyingAllSkills(potentialMembers, skills);
+                    response.setTeamMembers(qualifiedMembers);
+                }
+            } else {
+                if (inclusive == null || !inclusive) {
+                    final List<TeamMemberSkillDTO> qualifiedMembers = removeMembersNotRequested(potentialMembers, members);
+                    response.setTeamMembers(qualifiedMembers);
+                } else {
+                    final List<TeamMemberSkillDTO> membersInList = removeMembersNotRequested(potentialMembers, members);
```
mkimberlin, line 77: line 74 and 77 are the same...pull outside of the if/else? for that matter, if you make the check for null or empty members at the top and only removed members in that case. I think you could simplify these nested conditions. Not a big deal, just a thought.

server/src/main/java/com/objectcomputing/checkins/services/member_skill/skillsreport/SkillsReportController.java
mkimberlin: sorry, I was miss reading this...that's what I get for rushing.

mkimberlin (approved):
Couple of minor comments, but nothing that would prevent merging.

mkimberlin:
I just realized you are working off of a fork. For some reason that doesn't trigger our build it looks like. Since you are on the team, you don't have to use a fork and can push your feature branches directly. Though, we should probably try to support forks. :-\ I just didn't realize that was a problem.

server/src/main/java/com/objectcomputing/checkins/services/member_skill/skillsreport/SkillsReportController.java
sonndinh: So is the tag name `skills-report` fine? or should it also be `/reports/skills`?

server/src/main/java/com/objectcomputing/checkins/services/member_skill/skillsreport/SkillsReportServicesImpl.java
sonndinh: I was assuming changing skill level to enum involves also updating `MemberSkill` and the related code since its skill level field is also `String`. So I was inclined to create another PR with that assumption. If converting to enum is only for this service, then I'll update this PR to do that. What do you think?

server/src/main/java/com/objectcomputing/checkins/services/member_skill/skillsreport/SkillsReportServices.java
sonndinh: Yeah, I saw other services not using DTO directly as parameters but wasn't sure what would be a good practice. I'll change the input parameters. Is returning `SkillsReportResponseDTO` ok or should I change its type?

server/src/main/java/com/objectcomputing/checkins/services/member_skill/skillsreport/SkillsReportServicesImpl.java
sonndinh: > line 74 and 77 are the same...pull outside of the if/else?
Good point. Thanks.

server/src/main/java/com/objectcomputing/checkins/services/member_skill/skillsreport/SkillsReportServicesImpl.java
sonndinh: Updated to use enum.

server/src/main/java/com/objectcomputing/checkins/services/member_skill/skillsreport/SkillsReportController.java
mkimberlin: it's fine. we need to reorganize all of them anyway later.

server/src/main/java/com/objectcomputing/checkins/services/member_skill/skillsreport/SkillsReportServices.java
mkimberlin: I think I would make drop the DTO from it's name (the response obj), since it's being returned from the service. if at some point we need to return something organized differently from the controller we will call that thing a DTO

server/src/main/java/com/objectcomputing/checkins/services/member_skill/skillsreport/SkillsReportServicesImpl.java
```diff
@@ -0,0 +1,213 @@
...
+                    return SkillLevel.EXPERT;
+                default:
+                    throw new BadArgException(String.format("Invalid skill level %s", level));
+            }
+        }
+
+        public boolean greaterThanOrEqual(SkillLevel other) {
+            return value >= other.getValue();
+        }
+    }
```
mkimberlin, line 54: I think I would pull this enum out into its own top level class. You can use it in the DTOs as well as in this service layer.

hwilliams314 (approved):

server/src/main/java/com/objectcomputing/checkins/services/member_skill/skillsreport/SkillsReportServicesImpl.java
sonndinh: This is fixed now.

PR #740, opened 2021-03-03
Created service for tagging entities

moshirim:


LICENSE
```diff
@@ -31,16 +31,16 @@
       "License" shall mean the terms and conditions for use, reproduction,
       and distribution as defined by Sections 1 through 9 of this document.
 
-      "Licensor" shall mean the copyright owner or entity authorized by
+      "Licensor" shall mean the copyright owner or Tag authorized by
```
sonndinh, line 34: Looks like "entity"s are replaced by "Tag"s by mistake in this file?

sonndinh (commented):
Overall the code looks good to me; most of my comments are minor. Another structural thing is that there are 2 new services added to the same subdirectory `checkins/services/tags`, but seems to me we should have one subdirectory for each new service, based on how other services are organized.

LICENSE
mkimberlin, line 34: All of the changes in this file need to be reverted. It's the License for the project.

server/src/main/java/com/objectcomputing/checkins/services/tags/EntityTagServicesImpl.java
```diff
@@ -0,0 +1,89 @@
...
+
+    public Set<EntityTag> findByFields(UUID entityId, UUID tagId, EntityType type) {
+
+            return entityTagRepository.search(nullSafeUUIDToString(entityId), nullSafeUUIDToString(tagId), type);
+    }
+
+    public EntityTag update(@NotNull EntityTag entityTag) {
+
+        final boolean isAdmin = currentUserServices.isAdmin();
+        permissionsValidation.validatePermissions(!isAdmin, "User is unauthorized to do this operation");
```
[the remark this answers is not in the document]
mkimberlin: Yes, We do. Please add this @moshirim

server/src/main/java/com/objectcomputing/checkins/services/tags/TagServicesImpl.java
```diff
@@ -0,0 +1,76 @@
...
+
+        public Set<Tag> findByFields(String name) {
+
+            return tagRepository.search(name);
+        }
+
+        public Tag update(@NotNull Tag tag) {
+
+            final boolean isAdmin = currentUserServices.isAdmin();
+            permissionsValidation.validatePermissions(!isAdmin, "User is unauthorized to do this operation");
```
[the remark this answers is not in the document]
mkimberlin: +1 on this as well, @moshirim

server/src/main/resources/db/common/V41__create_tags_table.sql
```diff
@@ -0,0 +1,15 @@
+drop table if exists tags;
```
mkimberlin, line 1: you are missing a drop if exists on `entity_tags`

mkimberlin (changes requested):
See comments. I also unresolved a number of @sonndinh 's comments as they have not been addressed in the code yet and are valid.

moshirim:
All of issues mentioned 

> Overall the code looks good to me; most of my comments are minor. Another structural thing is that there are 2 new services added to the same subdirectory `checkins/services/tags`, but seems to me we should have one subdirectory for each new service, based on how other services are organized.

Thank you for your feedback Son, I took care of all of the things you mentioned. Let me know what you think.

moshirim:
> See comments. I also unresolved a number of @sonndinh 's comments as they have not been addressed in the code yet and are valid.

Should be all good now.

server/src/main/resources/db/common/V41__create_tags_table.sql
mkimberlin: You are still missing a drop if exists on `entity_tags`.

server/src/main/resources/db/common/V41__create_tags_table.sql
```diff
@@ -0,0 +1,16 @@
...
+drop table if exist entity_tags;
```
mkimberlin, line 2: `exists` not `exist`

sonndinh (approved):

PR #755, opened 2021-03-10
Split member name

sonndinh:


server/src/main/java/com/objectcomputing/checkins/services/member_skill/skillsreport/SkillsReportServicesImpl.java
```diff
@@ -108,7 +112,8 @@ public SkillsReportResponseDTO report(@NotNull SkillsReportRequestDTO request) {
                 final TeamMemberSkillDTO dto = new TeamMemberSkillDTO();
                 dto.setId(memberId);
 
-                final String memberName = memberProfileRepo.findNameById(memberId);
+                //final String memberName = memberProfileRepo.findNameById(memberId);
+                final String memberName = memberProfileServices.getName(memberId);
```
Maroju100, line 116: getName is undefined in the code.

sonndinh: Yeah, I'm updating the related classes. This isn't complete yet.

sonndinh:
All unit tests passed and the PR is ready to review. One thing I'm not totally sure about is the changes to the `currentuser` service. Because member names are split, I've added a `firstName` and `lastName` authentication attributes in place of the existing `name` attribute. I'm not sure if this is ok since the attributes are read from Google authentication, as noted in the `CurrentUserController` class.

server/src/main/java/com/objectcomputing/checkins/services/memberprofile/MemberProfileCreateDTO.java
```diff
@@ -13,9 +13,21 @@
...
+
+    @NotBlank
+    @Schema(description = "last name of the employee")
+    private String lastName;
+
     @Nullable
-    @Schema(description = "full name of the employee")
-    private String name;
+    @Schema(description = "suffix of the employee")
+    private String suffix;
```
mkimberlin, line 30: We need to keep "name" in our dtos and set it using your convenience method. Otherwise the frontend has to append the name together everywhere and that will be a pain.

server/src/main/java/com/objectcomputing/checkins/services/memberprofile/MemberProfileResponseDTO.java
```diff
@@ -16,9 +16,21 @@
...
+    @Schema(description = "middle name of the employee")
+    private String middleName;
+
+    @NotBlank
+    @Schema(description = "last name of the employee")
+    private String lastName;
+
     @Nullable
-    @Schema(description = "full name of the employee")
-    private String name;
```
mkimberlin, line 21: Ditto here on keeping name around and populating it with your new method.

server/src/main/java/com/objectcomputing/checkins/services/memberprofile/MemberProfileUpdateDTO.java
```diff
@@ -16,9 +16,21 @@
...
+    @Schema(description = "middle name of the employee")
+    private String middleName;
+
+    @NotBlank
+    @Schema(description = "last name of the employee")
+    private String lastName;
+
     @Nullable
-    @Schema(description = "full name of the employee")
-    private String name;
```
mkimberlin, line 21: keep name around...

server/src/main/java/com/objectcomputing/checkins/services/memberprofile/currentuser/CurrentUserDTO.java
```diff
@@ -5,16 +5,21 @@
...
 import javax.validation.constraints.NotNull;
 import java.util.List;
 import java.util.Objects;
 
 @Introspected
 public class CurrentUserDTO {
 
-    @Nullable
-    @Schema(description = "full name of the user")
-    private String name;
```
mkimberlin, line 17: keep name...

server/src/main/java/com/objectcomputing/checkins/services/team/member/TeamMemberResponseDTO.java
```diff
@@ -12,43 +11,47 @@
     @Schema(description = "id of the entry", required = true)
     private UUID id;
 
-    @Schema(description = "name of the member this entry is associated with", required = true)
-    private String name;
```
mkimberlin, line 16: keep name...

mkimberlin (changes requested):
Need to keep name around to avoid breaking the frontend.

server/src/main/java/com/objectcomputing/checkins/security/LocalLoginController.java
```diff
@@ -67,10 +67,10 @@ public LocalLoginController(Authenticator authenticator,
...
             if (authenticationResponse.isAuthenticated() && authenticationResponse.getUserDetails().isPresent()) {
                 UserDetails userDetails = authenticationResponse.getUserDetails().get();
-                MemberProfile memberProfile = currentUserServices.findOrSaveUser(email, email);
-                String name = memberProfile.getName() != null ? memberProfile.getName() : "";
-                userDetails.setAttributes(Map.of("email", memberProfile.getWorkEmail(), "name", name,
-                        "picture", ""));
+                // The values for attributes email, firstName, lastName match the arguments passed to the call
+                // to method CurrentUserServices.findOrSaveUser in LocalUserPasswordAuthProvider.authenticate
+                userDetails.setAttributes(Map.of("email", email, "firstName", email,
+                        "lastName", email, "picture", ""));
```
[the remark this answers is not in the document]
sonndinh: > So, you are skipping the call to find or save here because it's already happening in the authprovider?
> 
Yes, the call to `findOrSaveUser` here is redundant to me. When user logs in the first time, `findOrSaveUser(email, email)` is called in `LocaUserPasswordAuthProvider.authenticate` which should create a new MemberProfile for this user. The call to `findOrSaveUser` here is used only to get the created member's name, but we already know its name is `email` since `email` was passed as the member name in the first `findOrSaveUser` call. 

> I think I would keep the name logic in place for firstname/lastname and send those over as empty string if they don't exist, rather than sending email. Unless there's something I'm missing here?

As discussed above, `email` is used as the member name in the local login controller. (In the local login web interface, it has 2 fields for email and role. I think those two are used as the username and password in the local environment.) So the name passed to `userDetails.setAttributes()` is always `email` although it isn't obvious in the old code.

Since now we split name to first name/last name but the front-end only sends email and role, I am using the given email as both first name and last name (https://github.com/oci-labs/check-ins/blob/c044a7a0cb216a15b99e64a1d462afe92f999413/server/src/main/java/com/objectcomputing/checkins/security/LocalUserPasswordAuthProvider.java#L39).

server/src/main/java/com/objectcomputing/checkins/security/LocalUserPasswordAuthProvider.java
```diff
@@ -36,7 +36,7 @@ public LocalUserPasswordAuthProvider(CurrentUserServices currentUserServices,
     @Override
     public Publisher<AuthenticationResponse> authenticate(@Nullable HttpRequest<?> httpRequest, AuthenticationRequest<?, ?> authReq) {
         String email = authReq.getIdentity().toString();
-        MemberProfile memberProfile = currentUserServices.findOrSaveUser(email, email);
+        MemberProfile memberProfile = currentUserServices.findOrSaveUser(email, email, email);
```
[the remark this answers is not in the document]
sonndinh: Please see my comment for `LocalLoginController.java`. The input parameter `authReq` only contains username and password, which are the email and role, respectively, passed by the front-end. We can't get the first name/last name here as far as I understand.

server/src/main/java/com/objectcomputing/checkins/services/memberprofile/MemberProfileCreateDTO.java
sonndinh: Yeah, I was wondering if this would entail changes to the front-end. This is a create DTO which is sent by the front-end, right. Do you mean it should contain both the full name and its components (first, middle, last name, and suffix)?

server/src/main/java/com/objectcomputing/checkins/security/LocalLoginController.java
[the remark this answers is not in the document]
mkimberlin: That's only true if the user isn't in the loaded data. I believe that if the user is in the loaded test data, then this would cause that user not to be recognized as the user entering the system.

server/src/main/java/com/objectcomputing/checkins/security/LocalUserPasswordAuthProvider.java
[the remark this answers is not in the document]
mkimberlin: If I'm wrong about the above, then you are correct. If not, I think this will need to change.

You can validate it by trying to login locally with my email address or Holly's.

server/src/main/java/com/objectcomputing/checkins/services/memberprofile/MemberProfileCreateDTO.java
mkimberlin: Yes, exactly.

server/src/main/java/com/objectcomputing/checkins/security/LocalLoginController.java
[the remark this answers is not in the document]
sonndinh: You're right. I forgot about the loaded data. I've changed this to set the correct first and last names.

server/src/main/java/com/objectcomputing/checkins/security/LocalUserPasswordAuthProvider.java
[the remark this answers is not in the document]
sonndinh: The existing member profiles are loaded correctly. Only new member profiles with non-existing emails will have their first and last names equal to their emails.

mkimberlin (approved):

mkimberlin (approved):

moshirim (approved):

Issue #763, opened 2021-03-12
Error in Skills Report service (S)


mkimberlin:
## 400 Error
```
{
  "message": "Invalid skill level 4",
  "_links": {
    "self": {
      "href": "/reports/skills",
      "templated": false
    }
  }
}
```

## Example Payloads
Pretty much everything I tried...

```
{
  "skills": [{"id":"29017b81-2d32-4b7f-9f6b-d1b1f982d505","level":"INTERMEDIATE"}],
  "inclusive": true
}
```

```
{
  "skills": [{"id":"29017b81-2d32-4b7f-9f6b-d1b1f982d505","level":"NOVICE"}],
  "inclusive": true
}
```

```
{
  "skills": [{"id":"29017b81-2d32-4b7f-9f6b-d1b1f982d505","level":2}],
  "inclusive": true
}
```

```
{
  "skills": [{"id":"29017b81-2d32-4b7f-9f6b-d1b1f982d505","level":2}],
  "inclusive": false
}
```

PR #770, opened 2021-03-19
Fix skills report service

sonndinh:
I couldn't reproduce the error reported in issue #763; the service is working fine for me. However, I found a couple of other bugs that I fixed in this PR.

### DIFF
--- a/server/src/main/java/com/objectcomputing/checkins/services/memberprofile/MemberProfileRepository.java
+++ b/server/src/main/java/com/objectcomputing/checkins/services/memberprofile/MemberProfileRepository.java
@@ -15,37 +15,40 @@ import java.util.UUID;
 public interface MemberProfileRepository extends CrudRepository<MemberProfile, UUID> {
 
     @Query(value = "SELECT id, " +
-            "PGP_SYM_DECRYPT(cast(mp.name as bytea),'${aes.key}') as name, " +
-            "PGP_SYM_DECRYPT(cast(title as bytea),'${aes.key}') as title, " +
+            "PGP_SYM_DECRYPT(cast(mp.name as bytea), '${aes.key}') as name, " +
+            "PGP_SYM_DECRYPT(cast(title as bytea), '${aes.key}') as title, " +
             "pdlid, " +
-            "PGP_SYM_DECRYPT(cast(location as bytea),'${aes.key}') as location, " +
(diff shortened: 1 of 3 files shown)


moshirim (approved):

server/src/main/java/com/objectcomputing/checkins/services/memberprofile/MemberProfileRepository.java
```diff
@@ -15,37 +15,40 @@
...
+            "AND (:supervisorId IS NULL OR mp.supervisorId = :supervisorId) ", nativeQuery = true)
+    List<MemberProfile> search(@Nullable String name, @Nullable String title, @Nullable String pdlId,
+                               @Nullable String workEmail, @Nullable String supervisorId);
 
     List<MemberProfile> findAll();
 
-    String findNameById(@NotNull UUID id);
+    @Query(value = "SELECT PGP_SYM_DECRYPT(cast(mp.name as bytea), '${aes.key}') FROM member_profile as mp " +
+            "WHERE mp.id = :id", nativeQuery = true)
+    String findNameById(@NotNull String id);
```
mkimberlin, line 53: rather than changing this to String, I think I would have added an explicit cast to string in the where clause. this would be consistent with what is happening with the type hints in the entity objects.

mkimberlin (commented):
Would like to hear your thoughts on using the explicit cast in the query rather than changing the UUID to string in the code.

server/src/main/java/com/objectcomputing/checkins/services/memberprofile/MemberProfileRepository.java
sonndinh: The `id` column of the member_profile table is of type `varchar`. If I keep `UUID` as the method's parameter type, I would need to convert `UUID` to a string. How would you do that conversion in the query?

mkimberlin: `WHERE mp.id = cast(:id as varchar)` I believe

sonndinh: Okay, I'll try that.

mkimberlin (approved):

PR #777, opened 2021-03-26
feature-766/pdl-list-ordered-by-last-name

warnerj-oci:
Sort PDL list by last name

### DIFF
--- a/web-ui/src/api/member.js
+++ b/web-ui/src/api/member.js
@@ -10,6 +10,14 @@ export const getAllMembers = async (cookie) => {
   });
 };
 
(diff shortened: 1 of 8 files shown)


moshirim (approved):

mkimberlin (changes requested):
This will work...*but* I would do it in a [selector](https://github.com/oci-labs/check-ins/blob/develop/web-ui/src/context/selectors.js). Then it can be tested (yes, I know the other's aren't, but they should be). I can show you how to approach this if you need pointed in the right direction. Grab 20 min on my calendar if it would be helpful.

mkimberlin (approved):

PR #836, opened 2021-04-21
chore-810/convert-member-profile-service-url-to-plural

warnerj-oci:
Change service and web-ui urls to plural and update tests


moshirim (approved):

Maroju100:
member-profile is references at various places, did we change in all those places
Screen Shot 2021-04-22 at 12 07 01 PM

warnerj-oci:
There are three different stories for member-profile:  plain
member-profile, member-profile/current and member-profile/member-photo.
Between the three stories and pull requests, I do believe I got all of them.

Maroju100 (approved):

PR #869, opened 2021-05-10
Feature 802/create survey rest service

Maroju100:


moshirim (approved):

server/src/main/java/com/objectcomputing/checkins/services/survey/SurveyController.java
```diff
@@ -0,0 +1,116 @@
...
+import javax.validation.Valid;
+import javax.validation.constraints.NotNull;
+import java.net.URI;
+import java.util.Optional;
+import java.util.Set;
+import java.util.UUID;
+import java.util.concurrent.ExecutorService;
+
+
+@Controller("/services/survey")
```
warnerj-oci, line 26: Should the endpoint be plural to be consistent with the other endpoints?

PR #927, opened 2021-06-03
Changes for birthday service

bhatiam12:


server/src/main/resources/db/common/V53__alter_member_profile_table.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE member_profile
+ADD COLUMN birthDate date;
```
warnerj-oci, line 2: The name of this file will need to be changed, there was a V53 in the create settings merge

warnerj-oci (approved):

zsbrown97 (approved):

mkimberlin (approved):

PR #935, opened 2021-06-04
Feature 903/add contribution hours to profile page

zsbrown97:


web-ui/src/pages/ProfilePage.jsx
```diff
@@ -133,39 +154,55 @@ const ProfilePage = () => {
           />
         </CardContent>
       </Card>
+      <div>
```
mkimberlin, line 157: I don't think you need this outer `div`

mkimberlin (commented):
I'd probably kill off that outer div, but then I think you're good.

mkimberlin (approved):

moshirim (approved):

PR #971, opened 2021-06-08
Feature 853/create team skills report closes #853

### LINKED ISSUE #853: Create a team skills report (M)
Depends on #851 

Report should be empty by default (don't load results initially). It should be available under the "Reports -> Team Skill"

The top of the report should display a radar chart that shows each included employee, with the data indexed by employee name, and each included skill and level displayed in the report. (see [this example](https://codesandbox.io/s/nivo-basic-demo-forked-h04gq))

Below the radar chart, skill results should be displayed just like in the skills report.

There should be two modes in which this report can be used:
 - [x] existing team
 - [x] ad hoc team

These modes should be differentiated using a [Stepper](https://material-ui.com/components/steppers/) at the top of the page. (see [this example](https://codesandbox.io/s/material-demo-forked-rzwow?file=/demo.js))

The existing team mode should have a team selection drop down filter.

The ad hoc team mode should have a multiselect autocomplete for adding team members.

Results should be displayed for all members of the resulting team.

jahanner:
There are definitely a few areas in the code that will look a bit strange. This is because of a combination of the radar component not allowing spaces in some of the data(had to trim the member name) and teamMembers of a team having both an id and a memberId


web-ui/src/pages/TeamSkillReportPage.css
```diff
@@ -0,0 +1,72 @@
+.active {
```
mkimberlin, line 1: You need to do some prefixing on these names. These are too generic and will eventually have clash if they don't already. Either restrict them to only applying to things that are inside this component (set a class at the top level element and include it in your match patterns in this file) or make them very unlikely to clash (prefix the classnames themselves).

web-ui/src/pages/TeamSkillReportPage.jsx
```diff
@@ -0,0 +1,306 @@
...
+      );
+    } else {
+      setSearchResults([]);
+      setAllSearchResults([]);
+      setAllSearchResults([]);
+    }
+    setShowRadar(true);
+  };
+
+  function skillsToSkillLevel(skills) {
```
mkimberlin, line 66: You could turn this into a selector...not a must, just a thought.

yarn.lock
```diff
@@ -0,0 +1,276 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
mkimberlin, line 1: Why does this file exist at the top level? Did you accidentally yarn install there? This should go away.

mkimberlin (changes requested):

zsbrown97 (approved):

web-ui/src/pages/TeamSkillReportPage.jsx
```diff
@@ -0,0 +1,306 @@
...
+      setAllSearchResults(memberSkillsFound);
+      setSearchResults(
+        memberSkillsFound.filter((mSkill) =>
+          selectedMembers.some((member) => member.id === mSkill.id)
+        )
+      );
+    } else {
+      setSearchResults([]);
+      setAllSearchResults([]);
+      setAllSearchResults([]);
```
warnerj-oci, line 61: Is there a reason you setAllSearchResults twice here?

warnerj-oci (approved):

mkimberlin (approved):

PR #989, opened 2021-06-09
Feature 970/create card component feedback template

sagetheherb:
This PR creates a card component and styles it. 

Screen Shot 2021-06-09 at 3 00 21 PM


moshirim (approved):

pieperm (approved):
Was missing a Storybook, but that's been added with PR #1009, so it looks good to me

web-ui/src/components/template-card/TemplateCard.css
```diff
@@ -0,0 +1,53 @@
+.card{
```
mkimberlin, line 1: Be warned...these CSS classes will bleed. They are not magically keyed to this component like classes that are generated using the MUI templating.  You should likely prefix them all OR put a class on the top level DOM element for the class and change these to specify elements with the class `card` that are children of something that has the top level class

pieperm (approved):

web-ui/src/components/template-card/TemplateCard.css
sagetheherb: thanks! it should be fixed.

mkimberlin (approved):

PR #990, opened 2021-06-09
Feature 977/create confirmation page

moshirim:
Screen Shot 2021-06-08 at 3 34 15 PM


sagetheherb (approved):

zsbrown97:
Screen Shot 2021-06-10 at 10 54 45 AM

Once the confirmation screen is hit, should the Back and Finish buttons still be visible? If this is the last step in the process do we still need that part?

mkimberlin (commented):
I'm not a big fan of using an image for the checkmark.  I would have used either:
```
import CheckCircleIcon from '@material-ui/icons/CheckCircle';
```
or
```
import CheckCircleOutlineIcon from '@material-ui/icons/CheckCircleOutline';
```

zsbrown97 (approved):

smithj11 (commented):
Looks like change to icon is fine!

Issue #994, opened 2021-06-10
FE: Change stepper to three steps and a conformation page (S)


sagetheherb:
## DOD 

- [x] stepper is three bubbles long
- [x] submit is final stepper option
- [x] submit takes user to the submitted confirmation page

PR #1019, opened 2021-06-15
Feature 994/change stepper

JEdwardBrown:
Solved issues of #994


mkimberlin (approved):

smithj11 (approved):
Works with original step checks. Five stars!

PR #1028, opened 2021-06-17
Feature 888/retention report service

warnerj-oci:
Create service to generate data for retention reports


server/src/main/java/com/objectcomputing/checkins/services/memberprofile/retentionreport/RetentionReportServicesImpl.java
```diff
@@ -0,0 +1,352 @@
...
+@Singleton
+public class RetentionReportServicesImpl implements RetentionReportServices {
+
+    private final MemberProfileServices memberProfileServices;
+
+    public RetentionReportServicesImpl(MemberProfileServices memberProfileServices) {
+        this.memberProfileServices = memberProfileServices;
+    }
+
+    @Override
```
bhatiam12, line 21: do we need any validations here like only admin can run the report ?

warnerj-oci: Yes, we probably do. It wasn't in the AC, so I guess I missed it.

server/src/main/java/com/objectcomputing/checkins/services/memberprofile/retentionreport/RetentionReportServicesImpl.java
```diff
@@ -0,0 +1,352 @@
...
+                case WEDNESDAY:
+                    startDate = periodStartDate.plusDays(3);
+                    break;
+                case THURSDAY:
+                    startDate = periodStartDate.plusDays(2);
+                    break;
+                case FRIDAY:
+                    startDate = periodStartDate.plusDays(1);
+                    break;
+                default:
```
jahanner, line 63: Do you not need a SATURDAY in here? Or is the periodStartDate Sat by default?

warnerj-oci: Right, the default is Saturday, the other cases are adjusting to the Saturday of the week.

server/src/main/java/com/objectcomputing/checkins/services/memberprofile/retentionreport/RetentionReportServicesImpl.java
warnerj-oci: Added validation

jahanner (approved):

mkimberlin (approved):
eeeeexcellent... yeah...if this turn out the right numbers (which i think it should), I will do a happy dance.

PR #1029, opened 2021-06-17
Feature 919/create opportunities service

Maroju100:


moshirim (approved):

server/src/main/java/com/objectcomputing/checkins/services/opportunities/OpportunitiesServicesImpl.java
```diff
@@ -0,0 +1,98 @@
...
+    public Set<Opportunities> findByFields(String name, String description) {
+        Set<Opportunities> opportunitiesResponse = new HashSet<>();
+        opportunitiesResponseRepo.findAll().forEach(opportunitiesResponse::add);
+        if(name!=null){
+            opportunitiesResponse.addAll(opportunitiesResponseRepo.findByName(name));
+        }
+        else if(description!=null){
+            opportunitiesResponse.addAll(opportunitiesResponseRepo.findByDescription(description));
+        }
+        return opportunitiesResponse;
```
warnerj-oci, line 96: You are adding all opportunities here regardless of what the search string is. Did you mean to use the retain all feature in your conditional statements instead?

mkimberlin: @warnerj-oci I think you are probably right (that @Maroju100 meant to use retailAll here)...but the retainAll approach to these find by fields methods is also not desirable. We should be using the "is null or..." approach in our queries to push that to the database and not put us in the position of loading everything into memory and then filtering. Also, it allows for a single query rather than subsequent queries to the database to filter down.

mkimberlin (changes requested):
See my response to Joe's comment...need to use our is null or check on the queries instead of retainall approach

bhatiam12:
> See my response to Joe's comment...need to use our is null or check on the queries instead of retainall approach

@mkimberlin I changed it to use native query. Please review. 
@Maroju100 please verify the changes. I did run all the tests and they passed so I think I didn't break anything.

Maroju100:
> > See my response to Joe's comment...need to use our is null or check on the queries instead of retainall approach
> 
> @mkimberlin I changed it to use native query. Please review.
> @Maroju100 please verify the changes. I did run all the tests and they passed so I think I didn't break anything.

@bhatiam12 reviewed the code and checked tests at my end, native query implementation looks good to me.

bhatiam12 (approved):

server/src/main/java/com/objectcomputing/checkins/services/opportunities/OpportunitiesCreateDTO.java
```diff
@@ -0,0 +1,92 @@
...
+    private LocalDate expiresOn;
+
+
+    @NotNull
+    @Schema(required = true, description = "date for submittedOn")
+    private LocalDate submittedOn;
+
+    @NotNull
+    @Schema(required = true, description = "id of the associated member")
+    private UUID submittedBy;
```
mkimberlin, line 36: I should have noticed this before, but I don't think we want to send these in. This is whoever the current user is...and the current date.

server/src/main/java/com/objectcomputing/checkins/services/opportunities/OpportunitiesRepository.java
```diff
@@ -0,0 +1,31 @@
...
+    @Query(value = "SELECT id, " +
+            "PGP_SYM_DECRYPT(cast(name as bytea),'${aes.key}') as name, " +
+            "PGP_SYM_DECRYPT(cast(description as bytea),'${aes.key}') as description, " +
+            "PGP_SYM_DECRYPT(cast(url as bytea),'${aes.key}') as url, " +
+            "expiresOn, submittedOn, submittedBy,pending " +
+            "FROM opportunities op " +
+            "WHERE (:name IS NULL OR PGP_SYM_DECRYPT(cast(op.name as bytea),'${aes.key}') = :name) " +
+            "AND (:description IS NULL OR PGP_SYM_DECRYPT(cast(op.description as bytea),'${aes.key}') = :description) ", nativeQuery = true)
+    List<Opportunities> searchByValues( @Nullable String name,
+                                   @Nullable String description
```
mkimberlin, line 28: You know, I think we are going to want to search by submittedBy as well. :-( Sorry.

Maroju100: Added search by submittedBy.

mkimberlin (approved):

PR #1037, opened 2021-06-17
include team skills url to the logic check for open/closed sub lists

HootDunk:
The function I made initially to check whether a collapsible list should be open was hardcoded, so it didn't check for the newly added 'TEAM SKILLS' link.  I rewrote it so that it is no longer hard coded.  I also noticed that links weren't being rendered to the DOM with the <a></a> tag.  To fix this I used the Material UI component prop and passed the react-router-dom link component (https://material-ui.com/guides/composition/#link).

### DIFF
--- a/web-ui/src/components/menu/Menu.jsx
+++ b/web-ui/src/components/menu/Menu.jsx
@@ -1,5 +1,5 @@
 import React, { useContext, useEffect, useRef, useState } from "react";
-import { useHistory, useLocation } from "react-router-dom";
+import { useLocation, Link } from "react-router-dom";
 import { AppContext } from "../../context/AppContext";
 import { getAvatarURL } from "../../api/api";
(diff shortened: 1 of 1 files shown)


web-ui/src/components/menu/Menu.jsx
```diff
@@ -75,16 +75,22 @@ const useStyles = makeStyles((theme) => ({
...
+  ["/teams", "TEAMS"]
+]
+
+const reportsLinks = [
+  ["/checkins-reports", "CHECK-INS"], 
+  ["/skills-reports", "SKILLS"],
+  ["/team-skills-reports", "TEAM SKILLS"]
+]
+
+const isCollapsibleListOpen = (linksArr, loc) => {
```
jahanner, line 91: nice

Issue #1077, opened 2021-06-25
Issue POST request when submitting a feedback request (M)


pieperm:
## Definition of Done
- [x] New API file `feedback_request.js` for making calls to the backend
- [x] The `onClick` listener upon submission should make an API call
- [x] Validations for ensuring all required fields are defined before making POST
- [x] Ensure the database is updated from the POST

Depends on #1013

PR #1081, opened 2021-06-25
Feature 1057/implement validation for the feedback request process

Approve212:


JEdwardBrown (approved):

moshirim (approved):

zsbrown97 (approved):

web-ui/src/pages/FeedbackRequestPage.jsx
```diff
@@ -78,25 +82,31 @@ const FeedbackRequestPage = () => {
...
+  const getStep = useCallback(() => {
+    if(!stepQuery || stepQuery < 1 || !/^\d+$/.test(stepQuery))
+      return 1;
+    else return parseInt(stepQuery);
+  },[stepQuery]);
 
-  let sendDate = query?.sendDate ? query.sendDate: todayDate.toString();
-  let dueDate = query?.dueDate ? query.dueDate: null
-  let activeStep = location?.search ? parseInt(stepQuery) : 1;
-  const numbersOnly = /^\d+$/.test(stepQuery);
```
pieperm, line 89: Deleting `sendDate` and `dueDate` here prevents the front-end from building, as they are referenced below in `<SelectDate>`

pieperm (changes requested):
Merge conflicts

JEdwardBrown (approved):
Looks good!

web-ui/src/pages/FeedbackRequestPage.jsx
```diff
@@ -72,60 +76,98 @@ function getTemplates() {
...
+  const urlIsValid = useCallback(() => {
+    switch (activeStep) {
+      case 1:
+        return true;
+      case 2:
+        return hasTemplate();
+      case 3:
+        return hasTemplate() && hasFrom();
+      case 4:
+        return hasTemplate() && hasFrom() && hasDue();
```
pieperm, line 162: I don't think `case 4` will ever be reached, since `activeStep` only goes to 3 before we reroute the user to the confirmation page.

pieperm: This does bring up the issue that we have no validation for reaching the confirmation page at `/feedback/request/confirmation`. Maybe that should be a separate issue? I'm guessing that's why `handleSubmit` is left blank for now

web-ui/src/pages/FeedbackRequestPage.jsx
```diff
@@ -145,24 +187,18 @@ const FeedbackRequestPage = () => {
             <div>
               <Link
                 className={`no-underline-link ${activeStep <= 1 ? 'disabled-link' : ''}`}
-                to={getFeedbackArgs(activeStep - 1)}
+                to={`?step=${activeStep - 1}`}
```
pieperm, line 190: This now erases the selected template, recipients, dates, etc. from the URL when going back. Is this intended behavior?

pieperm: Additionally, if you select a template and then recipients, and then hit the back button, the page breaks and throws an error in the console: `Uncaught Error: FeedbackRequestPage(...): Nothing was returned from render. This usually means a return statement is missing. Or, to render nothing, return null.`

web-ui/src/pages/FeedbackRequestPage.jsx
```diff
@@ -72,60 +76,98 @@ function getTemplates() {
...
+  }, [fromQuery])
+
+  const isValidDate = useCallback((dateString) => {
+    const timestamp = Date.parse(dateString);
+    return !isNaN(timestamp);
+  }, []);
+
+  const hasDue = useCallback(() => {
+    return (dueQuery && isValidDate(dueQuery))
+  }, [dueQuery, isValidDate]);
```
pieperm, line 129: There is no validation here to make sure that the send date and the due date are on or later than the current date. We could potentially have a separate story for this since it would involve displaying error messages for the text fields.

web-ui/src/components/feedback_date_selector/SelectDate.jsx
```diff
@@ -18,75 +21,56 @@ pickerContain: {
...
+                    margin="normal"
+                    id="set-due-date"
+                    label="Due Date:"
+                    emptyLabel="No due date"
+                    value={dueDate}
+                    onChange={handleDueDateChange}
+                    KeyboardButtonProps={{
+                       'aria-label': 'change date',
+                    }}
+                 />
```
pieperm, line 72: There is currently no way to set the due date as "No due date" although that's likely going to be a separate issue since it might require an additional button on the input field.

pieperm: Also, the date inputs are filled in by default but not included in the URL (so they are not used in the validation process). To most users, it will appear that the dates are valid and they should be able to proceed, but the next button is grayed out. It may be more intuitive to have those date fields blank by default. Or we could leave the default values and add them to the URL for validation as soon as they enter the date picker step.

pieperm (changes requested):

web-ui/src/pages/FeedbackRequestPage.jsx
mkimberlin: This is a bug...I would argue for simply doing`history.goBack()`

web-ui/src/pages/FeedbackRequestPage.jsx
mkimberlin: Agreed, case 4 can be removed. I also don't think I would try to address the submit in this issue. That should definitely be split.

web-ui/src/pages/FeedbackRequestPage.jsx
mkimberlin: Definitely would split into another story. There probably needs to be some discussion about whether validation like that is a server or UI concern too.

web-ui/src/components/feedback_date_selector/SelectDate.jsx
mkimberlin: defaulting them to blank seems reasonable to me

web-ui/src/pages/FeedbackRequestPage.jsx
pieperm: Ok, that validation will be addressed in issue #1077

web-ui/src/pages/FeedbackRequestPage.jsx
pieperm: Split into a separate issue at #1098

mkimberlin (approved):

pieperm (approved):

PR #1096, opened 2021-06-30
Feature 1026/make mobile friendly

smithj11:
Rearranged layout on smaller devices to make mobile operations possible.


JEdwardBrown (approved):

pieperm (commented):
Looks good overall, though we may want to make the page title a little smaller on mobile screens:
Screen Shot 2021-06-30 at 5 07 04 PM

We may also want to consider in the future that a list layout may be more convenient than cards for selecting templates and recipients in a mobile view. For now the cards work pretty well though.

warnerj-oci (approved):

Issue #1099, opened 2021-06-30
FE: Change ad hoc template creation flow (S)


pieperm:
## Description
Currently it is too difficult to distinguish existing templates with the card for creating a new ad-hoc template. This can be simplified for the user by slightly changing how these elements are displayed.

## Definition of Done
- [x] Change ad-hoc card to be a separate button for creating a new ad-hoc template
- [x] New ad-hoc button should have a tooltip that says "Ask a single question"
- [x] The preview modal has a default title of "Ad Hoc"
- [x] The preview modal has a default description of "Ask a single question"
- [x] The default values of the preview modal can be changed

PR #1103, opened 2021-06-30
Feature 1099/change ad hoc template creation flow

pieperm:
Closes issue #1099


JEdwardBrown (commented):
Is FeedbackTemplateSelector what we are using now? It seems like every half the requests I see with FeedbackRequestPage in it has it, and the other one's do not.

pieperm:
> Is FeedbackTemplateSelector what we are using now? It seems like every half the requests I see with FeedbackRequestPage in it has it, and the other one's do not.

Yes, FeedbackTemplateSelector is the new way of rendering the template cards. The logic for showing the preview modal and rendering the list of cards has been moved into that component. Newer branches seem to have the refactor, while older branches tend to not have/use the component. Unfortunately this has resulted in a lot of merge conflicts due to hanging PRs.

moshirim (approved):

Maroju100 (approved):

Approve212 (approved):

Issue #1112, opened 2021-07-06
Make use of github actions environments for development and prod


mkimberlin:
See the feedback branch

PR #1113, opened 2021-07-06
Establish environments in GitHub Actions

mkimberlin:


.github/workflows/gradle-build-development.yml
```diff
@@ -1,7 +1,9 @@
 name: Gradle Build & Deploy - Develop
 on:
   push:
-    branches: [develop]
+    branches:
+      - 'develop'
+      - 'feature-1112/*'
```
bhatiam12, line 6: is this intentional ? don't we ned this script for all feature-* branches or this is only to build and test 1112 ?

mkimberlin: This only triggers for branches related to issue #1112. (basically this branch, I was just lazy.) I'll be removing that trigger before merge anyway.

zsbrown97 (approved):

PR #1157, opened 2021-07-08
Feature 1144/create view feedback component

smithj11:
To see the new page and components, either start Storybook and go to ViewFeedbackSelector OR fire up the frontend and go to /viewfeedback/selector


JEdwardBrown (approved):

Approve212 (approved):

HootDunk (approved):
Looks good to me.
I noticed the data for the ViewFeedbackSelector is currently hardcoded and imagine it will later be passed in as props.  If this was done so that there was some sample data for the component in Storybook then might I suggest doing it as shown here ->   https://github.com/oci-labs/check-ins/blob/develop/web-ui/src/components/kudos/Kudos.stories.js

This way you can have sample data passed in as props for displaying the component in Storybook while writing the component as it is intended to be used.   It also makes it easier to document the component in Storybook in situations where the component reacts conditionally based on the props that were passed to it (although in this case I don't think that's relevant).

zsbrown97 (approved):

bhatiam12 (approved):

bhatiam12:
@smithj11 there is a conflict on this branch.

PR #1176, opened 2021-07-09
Fetaure 1151/ created layout for feedback submission form.

moshirim:


web-ui/src/pages/FeedbackSubmitPage.jsx
```diff
@@ -1,14 +1,19 @@
...
+  const [showTips, setShowTips] = useState(true);
+
   return (
     <div className="feedback-submit-page">
-      <FeedbackSubmissionTips/>
+      {showTips ?
+        <FeedbackSubmissionTips onNextClick={() => setShowTips(false)} /> :
+        <FeedbackSubmitForm requesteeName={"JohnDoe"}/>
+      }
     </div>
```
JEdwardBrown, line 17: Wouldn't we want to dynamically pass in the requestee name here, similar to how we pass it in for the request?

pieperm: Yes, that is the eventual plan, but we've made a separate issue for that (feature #1178). This feature is more concerned with the view rather than the functionality.

JEdwardBrown: Okay, perfect.

JEdwardBrown (approved):

zsbrown97 (approved):

PR #1188, opened 2021-07-12
Feature 934/implementing cypress in the development environment

pieperm:
In this PR, Cypress was implemented and configured in the Check-ins application. It now runs as part of the Gradle build. 

If you'd like to see the Cypress tests run: use the command `npx Cypress run`. (Note: The test that runs is a dummy test - only to check the functionality of Cypress in the app and the Github workflow.)


.github/workflows/gradle-build-feature.yml
```diff
@@ -27,6 +41,23 @@ jobs:
...
+        run: ./gradlew dockerBuild
+      - name: Run docker container
+        run: ./gradlew dockerRun
+      - name: Cypress runs tests
+        uses: cypress-io/github-action@v2
+        with:
+          working-directory: web-ui
+          config: pageLoadTimeout=100000,baseUrl=http://localhost:8080
+          build: yarn build
+          start: yarn start
```
mkimberlin, line 54: if we are building and running a docker image above (and pointing to that running image), why do we need the yarn build and yarn start here?

server/build.gradle
```diff
@@ -1,8 +1,9 @@
 plugins {
     id("com.github.johnrengelman.shadow") version "7.0.0"
-    id("io.micronaut.application") version "1.5.3"
+    id("io.micronaut.application") version "1.4.2"
```
mkimberlin, line 3: Was this downgrade intentional?

moshirim: It was unintentional - thanks for pointing that out. I reverted it

.github/workflows/gradle-build-feature.yml
moshirim: It works without it. I deleted it. Thanks!

Bagurp:
Any chance we could squash some of the commits in this PR before merging it?

mkimberlin:
The login test does not complete. It shows as "pending" in the log:
https://github.com/oci-labs/check-ins/runs/4084677350?check_suite_focus=true

moshirim:
> The login test does not complete. It shows as "pending" in the log: https://github.com/oci-labs/check-ins/runs/4084677350?check_suite_focus=true

Yes that was intentional. I used `it.skip` so that it does not run. That test passes successfully in local but fails on remote because we are unable to access localhost 8080. I was planning on discussing that with you and how we can get around that problem when writing Cypress tests. I have created an issue for that as well. @mkimberlin

moshirim:
> Any chance we could squash some of the commits in this PR before merging it?

Addressing that now.

mkimberlin:
I think instead of localhost, you want to point it at the docker host. Try `host.docker.internal`

mkimberlin:
If that doesn't work, then try `172.17.0.1`

PR #1190, opened 2021-07-12
Feature 1178/add requestee to confirmation page

smithj11:
Confirmation page now dynamically displays requestee name, message dependent on send date, and recipient names

### DIFF
--- a/web-ui/src/components/feedback_request_confirmation/FeedbackRequestConfirmation.jsx
+++ b/web-ui/src/components/feedback_request_confirmation/FeedbackRequestConfirmation.jsx
@@ -1,46 +1,65 @@
-import React, {useContext} from "react";
+import React, { useContext } from "react";
 import Typography from "@material-ui/core/Typography";
 import CheckCircleIcon from '@material-ui/icons/CheckCircle';
-import {makeStyles} from '@material-ui/core/styles';
-import {selectProfile} from "../../context/selectors";
-import {AppContext} from "../../context/AppContext";
-import {useLocation, useHistory } from 'react-router-dom';
+import { makeStyles } from '@material-ui/core/styles';
+import { selectProfile } from "../../context/selectors";
+import { AppContext } from "../../context/AppContext";
+import { useLocation } from 'react-router-dom';
 import queryString from 'query-string';
 import DateFnsUtils from "@date-io/date-fns";
 import "./FeedbackRequestConfirmation.css";
-import {green} from "@material-ui/core/colors";
+import { green } from "@material-ui/core/colors";
 
 const dateUtils = new DateFnsUtils();
 
-const formatNamesAsList = (names) => {
-  return names.join(", ")
-}
 const useStyles = makeStyles({
-announcement: {
-  textAlign: "center"
-},
+  announcement: {
+    textAlign: "center"
+  },
 });
 
-    let today = new Date();
-    today = dateUtils.format(today, "yyyy-MM-dd");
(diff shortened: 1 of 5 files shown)


pieperm (commented):
Maybe we could use the ID of the feedback request returned from the POST as the URL parameter?

web-ui/src/components/feedback_request_confirmation/FeedbackRequestConfirmation.jsx
```diff
@@ -1,28 +1,65 @@
...
 const FeedbackRequestConfirmation = (props) => {
   const classes = useStyles();
+  const { state } = useContext(AppContext);
+  const location = useLocation();
+  const query = queryString.parse(location?.search);
+  const forQuery = query.for?.toString();
+  const fromQuery = query.from?.toString();
+  const sendQuery = query.send?.toString();
+  const requestee = selectProfile(state, forQuery);
+  let recipientInfo = getRecipientNames();
```
jahanner, line 33: this doesn't complain about being called before initialization?

JEdwardBrown (approved):

web-ui/src/components/feedback_request_confirmation/FeedbackRequestConfirmation.jsx
HootDunk: Because it's defined with the function keyword it gets 'hoisted' so you can call it before its defined in the file without error.  Arrow functions aren't hoisted but functions defined with the function keyword are.  Basically JS is weird haha.

jahanner (approved):

web-ui/src/components/feedback_request_confirmation/FeedbackRequestConfirmation.jsx
```diff
@@ -1,28 +1,65 @@
...
+
+      } else {
+        recipientProfiles.push(fromQuery)
+
+      }
+      return recipientProfiles;
+
+    }
+
+  }
```
pieperm, line 63: Would it be more convenient to move the calls to `selectProfile` into the loop in this function? You could then return an array of the names instead an array of all the profile information.

web-ui/src/components/feedback_request_confirmation/FeedbackRequestConfirmation.jsx
```diff
@@ -1,28 +1,65 @@
...
   return (
     <div className="request-confirmation">
-      <CheckCircleIcon style={{ color: green[500] , fontSize: '40vh' }}>checkmark-image</CheckCircleIcon>
-      <Typography className={classes.announcement} variant="h3"><b>Feedback scheduled for today</b></Typography>
-      <Typography className="recipients-list" variant="h6"><b>From:</b> {formatNamesAsList(["Alice", "Bob", "Eve"])}</Typography>
+      <CheckCircleIcon style={{ color: green[500], fontSize: '40vh' }}>checkmark-image</CheckCircleIcon>
+      <Typography className={classes.announcement} variant="h3"><b>Feedback request {sendQuery > today ? " scheduled on: " + sendQuery : " sent"} for {requestee?.name} </b></Typography>
+      <Typography className="recipients-list" variant="h6"><b>Sent to: </b>
+        {recipientInfo?.map((member, index) =>
+          `${selectProfile(state, member)?.name}${index === recipientInfo.length - 1 ? "" : ', '}`
```
pieperm, line 70: We could add "and" here to go before the last element as well (if you want)

PR #1220, opened 2021-07-15
Feature 1166/support markdown in notes

HootDunk:
I added the ReactQuill dependency which is a React component built around Quill.js.  It is an WYSIWYG editor that will allow for better formatting when writing notes.  ReactQuill is MIT licensed and Quill.js has the BSD 3-Clause "New" or "Revised" License.  Both allow for private and commercial use. 

To test it out, create a new check-in and include some notes.  The component is also in storybook.

### DIFF
--- a/web-ui/src/components/notes/Note.jsx
+++ b/web-ui/src/components/notes/Note.jsx
@@ -128,13 +128,3 @@ const Notes = (props) => {
 };
 
 export default Notes;
-
-
-{/* <textarea
-disabled={
-  currentCheckin?.completed ||
-  note === undefined || Object.keys(note) === 0
-}
-onChange={handleNoteChange}
-value={note && note.description ? note.description : ""}
-/> */}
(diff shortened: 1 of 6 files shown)


web-ui/src/components/notes/Note.jsx
```diff
@@ -85,41 +85,43 @@ const Notes = (props) => {
...
       return newNote;
     });
-  };
+  }
 
   return (
       <Card className="note">
         <CardHeader avatar={<NotesIcon />} title={`Notes for ${currentMember?.name}`} titleTypographyProps={{variant: "h5", component: "h2"}} />
         <CardContent>
-          <div className="container">
```
jahanner, line 104: Just heads up, you'll probably have to update the snapshot tests for these files

jahanner (approved):

web-ui/src/components/notes/Note.jsx
HootDunk: I'm glad you mentioned that.  I had run jest test and all tests passed but that is definitely not what should have happened. Looking into it now.

bhatiam12 (approved):

PR #1235, opened 2021-07-19
Feature 1169/template url validation

sagetheherb:


JEdwardBrown (approved):

moshirim (approved):

web-ui/src/pages/FeedbackRequestPage.jsx
```diff
@@ -201,19 +244,22 @@ const handleSubmit = () =>{
...
+          case 3:
+            return hasFor() && templateIsValid && hasFrom();
+          case 4:
+            return hasFor() && templateIsValid && hasFrom() && hasSend();
+          default:
+            return false;
+        }
       }
-    }, [activeStep, hasFor, hasTemplate, hasFrom, hasSend]);
+      return true;
```
pieperm, line 261: In this case, couldn't I omit URL parameters (e.g. I don't put in a `for` parameter) and still be allowed to get to step 1? It seems like we should not be able to get to step 1 without having `for` in the URL.

pieperm (approved):
I noted a potential issue in another comment. Other that that, it looks good!

jahanner (approved):

PR #1237, opened 2021-07-19
Feature 1225/create view responses layout

smithj11:
Go to Storybook and go to "ViewFeedbackResponses" to see the sample look of this component.

### DIFF
--- a/web-ui/src/components/view_feedback_responses/ViewFeedbackResponses.jsx
+++ b/web-ui/src/components/view_feedback_responses/ViewFeedbackResponses.jsx
@@ -1,71 +1,19 @@
 import React from 'react';
 import Grid from '@material-ui/core/Grid';
-import { Link } from "react-router-dom";
-import Divider from '@material-ui/core/Divider';
 import TextField from '@material-ui/core/TextField';
 import { Typography } from '@material-ui/core';
 import "./ViewFeedbackResponses.css";
 import { makeStyles } from '@material-ui/core/styles';
 import Card from '@material-ui/core/Card';
 import CardContent from '@material-ui/core/CardContent';
-import CardActions from '@material-ui/core/CardActions';
 import AvatarComponent from '../avatar/Avatar'
 import MoodBadIcon from '@material-ui/icons/MoodBad';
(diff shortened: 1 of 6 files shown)


moshirim (approved):

pieperm (approved):

zsbrown97 (approved):

web-ui/src/components/view_feedback_responses/ViewFeedbackResponses.jsx
```diff
@@ -0,0 +1,164 @@
...
+        paddingBottom: '16px',
+      }
+    }
+  }, { name: "MuiCardContent" })
+
+
+
+
+const ViewFeedbackResponses = (props) => {
+     useStylesCardContent();
```
HootDunk, line 28: I believe this needs assignment to a variable to use it in the component.

web-ui/src/components/view_feedback_responses/ViewFeedbackResponses.jsx
```diff
@@ -0,0 +1,164 @@
...
+    
+        {
+            answer: "I always thought they were sort of nasty creatures...",
+            questionId: 2,
+            responderName: "Erin Deeds",
+            sentiment: 0.2
+    
+        },
+        {
+            answer: "Opossums are very misunderstood. People think they are dirty and diseased, but their drooling and hissing is a defense mechanism. They eat all kinds of pests, like ticks and mice, keeping disease down. They are wonderful critters!",
```
HootDunk, line 75: Opossums are definitely misunderstood.

smithj11: Glad I could educate someone on the wonders of the North American opossum!

web-ui/src/components/view_feedback_responses/ViewFeedbackResponses.jsx
smithj11: It actually complains if I assign the styles hook to a variable and never use that variable. Just calling the hook in a component is enough to override the default Material UI classes that it attached to the card component automatically.

HootDunk (approved):

PR #1308, opened 2021-07-26
Feature 1275/prevent recipient access to feedback request before send date 

sagetheherb:
this PR prevents the user from accessing a request for a date that is before the current date. To test try changing the test data's date to be after today's date and try to access the request.


server/src/main/java/com/objectcomputing/checkins/services/feedback_request/FeedbackRequestServicesImpl.java
```diff
@@ -186,9 +186,14 @@ private boolean createIsPermitted(@NotNull UUID requesteeId) {
         return isAdmin || currentUserId.equals(requesteePDL);
     }
 
-    private boolean getIsPermitted(@NotNull UUID requesteeId, @NotNull UUID recipientId) {
-        UUID currentUserId = currentUserServices.getCurrentUser().getId();
-        return createIsPermitted(requesteeId) || currentUserId.equals(recipientId);
+    private boolean getIsPermitted(@NotNull UUID requesteeId, @NotNull UUID recipientId, LocalDate sendDate) {
+        LocalDate today = LocalDate.now();
+        if(sendDate.isBefore(today)){
+            return createIsPermitted(requesteeId) || currentUserServices.getCurrentUser().getId().equals(recipientId);
```
pieperm, line 192: This prevents admins and the PDL from being able to get a feedback request before the send date. We really only want this restriction to apply to the recipient.

If you change this, it might also be a good idea to write a couple more tests to ensure that:
- An admin can access it before the send date
- The requestee's PDL can access it before the send date
- The recipient *cannot* access it before the send date (this test already exists)

pieperm (changes requested):

server/src/main/java/com/objectcomputing/checkins/services/feedback_request/FeedbackRequestServicesImpl.java
sagetheherb: thanks for this ill fix it!

HootDunk (approved):

jahanner (approved):

pieperm (approved):

moshirim (approved):

PR #1312, opened 2021-07-27
Closes Bugfix 1304/fix styling misalignments

Approve212:


pieperm:
This feature fixes a visual bug on the check-ins paged introduced by a previously merged pull request on the `feedback` branch (due to bleeding styles). Fixing this bug involved a small refactor that splits a feedback response into its own component. 

Additionally, as part of this issue, the list of feedback responses should now look much less squished on mobile. We also added snapshot tests and added the component to Storybook (under `FeedbackResponseCard`)

moshirim (approved):

Issue #1298, opened 2021-07-26
Refactor feedback request to reference feedback template (S)


pieperm:
Depends on #1297 

## Definition of Done
- [x] A feedback request references a specific template ID
- [x] Adjust integration tests accordingly

Issue #1296, opened 2021-07-26
BE: Revert frozen template questions and frozen templates [S]


smithj11:
## Definition of Done
- [x] Delete frozen template question model and related implementations from module
- [x] Delete frozen template question table from database
- [x] Delete frozen template model and related implementations from module
- [x] Delete frozen template table from database
- [x] Change feedback answers to include a request id
- [x] Change feedback answer table to reference question id of template
- [x] Change related migrations to edit already existing feedback answer tables
- [x] Change related tests to accommodate new data models

PR #1315, opened 2021-07-28
Feature 1300/alter template backend update

### LINKED ISSUE #1300: BE: Alter template backend to not have edit endpoint (XS)
Depends on #1298 

## Definition of Done
- [x] Make template not editable
- [x] Make template questions not editable
- [x] Update tests accordingly

pieperm:
Closes #1300 

This feature is another installment of the major refactors started in issues #1296 and #1298. The goal of this issue is to remove the ability to update certain fields of templates, including the template title and description. 

The only field that can be updated is `active`, which is a method for soft-deleting templates. When a template is soft-deleted, it will still show up as part of past feedback requests, but it will be inaccessible for creating new feedback requests (unless it is reactivated)

### DIFF
--- a/server/src/main/java/com/objectcomputing/checkins/services/feedback_template/FeedbackTemplateServicesImpl.java
+++ b/server/src/main/java/com/objectcomputing/checkins/services/feedback_template/FeedbackTemplateServicesImpl.java
@@ -3,7 +3,6 @@ package com.objectcomputing.checkins.services.feedback_template;
 import com.objectcomputing.checkins.exceptions.BadArgException;
 import com.objectcomputing.checkins.exceptions.NotFoundException;
 import com.objectcomputing.checkins.exceptions.PermissionException;
-import com.objectcomputing.checkins.services.feedback_template.template_question.*;
 import com.objectcomputing.checkins.services.memberprofile.MemberProfileServices;
 import com.objectcomputing.checkins.services.memberprofile.currentuser.CurrentUserServices;
 import com.objectcomputing.checkins.util.Util;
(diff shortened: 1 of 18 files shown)


server/src/main/java/com/objectcomputing/checkins/services/feedback_template/FeedbackTemplateServicesImpl.java
```diff
@@ -106,32 +106,30 @@ public FeedbackTemplate getById(UUID id) {
...
     @Override
     public List<FeedbackTemplate> findByFields(@Nullable UUID creatorId, @Nullable String title) {
         if (!getIsPermitted()) {
             throw new PermissionException("You are not authorized to do this operation");
         }
 
         List<FeedbackTemplate> templateList = new ArrayList<>();
-        // Filters only active templates by default
+
         if (title != null) {
```
bhatiam12, line 117: This logic can be simplified if you write custom query in repository method. then you don't need to use retainall or addall at all and query will do all the work. Take a look at the query in OpportunitiesRepository.java. We can discuss if you have questions.

bhatiam12:
This has same code coverage issue, we will need to spend some time to add tests otherwise it will keep happening. @mkimberlin Please can we take a look at quality gates and see how we can tweak UI test coverage.

server/src/main/java/com/objectcomputing/checkins/services/feedback_template/FeedbackTemplateServicesImpl.java
```diff
@@ -99,53 +82,22 @@ public FeedbackTemplate getById(UUID id) {
...
-        if (!getIsPermitted()) {
-            throw new PermissionException("You are not authorized to do this operation");
-        }
 
         return feedbackTemplate.get();
     }
 
     @Override
     public List<FeedbackTemplate> findByFields(@Nullable UUID creatorId, @Nullable String title) {
-        if (!getIsPermitted()) {
```
bhatiam12, line 112: why did we remove this validation ?

smithj11: getIsPermitted only checks if there is an authenticated user. However, we determined that the backend throws an error if there is not an authenticated user before it even hits this service implementation, making it a moot point.

bhatiam12 (approved):

moshirim (approved):

PR #1328, opened 2021-07-29
Feature 1150/send email notifications to guild leaders when changes are made to membership

HootDunk:
This will send emails to guild leads when changes are made to guild membership that they did not make themselves.  Tests that mock the sendEmail function are included.

### DIFF
--- a/server/src/main/java/com/objectcomputing/checkins/notifications/email/EmailSender.java
+++ b/server/src/main/java/com/objectcomputing/checkins/notifications/email/EmailSender.java
@@ -1,5 +1,5 @@
 package com.objectcomputing.checkins.notifications.email;
 
 public interface EmailSender {
-    public void sendEmail(String subject, String content);
+    public void sendEmail(String subject, String content, String... recipients);
 }
--- a/server/src/test/java/com/objectcomputing/checkins/services/guild/GuildControllerTest.java
+++ b/server/src/test/java/com/objectcomputing/checkins/services/guild/GuildControllerTest.java
@@ -29,8 +29,7 @@ import java.util.*;
 import static com.objectcomputing.checkins.services.role.RoleType.Constants.*;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertThrows;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.*;
 
 class GuildControllerTest extends TestContainersSuite implements GuildFixture,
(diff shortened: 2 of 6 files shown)


moshirim (approved):

server/src/main/java/com/objectcomputing/checkins/notifications/email/EmailSender.java
```diff
@@ -1,5 +1,5 @@
 package com.objectcomputing.checkins.notifications.email;
 
 public interface EmailSender {
-    public void sendEmail(String subject, String content);
+    public void sendEmail(String subject, String content, String... recipients);
```
jahanner, line 4: should this be String ...recipients or am I reading that wrong?

HootDunk: The three dots in java are called Varargs and they work a bit different then the spread operator for JavaScript.  I saw it being used this way in one of the commits in the feedback branch and had to look it up myself (it seemed like a good use though so I repeated it).  Java treats the recipients variable as an array of strings but when you call the method, if you only have one recipient, you can pass the string directly and skip putting it into an array. You can also explicitly pass multiple String recipients and Java will convert it to an array for you. Of course you can also just pass in an array as well if that's what you're working with already. 

So all of the following are valid ways to call the method:
sendEmail (subject, content, "someEmail@gmail.com");
sendEmail (subject, content, "someEmail@gmail.com", "anotherEmail@gmail.com");

String[] emails = new String[] {"someEmail@gmail.com", "anotherEmail@gmail.com", "aThirdEmail@gmail.com"};
sendEmail(subject, content, emails);

jahanner: Nice, I'll have to keep Varargs in mind then

server/src/test/java/com/objectcomputing/checkins/services/guild/GuildControllerTest.java
```diff
@@ -17,14 +18,18 @@
 import io.micronaut.http.client.HttpClient;
 import io.micronaut.http.client.annotation.Client;
 import io.micronaut.http.client.exceptions.HttpClientResponseException;
+import jnr.ffi.annotations.In;
```
mkimberlin, line 21: I'm guessing this was an accident?

mkimberlin (approved):

server/src/test/java/com/objectcomputing/checkins/services/guild/GuildControllerTest.java
HootDunk: Yep, will remove.

PR #1331, opened 2021-07-29
added basic error boundary

jahanner:
To test it just put `throw "Some stuff"` somewhere in a frontend component and the error boundary should handle it. You can then create a git issue for the problem on that page

### DIFF
--- a/.github/workflows/gradle-build-production.yml
+++ b/.github/workflows/gradle-build-production.yml
@@ -2,7 +2,7 @@ name: Gradle Build & Deploy - Production
 on:
   push:
     branches:
-      - 'master'
+      - "master"
 
 env:
   PROJECT_ID: ${{ secrets.RUN_PROJECT }}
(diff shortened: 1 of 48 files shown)


Approve212 (approved):

HootDunk (approved):

moshirim (approved):

.github/workflows/gradle-build-production.yml
```diff
@@ -2,7 +2,7 @@ name: Gradle Build & Deploy - Production
 on:
   push:
     branches:
-      - 'master'
+      - "master"
```
bhatiam12, line 5: any reason why these single quotes were changed ?

jahanner: No that's just prettier formatting from when I added the new env variable in there and saved. Shouldn't change anything but I can revert that if you want

bhatiam12 (approved):

PR #1379, opened 2021-08-05
Feature 1361/backend to make templates private or public

### LINKED ISSUE #1361: Backend to make templates private or public (S)
## Definition of Done
- [x] New column on feedback template that determines whether a template is private or public
- [x] A private template can only be used/seen by the template creator and admins
- [x] Consequently, only members who have access to a private template can use that template when creating a feedback request
- [x] A public template can be used by any authenticated member
- [x] Database migration
- [x] Adjust testing data to include examples of both private and public templates
- [x] Integration tests

pieperm:
To test this, start up the back-end, log in as Geetika, and go to:
```
localhost:3000/feedback/request/?for=1b4f99da-ef70-4a76-9b37-8bb783b749ad
```

There should be two private templates shown on the page: "Private template" and "Private template 2". If you then log in as a regular member who is not Geetika, those templates should not be visible on that page (because they are private templates created by Geetika). If you have an admin role, you should be able to see the private templates regardless.

Closes #1361


sagetheherb:
I am still seeing the private templates while logged in as myself or jessie.

moshirim (approved):

jahanner (approved):

PR #1388, opened 2021-08-06
Feature 1334/search responses on page

smithj11:
To see this feature in action:
1. Uncomment line 116 and comment out line 117 in ViewFeedbackResponses.jsx
2. Activate backend and frontend
3 . Log in as Geetika
4. Go to localhost:3000/feedback/view/responses
5. Search something :)


moshirim (approved):

pieperm (commented):
The search works well, but I think the original intent was to make it more of a filter than a general search:
Screen Shot 2021-08-10 at 3 21 08 PM

I don't think these are mutually exclusive though. We could keep the search bar for finding keywords in the responses, while using a separate field to filter by recipient(s)

pieperm:
> The search works well, but I think the original intent was to make it more of a filter than a general search:
> 
> I don't think these are mutually exclusive though. We could keep the search bar for finding keywords in the responses, while using a separate field to filter by recipient(s)

This has been addressed in pull request #1424

PR #1406, opened 2021-08-09
Feature 1374/fix url validations on feedback request flow

pieperm:
To test this, go to:
```
localhost:3000/feedback/request
```
You should get an error toast regarding the URL.

If you go to:
```
localhost:3000/feedback/request/?for=1b4f99da-ef70-4a76-9b37-8bb783b749ad
```
then that URL should work. Stepping through the request process should not produce any errors. If you change anything in the URL to be invalid (e.g. cut off part of an ID) then you should get an error toast (it will either redirect you or remove that specific query, depending on which ID you altered)

Additionally, for any of the URLs that are generated when stepping through (meaning you didn't manually alter the URL), you should be able to reload the page and return to the same screen with no redirect, errors, or toasts.

### DIFF
--- a/web-ui/src/pages/FeedbackRequestPage.jsx
+++ b/web-ui/src/pages/FeedbackRequestPage.jsx
@@ -1,4 +1,4 @@
-import React, {useContext, useCallback, useEffect, useState} from "react";
+import React, {useContext, useCallback, useEffect, useState, useRef} from "react";
 import { makeStyles } from "@material-ui/core/styles";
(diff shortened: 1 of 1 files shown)


pieperm:
Side note, I changed two `switch` statements to `if-else` blocks. This (somehow) resolved an editor warning that certain `case` branches were unreachable, even though they were actually being reached.

JEdwardBrown (approved):

HootDunk (approved):

moshirim (approved):

PR #1408, opened 2021-08-09
Feature 1282/filter feedback requests by date range on view page

pieperm:
To test this, log in as Geetika and go to:
```
localhost:3000/feedback/view
```

You should now be able to filter responses by the date they were sent (includes all requests within the selected time range). The current options for this are 3 months, 6 months, 1 year, and all time. This filter should work with the existing sort selector and the search field.


pieperm:
I'm also curious if there's a way to not duplicate the `SortOption` and `DateRange` objects (pseudo-enums). Attempted to `export` them from one file and `import` them in the other file, but since they are used in `PropTypes`, this results in an initialization error.

moshirim (approved):

jahanner (approved):

PR #1416, opened 2021-08-11
Bugfix 1415/fix requestee name on view responses

smithj11:


smithj11:
Changes rolled into 1418

PR #1473, opened 2021-09-02
Bugfix 1460/cannot remove self from guild

jahanner:


mkimberlin:
You've got some bleed from your other PR in here, but it's ok. That one is good too. :-)

HootDunk (approved):

PR #1494, opened 2021-09-09
Checkbox to include terminated members

Maroju100:


web-ui/src/components/admin/users/Users.jsx
```diff
@@ -114,6 +119,15 @@ const Users = () => {
             </div>
           )}
         </Grid>
+          <div className="checkbox-row">
+            <label htmlFor="includeterminated">Include Terminated Members</label>
```
jahanner, line 123: you'll need to update the snapshot test for this change and you should be good

jahanner (approved):

moshirim (approved):

PR #1505, opened 2021-09-16
Feature 1451/create permission endpoint

HootDunk:


moshirim (approved):

server/src/test/java/com/objectcomputing/checkins/services/permissions/PermissionsControllerTest.java
```diff
@@ -0,0 +1,91 @@
...
+        assertEquals(HttpStatus.OK, response.getStatus());
+        assertTrue(response.getBody().isPresent());
+        assertEquals(response.getBody().get(), list);
+    }
+
+    @Test
+    void testGetAllPermissionsEnsureAlphabeticalOrder() {
+        Permission lastPermission = createACustomPermission("z");
+        Permission firstPermission = createACustomPermission("A");
+        Permission middlePermisison = createACustomPermission("H");
```
Bagurp, line 54: Typo here

server/src/test/java/com/objectcomputing/checkins/services/permissions/PermissionsControllerTest.java
```diff
@@ -0,0 +1,91 @@
...
+    void testGetPermissionsNoneExistsReturnsEmptyList(){
+        final HttpRequest<Object> request = HttpRequest.
+                GET("/").basicAuth(MEMBER_ROLE, MEMBER_ROLE);
+
+        final HttpResponse<List<Permission>> response =
+                client.toBlocking().exchange(request, Argument.listOf(Permission.class));
+
+        assertEquals(HttpStatus.OK, response.getStatus());
+        Assertions.assertTrue(response.getBody().isPresent());
+        assertEquals(response.getBody().get(), List.of());
```
Bagurp, line 77: maybe assertTrue(response.getBody().get().isEmpty())?

Bagurp (approved):

Bagurp:
@HootDunk, looks like adding Swagger documentation is part of AC. Will swagger documentation be automatically updated or should we manually write it for the endpoints?

server/src/test/java/com/objectcomputing/checkins/services/permissions/PermissionsControllerTest.java
HootDunk: Yeah that's a better way doing that, will update.

HootDunk:
> @HootDunk, looks like adding Swagger documentation is part of AC. Will swagger documentation be automatically updated or should we manually write it for the endpoints?

It gets automatically generated as long as the right comments and annotations are in place.  For controllers, you place an @Tag annotation at the class level which groups the methods together under whatever name was given as the value for the @Tag.   Then the comments above each controller method are how swagger is able to generate the documentation for each method.  For entities you simply add an @Schema annotation over each field to describe it's purpose and it will be included in swagger with the other entities on the bottom.

PR #1593, opened 2021-11-16
revert to prior conditional

HootDunk:


server/src/main/java/com/objectcomputing/checkins/services/feedback_request/FeedbackRequestServicesImpl.java
```diff
@@ -122,7 +122,7 @@ public FeedbackRequest update(FeedbackRequestUpdateDTO feedbackRequestUpdateDTO)
         boolean dueDateUpdateAttempted = !Objects.equals(originalFeedback.getDueDate(), feedbackRequest.getDueDate());
         boolean submitDateUpdateAttempted = !Objects.equals(originalFeedback.getSubmitDate(), feedbackRequest.getSubmitDate());
 
-        if(currentUserId.equals(originalFeedback.getRecipientId()) && (dueDateUpdateAttempted || submitDateUpdateAttempted)) {
+        if(dueDateUpdateAttempted && !updateDueDateIsPermitted(feedbackRequest)) {
```
HootDunk, line 125: I just switched this back to what it had been previously. The conditional was being triggered because the submitDateUpdateAttempted value was true but this is to be expected as it gets updated when a user submits their feedback.

mkimberlin (approved):

PR #1617, opened 2022-01-26
Bug where user modal did not clear is fixed.

smithj11:
Also added some preliminary filtering to the user's input on the modal to provide feedback alerts in case they do not fill out all required fields, preventing the case where an admin tries to submit a new user without all fields and receives no indication of success or failure. Input checking/feedback entering will have to be expanded upon in another issue.

### DIFF
--- a/web-ui/src/components/admin/users/Users.jsx
+++ b/web-ui/src/components/admin/users/Users.jsx
@@ -95,13 +95,12 @@ const Users = () => {
           />
           {isAdmin && (
             <div className="add-member">
-              <Button startIcon={<PersonIcon />} onClick={()=>handleOpen(true)}>
+              <Button startIcon={<PersonIcon />} onClick={handleOpen}>
                 Add Member
               </Button>
 
               <MemberModal
                 member={{}}
-                isNewMember={true}
                 open={open}
                 onClose={handleClose}
                 onSave={async (member) => {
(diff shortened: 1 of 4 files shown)


web-ui/src/components/admin/users/Users.jsx
```diff
@@ -97,11 +95,13 @@ const Users = () => {
           />
           {isAdmin && (
             <div className="add-member">
-              <Button startIcon={<PersonIcon />} onClick={handleOpen}>
+              <Button startIcon={<PersonIcon />} onClick={()=>handleOpen(true)}>
```
mkimberlin, line 98: This line should be changed back. Handle open doesn't take a parameter, and doing this (inline arrow) will create a new function for every re-rendering.

web-ui/src/components/member-directory/MemberModal.jsx
```diff
@@ -41,6 +57,30 @@ const MemberModal = ({ member = {}, open, onSave, onClose }) => {
...
+      dispatch({
+        type: UPDATE_TOAST,
+        payload: {
+          severity: "error",
+          toast: "One or more required fields are empty. Check starred input fields",
+        },
+      });
+    }
+  }
+
```
mkimberlin, line 83: Since these are defined inside the render function, you want to create them with `useCallback`, otherwise a new method is created with every render (state changes, property changes, etc), even when it's not necessary.

mkimberlin (changes requested):
Couple minor changes...see my comments.

web-ui/src/components/admin/users/Users.jsx
smithj11: I've altered this function back to the original form.

crklinc (approved):

mkimberlin (approved):

PR #1619, opened 2022-01-26
Bugfix 1608/fix xss quill vulnerability

smithj11:
Notes, Private Notes, and Error Boundary are now sanitized of potentially malicious content and disallowed tags by a function in "helpers" called "sanitizeQuillElements." Let me know if you want sanitization to be moved to somewhere in the API calls or wherever else. Note that React escapes html script automatically, so it can be a bit harder to launch a test attack.

### DIFF
--- a/web-ui/package.json
+++ b/web-ui/package.json
@@ -58,6 +58,7 @@
     "react-scripts": "4.0.3",
     "react-test-renderer": "^17.0.2",
     "reselect": "^4.0.0",
+    "sanitize-html": "^2.6.1",
     "styled-components": "^5.3.1",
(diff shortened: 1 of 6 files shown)


mkimberlin:
Am I missing something, or did you not finish with PrivateNote? It just has a console.log and not the sanitize call from what I can see. Your approach is on target though!

smithj11:
> Am I missing something, or did you not finish with PrivateNote? It just has a console.log and not the sanitize call from what I can see. Your approach is on target though!

I fixed this issue! There should be a sanitizer in private note now.

mkimberlin (approved):

PR #1623, opened 2022-01-28
Added supervisor and PDL info to the member profile page

smithj11:
The member profile page now displays supervisor and PDL in the name card per request. Please let me know if you want more information (like birthday) to be displayed or if the supervisor/PDL info should be moved.

### DIFF
--- a/web-ui/src/pages/MemberProfilePage.jsx
+++ b/web-ui/src/pages/MemberProfilePage.jsx
@@ -9,7 +9,10 @@ import { getGuildsForMember } from "../api/guild";
 import { getAvatarURL } from "../api/api.js";
 import ProfilePage from "./ProfilePage";
 import { levelList } from "../context/util";
-import { getMember } from "../api/member.js";
+import {
+  selectOrderedPdls,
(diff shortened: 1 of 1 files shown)


web-ui/src/pages/MemberProfilePage.jsx
```diff
@@ -29,6 +30,37 @@ const MemberProfilePage = () => {
...
+        let supervisorProfile =
+          res.payload.data && !res.error ? res.payload.data : undefined;
+        setSupervisor(supervisorProfile ? supervisorProfile.name : "");
+      }
+    }
+    if (csrf) {
+      getSupervisorName();
+    }
+  }, [csrf, selectedMember]);
+
```
mkimberlin, line 63: All of the profiles get loaded into the context on load, you can just select those instead of having to make these calls.

smithj11: Fixed. Supervisor and PDL info now selected out of state.

mkimberlin (approved):

mkimberlin (approved):

PR #1627, opened 2022-02-01
Made feedback request cards wrap and space information better on mobile sizes

smithj11:
Mobile/tablet sizes should no longer suffer from cramped content or misaligned buttons when viewing Feedback Request Subcards.

### DIFF
--- a/web-ui/src/components/feedback_request_card/feedback_request_subcard/FeedbackRequestSubcard.jsx
+++ b/web-ui/src/components/feedback_request_card/feedback_request_subcard/FeedbackRequestSubcard.jsx
@@ -3,7 +3,7 @@ import { styled } from "@mui/material/styles";
 import PropTypes from "prop-types";
 import Grid from "@mui/material/Grid";
 import Typography from "@mui/material/Typography";
-import {Link} from "react-router-dom";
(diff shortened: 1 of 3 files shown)


web-ui/src/components/feedback_request_card/feedback_request_subcard/FeedbackRequestSubcard.jsx
```diff
@@ -3,7 +3,7 @@ import { styled } from "@mui/material/styles";
 import PropTypes from "prop-types";
 import Grid from "@mui/material/Grid";
 import Typography from "@mui/material/Typography";
-import {Link} from "react-router-dom";
+import { Link } from "react-router-dom";
```
mvolkmann, line 6: If you don't want Prettier to add those spaces inside curly braces and you don't want strings to be delimited with double-quotes, you can create a `.prettierrc` file at the root of your app containing one or more of these settings:

```json
{
  "arrowParens": "avoid",
  "bracketSpacing": false,
  "htmlWhitespaceSensitivity": "ignore",
  "printWidth": 80,
  "singleQuote": true,
  "trailingComma": "none",
  "useTabs": false
}
```

PR #1639, opened 2022-02-08
Feature 1635/add guild links backend

smithj11:
Backend now supports creation, update, and validation of guild links

### DIFF
--- a/server/src/main/java/com/objectcomputing/checkins/services/guild/Guild.java
+++ b/server/src/main/java/com/objectcomputing/checkins/services/guild/Guild.java
@@ -53,11 +53,11 @@ public class Guild {
 
 
 
-    public Guild(String name, String description, String link) {
+    public Guild(String name, String description, @Nullable String link) {
         this(null, name, description, link);
     }
 
-    public Guild(UUID id, String name, String description, String link) {
+    public Guild(UUID id, String name, String description, @Nullable String link) {
         this.id = id;
         this.name = name;
         this.description = description;
@@ -88,6 +88,7 @@ public class Guild {
         this.description = description;
     }
 
+    @Nullable
     public String getLink() {
         return this.link;
     }
--- a/server/src/main/java/com/objectcomputing/checkins/services/guild/Guild.java
+++ b/server/src/main/java/com/objectcomputing/checkins/services/guild/Guild.java
@@ -39,7 +39,7 @@ public class Guild {
     @Nullable
     @Column(name="link", unique=true)
     @ColumnTransformer(
-        read = "pgp_sym_decrypt(name::bytea,'${aes.key}')",
+        read = "pgp_sym_decrypt(link::bytea,'${aes.key}')",
         write = "pgp_sym_encrypt(?,'${aes.key}') "
     )
     @Schema(description="link to the homepage of the guild")
--- a/server/src/main/java/com/objectcomputing/checkins/services/guild/Guild.java
+++ b/server/src/main/java/com/objectcomputing/checkins/services/guild/Guild.java
@@ -11,8 +11,7 @@ import javax.persistence.Entity;
 import javax.persistence.Id;
 import javax.persistence.Table;
 import javax.validation.constraints.NotBlank;
-
-import edu.umd.cs.findbugs.annotations.Nullable;
+import io.micronaut.core.annotation.Nullable;
 
 import java.util.Objects;
 import java.util.UUID;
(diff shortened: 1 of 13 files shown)


server/src/main/java/com/objectcomputing/checkins/services/guild/Guild.java
```diff
@@ -11,6 +11,9 @@
 import javax.persistence.Id;
 import javax.persistence.Table;
 import javax.validation.constraints.NotBlank;
+
+import edu.umd.cs.findbugs.annotations.Nullable;
+
```
mkimberlin, line 16: I think you wanted this annotation, didn't you? `io.micronaut.core.annotation.Nullable`

server/src/main/java/com/objectcomputing/checkins/services/guild/Guild.java
```diff
@@ -33,6 +36,12 @@
     @Schema(description = "name of the guild")
     private String name;
 
+
+    @Nullable
+    @Column(name="link", unique=true)
+    @Schema(description="link to OCI compass page of guild")
+    private String link;
```
mkimberlin, line 43: You're going to hate me, but I think I would encrypt this field. It can provide hints that indicate persona to someone who gets into the database. It's a small risk, but it's there.

server/src/main/java/com/objectcomputing/checkins/services/guild/GuildServicesImpl.java
```diff
@@ -56,10 +59,25 @@ public void setEmailSender (EmailSender emailSender){
         this.emailSender = emailSender;
     }
 
+    public boolean validateLink (String link ) {
+        String regex = "(https://)?(www.)?compass\\.objectcomputing\\.com/S*";
```
mkimberlin, line 63: We should not be restricting this to an OCI site. This application is intended to be more broadly useful.

server/src/main/java/com/objectcomputing/checkins/services/guild/GuildUpdateDTO.java
```diff
@@ -27,18 +27,23 @@
     @Schema(description = "description of the guild")
     private String description;
 
+    @Nullable
+    @Schema(description="link to OCI compass page of guild")
```
mkimberlin, line 31: "link to the homepage of the guild"

server/src/main/java/com/objectcomputing/checkins/services/guild/Guild.java
smithj11: Fixed import per your suggestion.

server/src/main/java/com/objectcomputing/checkins/services/guild/Guild.java
smithj11: I encrypted the links.

server/src/main/java/com/objectcomputing/checkins/services/guild/GuildServicesImpl.java
smithj11: I changed the URL validation to more broadly parse URL content.

server/src/main/java/com/objectcomputing/checkins/services/guild/GuildUpdateDTO.java
smithj11: Changed per your suggestion.

server/src/main/java/com/objectcomputing/checkins/services/guild/Guild.java
```diff
@@ -39,7 +38,11 @@
 
     @Nullable
     @Column(name="link", unique=true)
-    @Schema(description="link to OCI compass page of guild")
+    @ColumnTransformer(
+        read = "pgp_sym_decrypt(name::bytea,'${aes.key}')",
```
mkimberlin, line 42: should be `link::bytea`

mkimberlin (changes requested):
One small miss!

mkimberlin:
Now that I think about it...there should probably be a test that would have caught that miss too. You might consider adding one.

server/src/main/java/com/objectcomputing/checkins/services/guild/Guild.java
smithj11: Fixed!

smithj11:
> Now that I think about it...there should probably be a test that would have caught that miss too. You might consider adding one.

I have a few tests in Guild that I changed to test the link.

PR #1651, opened 2022-02-16
Fixed bdays/anniversaries not being sorted after changing search options

smithj11:


web-ui/src/pages/BirthdayAnniversaryReportPage.jsx
```diff
@@ -47,6 +47,20 @@ const BirthdayAnniversaryReportPage = () => {
...
 
+  const sortBirthdays = (birthdayData) => {
+    return birthdayData.sort(
+      (a, b) =>
+        Number(
+          b.birthDay.substring(b.birthDay.indexOf("/"), b.birthDay.length)
+        ) -
+        Number(a.birthDay.substring(a.birthDay.indexOf("/"), a.birthDay.length))
+    );
+  };
```
mkimberlin, line 58: I don't think this will work if multiple months are selected, will it?

paulrdurham: I fixed this as it would cause issues sorting dates. I'm pushing the fix now.

paulrdurham (approved):
I believe the changes to the sort method should resolve any issues.

PR #1659, opened 2022-03-21
Feature 1646/add frontend support for different questionanswer formats

paulrdurham:


pieperm (commented):
I think we need to change the test data in `R_Load_testing_data.sql` to include sample data for the new column `input_type`. Also, there are some very large diffs in `yarn.lock` and `package-lock.json`

mkimberlin (approved):

Issue #1545, opened 2021-09-30
BE: Create service to add/remove a permission to/from a role


HootDunk:
There is a join table titled role_permissions which associates roles and permissions.  Services and controller methods need to be created so that the permissions associated with a role can be changed from the client.

POST request with body of :

{
  permissionId
  roleId
}  
should add create an association between a permission and role.

DELETE with body of:

{
  memberId
  roleId
}  

should remove an association between a permission and a role.

Issue #1134, opened 2021-07-06
FE - modify permissions for a role (M)


jahanner:
As an Admin, I should be able to change which permissions are assigned to which role.

AC
- [x] page is only accessible by an admin
- [x] each role will have a list of permissions with a checkbox, indicating if that permission is assigned. 
- [ ] checking a permission will assign/un-assign a permission for the selected role. This will also trigger a REST call
- [x] add a permissions search bar?

![image](https://user-images.githubusercontent.com/28159710/124667267-6824ab80-de6c-11eb-93ca-d9a48f602a99.png)

PR #1674, opened 2022-04-20
Feature 1134/modify permissions for a role

pieperm:
This creates a page for viewing and changing permissions for each role as outlined in issue #1134. Changing the checkboxes will not work until #1545 is complete.

To test, log in as an admin and go to http://localhost:3000/admin/permissions. The page should not be accessible by other roles.

Features:
- Table display with each row for a permission and each column for a role
- Checkboxes for changing whether a role has a certain permission
- Search bar to find permissions
- Mobile-friendly layout
- Snapshot test
- API for retrieving role and permission data from server
- The page is restricted to only admin access


pieperm:
permissions-page

PR #1683, opened 2022-04-28
Feature 1336/cancellation of feedback requests

### LINKED ISSUE #1335: Create cancel button for feedback requests (S)
## Definition of Done
- [ ] On the view feedback page, there is a button for canceling a feedback request (on a sub-card)
- [ ] The cancel button only appears for pending requests (i.e. current date is before the send date)

### LINKED ISSUE #1336: Implement backend for cancellation of feedback requests
## Definition of Done
- [x] New API method for soft deleting a feedback request
- [x] A canceled feedback request should: 
  - Have its `dueDate` set to `null`
  - Have its `status` set to `canceled`
- [x] Integration tests
- [x] A feedback request can only be canceled if the status is 'sent'
- [x] If a request is canceled, the feedback submission page should show a message saying this
- [x] If a request is canceled, the received feedback page should indicate canceled requests (and remove the edit link)
- [x] If a request is canceled, the view feedback page should indicate canceled requests
- [x] Confirmation dialog for canceling a feedback request

pieperm:
Closes #1336 and closes #1335

This replaces the existing functionality of deleting feedback requests with canceling them. This includes both front-end and back-end changes.

### DIFF
--- a/server/src/main/java/com/objectcomputing/checkins/services/feedback_answer/FeedbackAnswerServicesImpl.java
+++ b/server/src/main/java/com/objectcomputing/checkins/services/feedback_answer/FeedbackAnswerServicesImpl.java
@@ -42,6 +42,8 @@ public class FeedbackAnswerServicesImpl implements FeedbackAnswerServices {
         FeedbackRequest relatedFeedbackRequest = getRelatedFeedbackRequest(feedbackAnswer);
         if (!createIsPermitted(relatedFeedbackRequest)) {
             throw new PermissionException("You are not authorized to do this operation");
+        } else if (relatedFeedbackRequest.getStatus().equals("canceled")) {
+            throw new BadArgException("Attempted to save an answer for a canceled feedback request");
         }
         if (feedbackAnswer.getId() != null) {
             return update(feedbackAnswer);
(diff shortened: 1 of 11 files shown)


pieperm:
An alternative approach to this would be to completely delete the requests from the database, rather than soft-deleting them by marking them as "canceled". Much of the validation would remain the same.

Advantage:
- Clears up clutter in database and on pages for requests that will not be used again

Disadvantage:
- Could lead to broken links on requests where the user has already started editing, but has not yet submitted, but the request is then canceled

server/src/main/java/com/objectcomputing/checkins/services/feedback_answer/FeedbackAnswerServicesImpl.java
```diff
@@ -42,6 +42,8 @@ public FeedbackAnswer save(FeedbackAnswer feedbackAnswer) {
         FeedbackRequest relatedFeedbackRequest = getRelatedFeedbackRequest(feedbackAnswer);
         if (!createIsPermitted(relatedFeedbackRequest)) {
             throw new PermissionException("You are not authorized to do this operation");
+        } else if (relatedFeedbackRequest.getStatus().equals("canceled")) {
```
mkimberlin, line 45: There are 2 "l"s in cancelled. 😬

server/src/main/java/com/objectcomputing/checkins/services/feedback_request/FeedbackRequestServicesImpl.java
```diff
@@ -131,7 +131,11 @@ public FeedbackRequest update(FeedbackRequestUpdateDTO feedbackRequestUpdateDTO)
         boolean dueDateUpdateAttempted = !Objects.equals(originalFeedback.getDueDate(), feedbackRequest.getDueDate());
         boolean submitDateUpdateAttempted = !Objects.equals(originalFeedback.getSubmitDate(), feedbackRequest.getSubmitDate());
 
-        if(dueDateUpdateAttempted && !updateDueDateIsPermitted(feedbackRequest)) {
+        if (feedbackRequest.getStatus().equals("canceled") && originalFeedback.getStatus().equals("submitted")) {
```
mkimberlin, line 134: This is going to be all over the place (misspelled cancelled), I fear. I think I would create an enum that contains the acceptable statuses.

mkimberlin (approved):

PR #1714, opened 2022-06-06
Feature 1713/filter email recipients

### LINKED ISSUE #1713: Filter email recipients
Currently, when sending an email, the user cannot search for specific members to select as recipients. 

A search bar should allow the user to filter the shown recipients by name, guild, or team. For teams or guilds, there should be autocomplete that display as chips.

Another option would be two separate search bars: one to filter by a member name, and another to select multiple teams and guilds.

**Additional context**
Related to #1711

pieperm:
Closes #1713
- Adds a filter for selecting email recipients
- Filter options: name, guild, team, title, location

Should checked items in the left list remain visible even if they are not included by the filter?


pieperm:
Update: Checked items remain visible in the list even if they are not included by the filter

Issue #1557, opened 2021-10-14
Bug: Your session has expired. Please refresh the page to start a new session.


mitza-oci:
A user should be able to navigate the site without session expiration.

PR #1717, opened 2022-06-14
Feature 1557/micronaut 3 upgrade

devaughns:
**Solves:** #1557
**Closes:** #1468

Upgrades the entire check-ins application to Micronaut 3.4.3.
Items changed with the upgrade:

- reactive items replaced with project reactor references
- `UserDetails` replaced with `Authentication`
- Reference to `javax` annotations are replaced with `jakarta` annotations

Implements a background refresh token so the user does not have to refresh the page in order to stay logged in.


web-ui/src/api/api.js
```diff
@@ -26,14 +26,23 @@ export const getMyAxios = async () => {
...
+          baseURL: BASE_API_URL,
+          withCredentials: true,
+          timeout: 30000
+        })
+          .then(() => {
+            // retry original request
+            return myAxios(err.config);
+          })
+          .catch(() => {
+            return Promise.reject(err);
```
devaughns, line 44: This is the main logic for getting a refresh token. Axios will intercept the HTTP 401 errors and attempt to retrieve an access token with the refresh token stored in the browser's cookies. Upon completion of the refresh token request, the original request will trigger.

server/src/main/resources/db/common/V90__create_refresh_token_table.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE refresh_token
+(
+    id varchar PRIMARY KEY,
+    refresh_token varchar,
+    user_name varchar,
+    date_created timestamp,
+    revoked boolean
+);
```
devaughns, line 8: Adds a persistable refresh token. Included within this table is the `revoked` field. This is not yet implemented but could be useful in the future to revoke a refresh token if an employee no longer should have access but their browser still have an access token.

server/src/main/resources/application.yml
```diff
@@ -28,6 +28,14 @@ micronaut:
...
     enabled: true
     authentication: cookie
+    token:
+      jwt:
+        enabled: true
+        generator:
+          refresh-token:
+            enabled: true
+            path: /oauth/access_token
+            secret: ${ OAUTH_REFRESH_TOKEN_SECRET:'pleaseChangeThisSecretForANewOne' }
```
devaughns, line 38: Adds an optional secret to be set within the Cloud Run instance.

mkimberlin:
Hey @devaughns any clue on the failing test in this?

devaughns:
There seems to be an issue with using `@ColumnTransformer` on `content` in `Feedback`. It performs as expected with inserts and selects, but it is failing on updates. This seems to be a Micronaut issue. I'll need to hunt down whether it's implemented incorrectly, or if there is a bug in the version of Micronaut we're using.

PR #1721, opened 2022-06-16
Feature 1718/batch many email recipients with BCC

### LINKED ISSUE #1718: Batch many email recipients with BCC
The limit for the number of recipients in a single email is 50. For emails that are sent to more than 50 people, some will not receive the email.

If the number of intended recipients exceeds 50, the system should send out multiple emails, each with up to 50 members, all BCC. These could be organized by sorting emails in alphabetical order.

An alternative to BCC would just be CC.

**Additional context**
https://dev.mailjet.com/email/reference/send-emails/

pieperm:
Closes #1718

This reworks the email sender to allow sending emails to groups of 50+ recipients. For emails with more than 50 recipients, the emails are divided into multiple "batches" of recipients, where each batch contains a maximum of 50 members. All members receive the email as blank carbon copy.


pieperm:
Several tests are currently failing, and initially I believed there was an issue with reworking the `MailJetSender`. However, it appears the problem is actually how MailJet errors are being handled. Before, all errors were logged and ignored, which allowed tests to pass even though they failed to send the email. As part of this PR, an exception is thrown if MailJet fails to send the email. 

As a result, `GuildServices` is failing tests, and upon further investigation, has strange behavior for sending notification emails. When a guild is created or updated, `GuildServices` sends an email with a summary of everyone who was added or removed. Then, in `GuildMemberServices`, another email is sent for every person added or removed in the format of "person joined/left the guild".

PR #1726, opened 2022-06-22
Feature 1666/view past member checkins

zsbrown97:
Users who are no longer development partners, but still have an open check in, will appear within the same Personnel component, but under a divider, without the option to request feedback.

### DIFF
--- a/web-ui/src/components/personnel/Personnel.jsx
+++ b/web-ui/src/components/personnel/Personnel.jsx
@@ -8,6 +8,7 @@ import {
   selectCurrentUserId,
   selectMostRecentCheckin,
   selectCsrfToken,
+  selectProfile,
 } from "../../context/selectors";
 import Card from "@mui/material/Card";
 import CardHeader from "@mui/material/CardHeader";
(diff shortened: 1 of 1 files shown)


web-ui/src/components/personnel/Personnel.jsx
```diff
@@ -102,15 +160,46 @@ const createFeedbackRequestLink = (memberId) => (
...
+          if (!personnelIds.includes(checkins.teamMemberId)) {
+            pastIds.push(checkins.teamMemberId);
+          }
+          return pastIds;
+        }, []);
+      return pastPersonnelIds.map((memberId) => {
+        const person = selectProfile(state, memberId);
+        return formerPersonnelEntries(
+          person,
+          selectMostRecentCheckin(state, memberId),
```
mkimberlin, line 199: I think we actually need a new selector here. This one returns the most recent check in...but we really want the most recent check in that the current user was PDL for.  So, I think I would create one that does something like:

selectMostRecentCheckinWithPDL(state, memberId, pdlId)

mkimberlin:
Something's fishy in Denmark

PR #1754, opened 2022-06-27
Added new user route

BenVuong:
I added a new route for new users who are not logged in. I am currently controlling this with the variable isLoggedIn in App.js. Pending authentication and roles creation: this would help frontend developers to use this variable to create a new route locally.

### DIFF
--- a/web-ui/src/components/routes/NewUserRoutes.jsx
+++ b/web-ui/src/components/routes/NewUserRoutes.jsx
@@ -1,0 +1,19 @@
+import React, { useContext } from "react";
+import { Switch, Route } from "react-router-dom";
+import Header from "../header/Header";
+import HomePage from "../../pages/HomePage";
+import Test from "../../pages/Test";
+
+export default function NewUserRoutes() {
+  return (
+    <Switch>
+      <Route exact path="/">
+        <Header />
+        <HomePage />
+      </Route>
+      <Route path="/onboarding">
+        <Test />
+      </Route>
+    </Switch>
+  );
+}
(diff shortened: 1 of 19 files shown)


S78901 (dismissed):
This looks pretty good Ben. Thanks for formatting these files with Prettier.

MisheelSodgerel (approved):

web-ui/src/pages/BackgroundInformationPage.css
```diff
@@ -0,0 +1,15 @@
...
+  text-transform: unset !important;
+
+  position: absolute;
+  left: 5%;
+  bottom: 5%;
+}
+
+#introductionSurvey {
+  text-transform: unset !important;
+
```
S78901, line 11: Avoid extra line breaks

web-ui/src/pages/AccessCodePage.jsx
```diff
@@ -0,0 +1,37 @@
+import react from "react";
```
S78901, line 1: This should be:
`import React from "react";`

The lowercase is making it fail the gradle build.

web-ui/src/components/routes/NewUserRoutes.jsx
```diff
@@ -0,0 +1,54 @@
+import React, { useContext } from "react";
```
S78901, line 1: Delete the unused 'useContext' so the gradle build does not fail.

`import React from "react";`

li-brandon (approved):

PR #1758, opened 2022-06-29
feature-1666/view past member checkins with a vengeance

zsbrown97:


S78901 (commented):
This looks pretty good to me.

pieperm (commented):
Code looks good, though when I pulled the branch down, I was unable to access anything other than the homepage (and the sidebar was missing)

pieperm:
If I log in as Geetika, "Unreal Ulysses" is displayed twice in the list of former partners. Is this intentional behavior? Also, clicking on former partners only works if Geetika is logged in as an admin, otherwise it says I do not have permission.

check-ins

zsbrown97:
> 

How are you signing in to the program? I never ran into duplicates like that, I'll investigate.

pieperm (approved):

PR #1762, opened 2022-06-29
Added more to page

BenVuong:


onboard/src/pages/CongratulationsPage.jsx
```diff
@@ -2,12 +2,25 @@ import React from "react";
...
+import cakeImg from "./../assets/img/cake.png";
 
 const Congratulations = () => {
   return (
     <div>
       <center>
         <h1>Congratulations!</h1>
+        <h2>
+          <br />
+          Thank you for completing your job application.
```
S78901, line 14: These should be wrapped in a <Typography> import from MUI library, or a simple <p> tag. It could be that it is failing the test because the text is not wrapped.

S78901: Also, don't use <br /> tags. They're pretty outdated. If you want some spacing, I encourage you to experiment with the Typography element api.

https://mui.com/material-ui/api/typography/
https://mui.com/material-ui/react-typography/

S78901 (approved):
Great job here Ben!

MatthewBackes (approved):

PR #1771, opened 2022-07-01
feature-1728/filter-existing-role-members

zsbrown97:
When adding a user to a role, the drop-down list of users will now no longer include users that are already assigned to that role.


pieperm:
I logged in as an admin, navigated to `/admin/roles` and immediately got the following error:

roles-error

PR #1844, opened 2022-07-22
feature-1810/added elements to onboardee table

danie1r:


web-ui/src/pages/OnboardProgressPage.jsx
```diff
@@ -44,15 +51,15 @@ export default function OnboardProgressPage() {
       name: "Daniel Ryu",
       email: "ryud@objectcomputing.com",
       hireType: "Intern",
-      progress: "Incomplete",
+      //   progress: "Incomplete",
```
S78901, line 54: If we don't need these, should they be commented out or just removed?

web-ui/src/components/onboard_progress/ProgressIndicator.jsx
```diff
@@ -0,0 +1,55 @@
+import * as React from "react";
```
S78901, line 1: I believe this is an older way of importing React. Now we can just do `import React from 'react';`

S78901 (approved):
Great work.

BenVuong (approved):

PR #1864, opened 2022-07-28
Feature 1686/standardize argument validation in services

### LINKED ISSUE #1686: Standardize argument validation in services
There currently exists multiple different standards for checking whether arguments are valid in services:
- Use try-catch to catch an exception and then throw another exception of one's choosing
- Use the `ArgumentsValidation` class to throw a `BadArgException` (this is injected into the constructor)
- Define a `validate()` method in each service that mimics the behavior of `ArgumentsValidation`
- Write if/elseif blocks to check for the arguments and throw exceptions accordingly
- Use the `.orElseThrows()` method for calls that return an `Optional`

There should be a consistent way to validate arguments.

One advantage of using `ArgumentsValidation` is that it is immediately clear that it is validating arguments wherever it is used (as opposed to if/else and try/catch). The usage of `.orElseThrows()` may be more useful because it has a return value of the object being validated, which is used in multiple services.

Alternatives could be any of the options in the list above, or a hybrid of two or more methods of validation. There may be another yet unconsidered way to validate arguments that is not listed here.

**Additional context**

`FeedbackServicesImpl.java` (example of try-catch):
```
try {
    memberProfileServices.getById(feedback.getSentBy());
    memberProfileServices.getById(feedback.getSentTo());
} catch (NotFoundException e) {
    throw new BadArgException("Either the sender id or the receiver id is invalid");
}
```

`ActionItemCRUDValidator.java` (example of `ArgumentsValidation`):
```
argumentsValidation.validateArguments(actionItem.getId() != null, "Found unexpected id %s for action item", actionItem.getId());
argumentsValidation.validateArguments(checkInServices.read(checkinId) == null, "CheckIn %s doesn't exist", checkinId);
```

`RoleServicesImpl.java` (example of if-else):
```
if (roleType == null) {
    throw new BadArgException("Invalid role %s", role);
} else if (role.getId() != null) {
    throw new BadArgException("Found unexpected id %s for role", role.getId());
}
```

`CheckinNoteServicesImpl.java` (example of a local `validate` method):
```
validate(createById == null, "Invalid checkin note %s", checkinNote);
validate(id == null || checkinNoteRepository.findById(id).isEmpty(), "Unable to locate checkin note to update with id %s", checkinNote.getId());

private void validate(boolean isError, String message, Object... args) {
    if (isError) {
        throw new BadArgException(message, args);
    }
}
```

`CheckInServicesImpl.java` (example of `.orElseThrows`):
```
CheckIn checkinRecord = checkinRepo.findById(checkinId).orElseThrow(() -> {
    throw new NotFoundException("Checkin %s not found", checkinId);
});
```

pieperm:
This is a proposal to change how validation is done in back-end services, standardizing the many different ways validation was done as outlined in issue #1686.

Adds a new `Validation` class which is designed to mimic the behavior of the `Optional` object, so that all validation is done on the positive cases, and using the `.orElseThrows` syntax to define the type and message of the exception.

Closes #1686


mkimberlin:
This basically just buys us readability, no? I think I'm ok with it.

PR #1916, opened 2022-08-10
feature-1895/Prevent accordions from shrinking and add email preview

MatthewBackes:
Accordions on background information no longer shrink.
Added fields for preferred first and last name
Added preview for what work email will look like


S78901 (approved):
Good work here.

BenVuong (approved):

PR #1946, opened 2022-08-19
feature-1884/update-onboardsurvey-complete-table

danie1r:
Trying to retrieve "stepComplete" variable from BackgroundInformation in check-ins table from onboard app.

### DIFF
--- a/onboard/src/api/postBackgroundComplete.js
+++ b/onboard/src/api/postBackgroundComplete.js
@@ -1,0 +1,7 @@
+
+
+const postJobHistory = (jobHistoryData, accessToken) => {
+    return async (dispatch, getState)=>{
+        
+    }
+}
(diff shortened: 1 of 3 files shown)


onboard/src/api/postBackgroundComplete.js
```diff
@@ -0,0 +1,7 @@
+
+
+const postJobHistory = (jobHistoryData, accessToken) => {
+    return async (dispatch, getState)=>{
+
```
S78901, line 5: It looks like this is an empty method call.

web-ui/src/api/checkins.js
```diff
@@ -3,6 +3,15 @@ import { resolve } from "./api.js";
 const checkinsUrl = "/services/check-ins";
 const checkinsNoteUrl = "/services/checkin-notes";
 const checkinsPrivateNoteUrl = "/services/private-notes";
+const backgroundInformationUrl = "/services/background-information";
+
+export const updateBackgroundInformation = async (cookie) => {
+  return resolve({
+    url:backgroundInformationUrl,
```
S78901, line 10: If you merge the changes from develop in, I believe you will see that there is now a required field, 'emailAddress' which should also be sent through to the backend to retrieve data. However, it looks like the only available way currently in the background information controller to get info is by Id. This will need to change.

mkimberlin:
Needs some work, per Syd.

PR #1998, opened 2022-12-07
Feature 1986/canceled feedback requests

jahanner:
**Merging into Master**
Added `Canceled Requests` collapsable section to the bottom of `Feedback -> Requests Received` page.

Sorry that some of the changes are just Prettier formatting. I didn't see any standard we have for Prettier formatting but please me know if there is something I missed so I can adjust

### DIFF
--- a/web-ui/src/pages/ReceivedRequestsPage.jsx
+++ b/web-ui/src/pages/ReceivedRequestsPage.jsx
@@ -313,21 +313,15 @@ const ReceivedRequestsPage = () => {
         )}
         {!isLoading && (
           <div className="received-requests-container">
-            {receivedRequests.length === 0 && (
-              <div className="no-requests-message">
-                <Typography variant="body1">
-                  No received feedback requests
-                </Typography>
-              </div>
-            )}
-            {receivedRequests.length > 0 &&
-              filteredReceivedRequests.length === 0 && (
-                <div className="no-requests-message">
-                  <Typography variant="body1">
-                    No matching feedback requests
-                  </Typography>
-                </div>
-              )}
+            {receivedRequests.length === 0 ||
+              (receivedRequests.length > 0 &&
+                filteredReceivedRequests.length === 0 && (
+                  <div className="no-requests-message">
+                    <Typography variant="body1">
+                      No received feedback requests
+                    </Typography>
+                  </div>
+                ))}
             {filteredReceivedRequests.map((request) => (
               <ReceivedRequestCard key={request.id} request={request} />
             ))}
--- a/web-ui/src/pages/ReceivedRequestsPage.jsx
+++ b/web-ui/src/pages/ReceivedRequestsPage.jsx
@@ -35,7 +35,8 @@ const Root = styled("div")({
   [`& .${classes.pageTitle}`]: {
     paddingRight: "0.4em",
     minWidth: "330px",
-    ["@media screen and (max-width: 600px)"]: { // eslint-disable-line no-useless-computed-key
+    ["@media screen and (max-width: 600px)"]: {
+      // eslint-disable-line no-useless-computed-key
       fontSize: "30px",
       width: "100%",
       padding: 0,
@@ -48,7 +49,8 @@ const Root = styled("div")({
   },
   [`& .${classes.searchField}`]: {
     width: "100%",
-    ["@media screen and (max-width: 840px)"]: { // eslint-disable-line no-useless-computed-key
+    ["@media screen and (max-width: 840px)"]: {
+      // eslint-disable-line no-useless-computed-key
       marginBottom: "1em",
     },
   },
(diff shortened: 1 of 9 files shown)


web-ui/src/pages/ReceivedRequestsPage.jsx
```diff
@@ -16,77 +20,91 @@ import Divider from "@mui/material/Divider";
...
 };
 
-const Root = styled('div')({
+const Root = styled("div")({
   [`& .${classes.pageTitle}`]: {
     paddingRight: "0.4em",
     minWidth: "330px",
-    ['@media screen and (max-width: 600px)']: { // eslint-disable-line no-useless-computed-key
+    ["@media screen and (max-width: 600px)"]: {
+      // eslint-disable-line no-useless-computed-key
```
mkimberlin, line 39: This is why your build is failing...has to be on the same line. @jahanner

jahanner: Gratzie

web-ui/src/pages/ReceivedRequestsPage.jsx
```diff
@@ -98,55 +114,92 @@ const ReceivedRequestsPage = () => {
...
+          });
+        };
+        setFiltered(filteredCanceled);
+        setFiltered(filteredReceived);
+        setFiltered(filteredSubmitted);
+        // filteredCanceled = filteredCanceled.filter((request) => {
+        //   const creatorName = selectProfile(state, request.creatorId).name.toLowerCase();
+        //   const requesteeName = selectProfile(state, request.requesteeId).name.toLowerCase();
+        //   return creatorName.includes(query) || requesteeName.includes(query);
+        // });
```
mkimberlin, line 177: It looks like the goal of this code was to set `filteredCanceled`, `filteredReceived`, and `filteredSubmitted` to their filtered versions based on the query string. However, it seems that `filteredOption` is only a local parameter within the lambda method. As a result, setting `filteredOption = filteredOption.filter(...)` only modifies it locally. To achieve the desired effect, you could try returning `filteredOption` from the `setFiltered` function and calling it like this instead: `filteredCanceled = setFiltered(filteredCanceled);` This should allow you to modify `filteredCanceled`, `filteredReceived`, and `filteredSubmitted` as desired.

web-ui/src/pages/ReceivedRequestsPage.jsx
```diff
@@ -214,84 +284,167 @@ const ReceivedRequestsPage = () => {
...
+              </div>
+            )}
+            {receivedRequests.length > 0 &&
+              filteredReceivedRequests.length === 0 && (
+                <div className="no-requests-message">
+                  <Typography variant="body1">
+                    No matching feedback requests
+                  </Typography>
+                </div>
+              )}
```
mkimberlin, line 328: This could be collapsed into a single: `receivedRequests.length === 0 || (receivedRequests.length > 0 && filteredReceivedRequests.length === 0) && ...`

web-ui/src/pages/ReceivedRequestsPage.jsx
```diff
@@ -214,84 +284,167 @@ const ReceivedRequestsPage = () => {
...
+          aria-label="show more"
+          className={
+            canceledRequestsExpanded ? classes.expandOpen : classes.expandClose
+          }
+          size="large"
+        >
+          <ExpandMoreIcon />
+        </IconButton>
+      </div>
+      <Collapse in={!canceledRequestsExpanded} timeout="auto" unmountOnExit>
```
mkimberlin, line 405: The other collapsable (for submitted) has a `Divider` above the line corresponding to this. You should probably make sure they match.

mkimberlin (changes requested):
Some minor changes to be made on this.

web-ui/src/pages/ReceivedRequestsPage.jsx
```diff
@@ -311,21 +294,15 @@ const ReceivedRequestsPage = () => {
...
-              filteredReceivedRequests.length === 0 && (
-                <div className="no-requests-message">
-                  <Typography variant="body1">
-                    No matching feedback requests
-                  </Typography>
-                </div>
-              )}
+            {receivedRequests.length === 0 ||
+              (receivedRequests.length > 0 &&
+                filteredReceivedRequests.length === 0 && (
```
mkimberlin, line 299: So close! you need to close the parens after the `filteredRecievedRequests.length === 0` check. Otherwise the "No matching feedback requests" will only show in the case that receivedRequests.length > 0, and you want it to show when it's === 0 as well.

jahanner: Lol I think I may have had that before but prettier auto removes it...

jahanner: Sorry for all the line changes, added a prettier fix for that for any future people working on this. The line you'll care about is 335

mkimberlin (approved):

PR #2035, opened 2023-10-05
Feat/permission handler

romannaglic:
Applying CAN_DELETE_ORGANIZATION_MEMBERS and CAN_CREATE_ORGANIZATION_MEMBERS + updated tests


romannaglic:
Actually, the user is created when it doesn't exist yet, when the GUI uses the GET method to retrieve information about the user. I put the annotation on the method when new user is created on GUI which is different method.
But as far as I can see, the authorization logic with permissions is implemented only with the custom authentication method and not when the user authenticates via Google, so that part still needs to be implemented.

mkimberlin:
@romannaglic apologies for being so slow to reply. I am not getting notifications on this for some reason. Don't hesitate to ping me on chat if there's something here I need to attend to.

> Actually, the user is created when it doesn't exist yet, when the GUI uses the GET method to retrieve information about the user.

Yes, you are correct. I misspoke. 

> I put the annotation on the method when new user is created on GUI which is different method. But as far as I can see, the authorization logic with permissions is implemented only with the custom authentication method and not when the user authenticates via Google, so that part still needs to be implemented.

If that is true, then we have a problem on those feedback services currently. I can try to take a look at that today to confirm.

mkimberlin:
> Actually, the user is created when it doesn't exist yet, when the GUI uses the GET method to retrieve information about the user. I put the annotation on the method when new user is created on GUI which is different method.

I double-checked that the above service didn't turn around and go through the MemberProfileService. It doesn't...goes straight through to the Repository class. So, we are good there. My concern was invalid.

> But as far as I can see, the authorization logic with permissions is implemented only with the custom authentication method and not when the user authenticates via Google, so that part still needs to be implemented.

I tried to identify why you don't think this is implemented in all cases. I don't see anything that is keeping the SecurityRule for it from being injected in all cases. So, it should apply to anything that has the annotation on it, if I'm reading it correctly.  Feel free to correct me on that though, I could very well be missing something.

If I'm not wrong there, then I think we would need to put a database entry in place when we merge/deploy this, no?

mkimberlin:
@romannaglic Did you ever have a chance to look into the need for DB entries along with the deployment?

mkimberlin:
We had to play some branch games to get a working develop branch. So, this PR got taken care of with #2057

PR #2042, opened 2023-11-07
Step 1 - Add ability to retrieve feedback requests for multiple requestees

mkimberlin:
This approach wasn't my first choice, TBH. I ended up here because of the desire to not break existing code and discovering that I couldn't do a `is NULL` check in the `@Query` on either an expanded `List<String>` or with the `STRING_ARRAY` type.  As I type this, I wonder if I couldn't have explicitly cast it and done the check...but that may have had its own quirks. So, I'll go with this unless you have a better idea. ;-)

### DIFF
--- a/web-ui/src/components/reviews/TeamReviews.jsx
+++ b/web-ui/src/components/reviews/TeamReviews.jsx
@@ -261,7 +261,7 @@ const TeamReviews = ({ periodId }) => {
           : null;
       if (data && data.length > 0) {
         data = data.filter((review)=>"canceled".toUpperCase() !== review?.status?.toUpperCase());
-        newSelfReviews[data[0].requesteeId] = data[0];
+        data.forEach(selfReview => newSelfReviews[selfReview.requesteeId] = selfReview);
       }
     };
 
@@ -278,7 +278,12 @@ const TeamReviews = ({ periodId }) => {
           : null;
       if (data && data.length > 0) {
         data = data.filter((review)=>"canceled".toUpperCase() !== review?.status?.toUpperCase());
-        newReviews[data[0].requesteeId] = data;
+        data.forEach(review => {
+            if(!newReviews[review.requesteeId]) {
+                newReviews[review.requesteeId] = [];
+            }
+            newReviews[review.requesteeId].push(review);
+        });
       }
     };
 
@@ -286,10 +291,10 @@ const TeamReviews = ({ periodId }) => {
       loadingReviews.current = true;
       setSelfReviews({});
       setReviews(null);
-      loadingReviews.current = false;
-      loadedReviews.current = true;
       await getSelfReviewRequests(teamMembers);
       await getReviewRequests(teamMembers);
+      loadingReviews.current = false;
+      loadedReviews.current = true;
       setSelfReviews({...newSelfReviews});
       setReviews({...newReviews});
     }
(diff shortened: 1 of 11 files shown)


web-ui/src/components/reviews/TeamReviews.jsx
```diff
@@ -260,12 +261,13 @@ const TeamReviews = ({ periodId }) => {
           : null;
       if (data && data.length > 0) {
         data = data.filter((review)=>"canceled".toUpperCase() !== review?.status?.toUpperCase());
-        newSelfReviews[teamMember.id] = data[0];
+        newSelfReviews[data[0].requesteeId] = data[0];
```
mvolkmann, line 264: Is it okay to assume we just want the first team member?
Is there a possibility that `data` is an empty array?

web-ui/src/components/reviews/TeamReviews.jsx
```diff
@@ -276,23 +278,20 @@ const TeamReviews = ({ periodId }) => {
           : null;
       if (data && data.length > 0) {
         data = data.filter((review)=>"canceled".toUpperCase() !== review?.status?.toUpperCase());
-        newReviews[teamMember.id] = data;
+        newReviews[data[0].requesteeId] = data;
```
mvolkmann, line 281: Same questions as above.

web-ui/src/components/reviews/TeamReviews.jsx
```diff
@@ -276,23 +278,20 @@ const TeamReviews = ({ periodId }) => {
...
-      const promises = teamMembers.map(getSelfReviewRequest);
-      promises.push(...teamMembers.map(getReviewRequest));
-
-      Promise.all(promises).then((res) => {
-        loadingReviews.current = false;
-        loadedReviews.current = true;
-        setSelfReviews({...newSelfReviews});
-        setReviews({...newReviews});
-      });
+      loadingReviews.current = false;
```
mvolkmann, line 289: Should this line be after the two lines with `await` below?

web-ui/src/components/reviews/TeamReviews.jsx
```diff
@@ -276,23 +278,20 @@ const TeamReviews = ({ periodId }) => {
...
+      loadedReviews.current = true;
```
mvolkmann, line 290: Should this follow the two lines with `await` below?

web-ui/src/pages/ViewFeedbackPage.jsx
```diff
@@ -134,20 +136,14 @@ const ViewFeedbackPage = () => {
...
-                feedbackRequests.push(request);
-              }
-            });
-            resolve();
-          });
-        }));
+        const memberRequests = await getFeedbackRequestsForTeamMembers(teamMembers);
+        memberRequests.forEach((request) => {
+          if (!contains(request)) {
+            feedbackRequests.push(request);
```
mvolkmann, line 142: Maybe `feedbackRequests` should be a `Set` so it will automatically exclude duplicates.

web-ui/src/components/reviews/TeamReviews.jsx
mkimberlin: There's an existing check for `data` being empty a couple of lines up. That said...wow...bug miss on my part. I'm not sure where my head was on this one. Fixing now.

web-ui/src/pages/ViewFeedbackPage.jsx
mkimberlin: This is probably exposing how little I write code these days, but is there a way to provide Set with a custom comparator (see line 134/136)? Or is there some reason that wouldn't be necessary that I'm missing?

mvolkmann: I don't think there is a way to do that. I believe it compares primitive values by their values and objects by their address in memory. So I think a Set could contain two objects with identical properties as long as they are distinct objects in memory.

PR #2105, opened 2024-02-08
Upgraded node version to minimum required to work on M1 ARM chipsets

ZacharyKlein:


mkimberlin (approved):

mkimberlin:
This PR looks fine, but your branch name doesn't follow the [supported](https://github.com/objectcomputing/check-ins/blob/develop/.github/workflows/gradle-build-feature.yml) branch [naming conventions](https://github.com/objectcomputing/check-ins/blob/develop/CONTRIBUTING.md#branches). So, your build isn't triggering. If you rename your branch, it will build and we can pull it in.

PR #2117, opened 2024-03-06
Check for report permission before rendering REPORTS menu item

ZacharyKlein:


web-ui/src/context/selectors.js
```diff
@@ -40,6 +40,12 @@ export const selectIsAdmin = createSelector(
     userProfile && userProfile.role && userProfile.role.includes("ADMIN")
 );
 
+export const selectHasReportPermission = createSelector(
+    selectUserProfile,
+    (userProfile) =>
+        userProfile && userProfile.role && userProfile.permissions.any((permission) => permission.includes("REPORT"))
+);
+
```
mkimberlin, line 48: @ZacharyKlein permissions for the current user are already loaded in state?

mkimberlin:
@ZacharyKlein I would probably add a test or two for that new selector.

web-ui/src/context/selectors.js
ZacharyKlein: @mkimberlin I believe so, when I inspect the network response from `/member-profiles/current` contains this block:

```
"permissions": [
        {
            "id": "008f6641-0b0a-4e89-84f0-c580f912b80d",
            "permission": "CAN_VIEW_FEEDBACK_REQUEST"
        },
        {
            "id": "056e32b2-8d07-11ee-b9d1-0242ac120002",
            "permission": "CAN_VIEW_ANNIVERSARY_REPORT"
        },
```

ZacharyKlein:
> @ZacharyKlein I would probably add a test or two for that new selector.

I verified that (with my latest changes) the snapshot test for the admin menu fails without a "_REPORT" permission, and passes when the permission is present. I also added a test for the new selectHasReportPermission selector (and fixed a bug in the process)

web-ui/src/components/menu/Menu.jsx
```diff
@@ -137,7 +137,11 @@ function Menu() {
   const csrf = selectCsrfToken(state);
   const { id, workEmail } =
     userProfile && userProfile.memberProfile ? userProfile.memberProfile : {};
+
+  console.log("userProfile", userProfile);
+
```
mkimberlin, line 142: 😱

mkimberlin (approved):

PR #2122, opened 2024-03-13
Implemented selectors, UI change and tests for permission-based report availablity

ZacharyKlein:
@mkimberlin Not sure if I have the correct permissions for all of these reports, I just made my best guess - please review and advise.


server/src/main/java/com/objectcomputing/checkins/security/permissions/Permissions.java
```diff
@@ -15,6 +15,7 @@ public enum Permissions {
     CAN_VIEW_ANNIVERSARY_REPORT,
     CAN_VIEW_BIRTHDAY_REPORT,
     CAN_VIEW_PROFILE_REPORT,
+    CAN_VIEW_CHECKINS_REPORT,
```
mkimberlin, line 18: One more small thing needed with this. We need to load this permission into the DB and assign it to the ADMIN role in the load test data sql file.

mkimberlin (dismissed):

web-ui/src/context/selectors.test.js
```diff
@@ -1195,5 +1195,99 @@ describe("Selectors", () => {
...
+
+
+        const testState1 = {
+            userProfile: {
+                firstName: "Big",
+                lastName: "Boss",
+                role: "ADMIN",
+                permissions: [{permission: "CAN_VIEW_SKILLS_REPORT"}, {permission: "CAN_VIEW_ANNIVERSARY_REPORT"}],
+            }
+        };
```
pieperm, line 1279: This variable is not being used in this test

web-ui/src/context/selectors.test.js
```diff
@@ -1195,5 +1195,99 @@ describe("Selectors", () => {
...
```
pieperm, line 1256: This variable is not being used in this test

web-ui/src/context/selectors.test.js
```diff
@@ -1195,5 +1195,99 @@ describe("Selectors", () => {
...
+
+
+        const testState1 = {
+            userProfile: {
+                firstName: "Big",
+                lastName: "Boss",
+                role: "ADMIN",
+                permissions: [{permission: "CAN_VIEW_CHECKINS"}, {permission: "CAN_VIEW_ANNIVERSARY_REPORT"}],
+            }
+        };
```
pieperm, line 1233: This variable is not being used in this test

pieperm (commented):
Most of Check-Ins uses two spaces for indents in JavaScript. We could change it to be four spaces if we want, but maybe it should be a separate issue. The diff on this PR makes it harder to tell which changes are part of the feature and which changes are due to the auto-formatting.

mkimberlin (approved):

pieperm (dismissed):
There seems to be something broken with `getReportLinks` in Menu.jsx. Despite having logging in as SUPER and having all the necessary permissions for viewing reports, none of those pages appear in the menu. I can still access the report pages by manually entering the URL. I think the issue may be caused by the selectors for `selectHasBirthdayAnniversaryReportPermission`, `selectHasCheckinsReportPermission`, etc.

A snippet of the list of the user's permissions from the console:

```
[
  ...
  "CAN_VIEW_SKILLS_REPORT",
  "CAN_VIEW_RETENTION_REPORT",
  "CAN_VIEW_ANNIVERSARY_REPORT",
  "CAN_VIEW_BIRTHDAY_REPORT",
  "CAN_VIEW_PROFILE_REPORT",
  ...
]
```

pieperm (approved):

mkimberlin (approved):

PR #2125, opened 2024-03-14
Feature 2074/checkins elevated permissions

Samibanley:


S78901 (dismissed):
This generally speaking looks good to me. Granted, my specialty isn't Java.

ZacharyKlein (approved):

server/src/main/java/com/objectcomputing/checkins/services/checkins/CheckInServicesImpl.java
```diff
@@ -30,50 +35,67 @@ public class CheckInServicesImpl implements CheckInServices {
...
     public Boolean accessGranted(@NotNull UUID checkinId, @NotNull UUID memberId) {
         Boolean grantAccess = false;
 
-        MemberProfile memberTryingToGainAccess = memberRepo.findById(memberId).orElseThrow(() -> {
+        memberRepo.findById(memberId).orElseThrow(() -> {
             throw new NotFoundException(String.format("Member %s not found", memberId));
         });
         CheckIn checkinRecord = checkinRepo.findById(checkinId).orElseThrow(() -> {
             throw new NotFoundException(String.format("Checkin %s not found", checkinId));
         });
```
pieperm, line 71: Instead of being supplied with a function that throws an exception, `orElseThrow` should be given an exception:
```
call.orElseThrow(() -> new NotFoundException("message"));
```

This occurs a few times in this file, not due to your changes, but it seems worth changing. (see also: line 86)

server/src/main/java/com/objectcomputing/checkins/services/checkins/CheckInServicesImpl.java
```diff
@@ -30,50 +35,67 @@ public class CheckInServicesImpl implements CheckInServices {
...
+                               CurrentUserServices currentUserServices,
+                               RoleServices roleServices, PermissionServices permissionServices) {
         this.checkinRepo = checkinRepo;
         this.memberRepo = memberRepo;
         this.currentUserServices = currentUserServices;
         this.roleServices = roleServices;
+        this.permissionServices = permissionServices;
+    }
+
+    public Boolean hasElevatedAccessPermission(@NotNull UUID memberId) {
```
pieperm, line 51: This method should be annotated with `@Override`

server/src/main/java/com/objectcomputing/checkins/services/checkins/CheckInServicesImpl.java
```diff
@@ -30,50 +35,67 @@ public class CheckInServicesImpl implements CheckInServices {
...
+        memberRepo.findById(memberId).orElseThrow(() -> {
             throw new NotFoundException(String.format("Member %s not found", memberId));
         });
         CheckIn checkinRecord = checkinRepo.findById(checkinId).orElseThrow(() -> {
             throw new NotFoundException(String.format("Checkin %s not found", checkinId));
         });
 
+        boolean hasElevatedAccess = hasElevatedAccessPermission(memberId);
+        
+        // the check for Admin as a role should be removed when permissions are fully implemented
```
pieperm, line 75: Maybe make this comment a `TODO`?

server/src/main/java/com/objectcomputing/checkins/services/private_notes/PrivateNoteController.java
```diff
@@ -51,6 +53,7 @@ public PrivateNoteController(PrivateNoteServices privateNoteServices,
      * @return
      */
     @Post("/")
+    @RequiredPermission(Permissions.CAN_CREATE_PRIVATE_NOTE)
```
pieperm, line 56: If a user has the `ELEVATED_ACCESS` permission, they are prohibited from accessing this endpoint unless that user also has the `CAN_CREATE_PRIVATE_NOTE` permission. This goes for all the other endpoints annotated with `@RequiredPermission` as well.

web-ui/build.gradle
```diff
@@ -4,7 +4,7 @@ plugins {
 }
 
 node {
-    version = '14.17.6' // https://nodejs.org/en/
+    version = '16.20.2' // https://nodejs.org/en/
```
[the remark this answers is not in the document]
Samibanley: So I worked on this change with Kimberlin. The reason it is in this PR is because it was actually something that was breaking on my machine. It was causing a failure to build in gradle. I'm not opposed to cherry-picking this and merging it in on it's own though. I'll leave the decision up to @mkimberlin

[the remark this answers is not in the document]
mkimberlin: I'm not too concerned about this, tbh. Yes, ideally I would break it out...but I don't want to make work over it.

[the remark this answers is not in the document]
pieperm: Are we going to update `SETUP_MAC.md` with this?

server/src/main/java/com/objectcomputing/checkins/services/private_notes/PrivateNoteController.java
Samibanley: As per the conversation at stand up, we are going to keep all permissions separate and address permission dependencies at a later time.

borinquenkid (changes requested):
Talked in meeting about cleaning up order of operations in save method

server/src/test/java/com/objectcomputing/checkins/services/role/role_permissions/RolePermissionsControllerTest.java
```diff
@@ -61,7 +60,7 @@ void testGetAllRolePermissions() {
         assertEquals(adminRole.getId(), actual.get(0).getRoleId());
         assertEquals(adminRole.getRole(), actual.get(0).getRole());
         assertNotNull(actual.get(0).getDescription());
-        assertEquals(19, actual.get(0).getPermissions().size());
+        assertEquals(adminPermissions.size(), actual.get(0).getPermissions().size());
```
Samibanley, line 63: I changed this because having a hardcoded number here makes this test very brittle. I don't think it's necessary to have a test failing every time permissions are intentionally added or deleted.

S78901 (approved):

PR #2137, opened 2024-03-21
Filter terminated team members from suggestion results

mkimberlin:


server/src/main/java/com/objectcomputing/checkins/services/feedback/suggestions/FeedbackSuggestionServiceImpl.java
```diff
@@ -80,6 +82,12 @@ public List<FeedbackSuggestionDTO> getSuggestionsByProfileId(UUID id) {
 
             if(suggestions.size() >= maxSuggestions) break;
         }
+
+        suggestions = suggestions.stream().filter((FeedbackSuggestionDTO suggestion) -> {
+            MemberProfile suggested = memberProfileServices.getById(suggestion.getId());
+            LocalDate terminationDate = suggested.getTerminationDate();
+            return !(terminationDate != null && terminationDate.isBefore(LocalDate.now().atStartOfDay().toLocalDate()));
+        }).collect(Collectors.toList());
```
pieperm, line 90: By filtering the list of suggestions at the end, we may end up with fewer than the maximum number of suggestions in certain cases. If the first team members it finds are terminated, or the supervisor or PDL are terminated, then the result would have fewer suggestions, potentially zero. However, there may be additional non-terminated team members who should be suggested in their place.

mkimberlin: Well, that's what I get for not reading the entire service. 🤷

pieperm (approved):

S78901 (approved):
Looks good!

PR #2140, opened 2024-03-26
Upgrade react, react-scripts, storybook, and any other necessary libr…

mkimberlin:
…aries


Samibanley (dismissed):
I'm adding a subsection to the SETUP_MAC.md that will talk about the appropriate version to install and use, so I don't think you need to add it here. It should be a fairly close follow to this. Other than that, I think this all looks good. I'm on my phone right now, so I can't pull the code down and test it out myself.

PR #2141, opened 2024-03-26
Bugfix 2116/email export mishandles commas

### LINKED ISSUE #2116: Email export mishandles commas
**Describe the bug**
When exporting the recipients of an email list, the system mishandles values that contain commas. This is specifically problematic for Titles.

**To Reproduce**
1. Go to 'ADMIN'
2. Click on 'Send Email'
3. Select 'TEXT' format
4. Click 'NEXT'
5. Enter any text into the subject and email contents fields
6. Click 'NEXT'
7. Include someone with a comma in their Title in the Recipients list
8. Hit the export/download button above the Recipients list
9. Notice that the CSV produced does not correctly quote values that contain commas

**Expected behavior**
The produced file is properly handles commas in any values

**Additional context**
It should be noted that this appears to be a client-side behavior. The "Download Members" capability provided to admins via the avatar menu in the upper right (which makes a call to the server) works correctly. It might be worthwhile to unify the approach to export these two use. Perhaps a POST endpoint could be added to that CSV export that takes a list of user ids or some such?

pieperm:
Re-implemented member export service. Removed react-csv dependency from web-ui.

To test:
- Log in as admin
- Go to the Send Email page and export members (try ones with commas in titles or locations)
- Go to the profile icon in the top right and select "Download Members"
- In both of the above cases, commas should be handled correctly by wrapping those entries in quotes

Closes #2116


server/src/test/java/com/objectcomputing/checkins/services/TestContainersSuite.java
```diff
@@ -36,6 +37,9 @@ public abstract class TestContainersSuite implements RepositoryFixture, TestProp
     @Inject
     private Flyway flyway;
 
+    @Value("${aes.key}")
+    protected String key;
+
```
mkimberlin, line 42: I'm confused by this...

mkimberlin (approved):
Approving...because I think I answered my own confusion. However, I am curious about why you needed to do both mock tests and e2e tests.

server/src/test/java/com/objectcomputing/checkins/services/TestContainersSuite.java
borinquenkid: This is runtime injection of the aeskey which is required for the custom query

PR #2150, opened 2024-03-27
Chore 2131/change primitive entity columns to use wrappers

### LINKED ISSUE #2131: Change primitive entity columns to use wrappers
**Describe the bug**
In preparation for changing how Jackson serializes Java objects project-wide, we want to ensure that all entity fields can be null. This avoids cases where null values in the database would result in default primitive values in Java.

**Expected behavior**
Any classes that are serialized to JSON with a primitive-type field should use its corresponding wrapper class. These will be entities annotated with `@Entity` and DTOs annotated with `@Introspected`

### LINKED ISSUE #2132: Include empty fields in JSON responses
By default, Micronaut does not include null fields when serializing a Java object to JSON. When using any of our APIs, we would like to ensure that all fields are present in the JSON response, even if those fields are null.

This change would affect all services. This can be done by changing the Jackson configuration.

Alternatively, each entity and response DTO could be annotated with `@JsonInclude` to achieve this behavior. However, it could result in inconsistently defining how objects are serialized. Since this is behavior we want for all services, it makes more sense to configure that at the application level.

**Additional context**
Depends on #2131

pieperm:
Closes #2131 
Closes #2132


S78901 (dismissed):

server/src/main/resources/application.yml
```diff
@@ -135,3 +135,6 @@ mail-jet:
 ---
 github-credentials:
   github_token: ${ GITHUB_TOKEN }
+---
+jackson:
+  serialization-inclusion: ALWAYS
```
mkimberlin, line 140: I'm confused. Isn't this for #2132?

server/src/main/java/com/objectcomputing/checkins/services/employee_hours/EmployeeHours.java
```diff
@@ -55,29 +57,28 @@ public class EmployeeHours {
...
+    private Float targetHours;
 
     @Column(name="asofdate")
     @Schema(description = "as of Date")
     private LocalDate asOfDate;
 
-
-
-    public EmployeeHours(UUID id, @NotNull String employeeId, @NotNull float contributionHours, float billableHours, float ptoHours,LocalDate updatedDate,float targetHours, LocalDate asOfDate) {
+    public EmployeeHours(UUID id, String employeeId, Float contributionHours, Float billableHours, Float ptoHours, LocalDate updatedDate, Float targetHours, LocalDate asOfDate) {
```
mkimberlin, line 66: Why are the `@NonNull`s matching the below constructor not on this one? Looks like these were intentionally taken off?

server/src/main/resources/application.yml
borinquenkid: we decided to put them together since it is a one line change...probably should have made one story for both of them

server/src/main/java/com/objectcomputing/checkins/services/employee_hours/EmployeeHours.java
pieperm: Since I changed the datatypes of the floats, I just regenerated the whole constructor. I didn't realize it omitted the `@NotNull`s, thanks.

pieperm:
@borinquenkid and I have walked through the front-end of the app and did not find any issues due to the Jackson serialization change. We covered the majority of pages, but if anyone thinks of any obscure pages/flows that we may have overlooked, please let us know.

mkimberlin:
So, this looks good to me, except there seem to be some other inconsistencies with the @NotNull annotations between constructors. Otherwise, I'm ready to bring this in.

mkimberlin (approved):

weiqigao (approved):

PR #2156, opened 2024-03-28
Converted to Vite and upgraded gradle, frontend tests are disabled fr…

mkimberlin:
…om gradle check at the moment.


web-ui/src/api/api.js
```diff
@@ -1,4 +1,4 @@
-import axios from "axios";
+// import fetch from "node-fetch";
```
S78901, line 1: Why is this commented out if 'fetch' is used below?

web-ui/src/api/api.js
```diff
@@ -10,14 +10,19 @@ export const getAvatarURL = (email) =>
...
-  if (!myAxios) {
-    myAxios = axios.create({
-      baseURL: BASE_API_URL,
-      withCredentials: true,
-    });
+let myFetch = null;
+
+export const getMyFetch = async () => {
+  if (!myFetch) {
+    myFetch = fetchAbsolute(fetch)(BASE_API_URL);
```
S78901, line 21: Is there a synonymous param to 'withCredentials' for fetch to indicate usage of cookies?

web-ui/src/api/api.js
mkimberlin: It actually should be removed. We shouldn't be using node-fetch here, but rather the browser provided global`fetch`. I'll remove this commented line. Thanks!

web-ui/src/api/api.js
mkimberlin: Yes, `credentials: 'include'`, and it's just a bit above on line 14 and is built into the fetch wrapper returned by `fetchAbsolute`.

web-ui/src/api/api.test.js
```diff
@@ -1,14 +1,17 @@
 import { resolve, BASE_API_URL } from "./api.js";
```
pieperm, line 1: `BASE_API_URL` is not being used

web-ui/src/api/memberskill.js
```diff
@@ -25,11 +24,8 @@ export const getSelectedMemberSkills = async (id, cookie) => {
 
 export const getSkillMembers = async (id, cookie) => {
   return resolve({
-    url: memberSkillUrl,
+    url: `${memberSkillUrl}?skillid=${encodeURIComponent(id)}`,
     responseType: "json",
-    params: {
-      skillid: id,
-    },
```
pieperm, line 32: I'm curious why only this API function was changed to not use the `params` property. There are other API functions that still use `params`.

pieperm (approved):
There are still some leftover Jest snapshots (e.g. GuidesPanel.test.js.snap). Looks like the pattern is that it's all snapshots that end with .test.js.snap

web-ui/src/context/reducer.js
```diff
@@ -33,26 +33,26 @@ import {
...
+  skills: null,
+  teams: null,
+  guilds: null,
   toast: {
     severity: "",
     toast: "",
   },
-  userProfile: {},
+  userProfile: null,
   reviewPeriods: [],
```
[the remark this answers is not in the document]
mkimberlin: I was just trying to keep this changeset to those things that I had to change to get things working. I may well come back and adjust all of these state items. The ones changed here are the ones that were being loaded by AppContext on launch. I had to make these changes as a part of fixing the test issues generated by some bogus initialValue handling.

web-ui/tsconfig.json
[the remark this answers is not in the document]
mkimberlin: Vite/Vitest. Though we do have some typescript in the project, just not much. I think we purged most of it. ;-)

web-ui/src/api/memberskill.js
mkimberlin: This was a huge reminder, thank you! Params isn't a thing with fetch.

web-ui/src/components/private-note/__snapshots__/PrivateNote.test.jsx.snap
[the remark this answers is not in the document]
mkimberlin: It was even worse before:

```javascript
// Jest Snapshot v1, https://goo.gl/fbAQLP

exports[`renders correctly 1`] = `null`;
```

We need an epic to improve tests, to be frank

S78901 (approved):

pieperm (approved):

PR #2160, opened 2024-03-29
Temp files now created that will be deleted after being served.

### LINKED ISSUE #2155: Remove generated CSV files from the server
**Describe the bug**
When a CSV endpoint is accessed, a .csv file is generated on the server (member_profiles.csv and skill_records.csv). This happens when running controller tests that test those endpoints as well.

**To Reproduce**
Run `MemberProfileReportControllerTest` or `SkillsReportControllerTest`, which will generate a .csv file in check-ins/server

**Expected behavior**
These files should be temporary files that are returned from the endpoint and then discarded.

borinquenkid:
closes #2155


borinquenkid (commented):
Test fails incongruously if not run in order.

pieperm (approved):

Samibanley (approved):

mkimberlin (commented):
I don't think I understand why two separate temp file providers are required. It seems the only difference is the resulting filename. So why not pass that into the provider method and just have a generic temp file provider?

Issue #2136, opened 2024-03-20
Migrate From Permission Entity towards a Static Enumeration


ZacharyKlein:
Currently Permissions are represented by a Java Enum class, which provides a static list of all supported permissions, as well as a JPA Entity which combines a permission value with a description and a UUID. This entity is related to roles through the RolePermission entity. After reviewing the code with @mkimberlin, it appears that this entity relationship is not serving any particular feature and introduces additional complexity into the application.  Creating permissions dynamically and storing them in the database is not a use case for Checkins, and the current implementation requires changing both the code and updating the database in concert to role out any new permissions.

This issue will migrate away from the Permission entity class while maintaining the same basic API for retrieval/lookup of supported Permissions and user/role-assigned permissions. The Permissions Enum will be enhanced to include a human-readable `description` and `category` for each permission, and an API will be provided to make this set available as  aJSON payload.

Alternative options considered: N/A

The affected portions of the code include:

- [ ] The `/services/permissions` endpoint, which will require changes to source permissions from the Enum and not from the database. 
- [ ] The `findUserPermissions` method on `PermissionsService` will require changes to source the user permissions from the `RolePermission` entity (this already takes place but will be simplified by this change)
- [ ] The Repository and associated tests and endpoints for the `Permission` entity will be removed

mkimberlin:
I made some baby steps toward this with PR #2135

PR #2161, opened 2024-04-01
Feature 2136 / Migrate permission entity

ZacharyKlein:
This PR migrates away from the Permission entity class and replaces it with consistent using of the Permissions enum. The goal is to simplify the persistence around permissions/role permissions. 

More details: https://github.com/objectcomputing/check-ins/issues/2136

Many updates were made to test fixtures to get the test suite working. Functionality should not be greatly changed except for the loss of the permission id (UUID).


ZacharyKlein:
@mkimberlin Migration has been updated to copy over role_permission records to the new table structure

mkimberlin (approved):
I feel like this covers the server side pretty well...now what will break on the frontend? Just permission management screen maybe? I think it keys off of the ids

server/src/test/java/com/objectcomputing/checkins/services/fixture/PermissionFixture.java
```diff
@@ -10,83 +10,75 @@
...
 
     default void setPermissionsForAdmin(UUID roleID) {
-        List<Permission> permissions = getPermissionRepository().findAll();
-        for(Permissions adminPermission : adminPermissions) {
-            Optional<Permission> permission = permissions.stream().filter(s -> s.getPermission().equals(adminPermission.name())).findFirst();
-            permission.ifPresent(value -> setRolePermission(roleID, value.getId()));
+        List<Permission> permissions = List.of(Permission.values());
+        for(Permission adminPermission : adminPermissions) {
+            Optional<Permission> permission = permissions.stream().filter(s -> s.name().equals(adminPermission.name())).findFirst();
+            permission.ifPresent(value -> setRolePermission(roleID, value));
```
pieperm, line 65: Iterating over all the permissions no longer seems necessary now that the implementation is using only an enum. The check on line 64 will always return a permission since the list of `adminPermissions` is guaranteed to only contain a subset of the enum. (This applies to `setPermissionsForPdl` and `setPermissionsForMember` as well)

server/src/test/java/com/objectcomputing/checkins/services/role/role_permissions/RolePermissionsControllerTest.java
```diff
@@ -85,33 +84,33 @@ void testAssignPermissionToRole() {
         MemberProfile sender = createADefaultMemberProfile();
         assignAdminRole(sender);
 
-        Role memberRole = getRoleRepository().findByRole(RoleType.MEMBER.name()).orElseThrow();
-        Permission birthdayPermission = getPermissionRepository().findByPermission(Permissions.CAN_VIEW_BIRTHDAY_REPORT.name()).get(0);
+        Role memberRole = getRoleRepository().findByRole(RoleType.MEMBER.name()).get();
```
pieperm, line 87: I noticed that the `testGetAllRolePermissions` method still uses `.orElseThrow()` while `testAssignPermissionToRole` and `testRemovePermissionFromRole` were changed to use `.get()`. To me it makes sense to use `orElseThrow()` since these tests should fail if they can't get past that step.

server/src/main/java/com/objectcomputing/checkins/services/role/member_roles/MemberRoleServicesImpl.java
```diff
@@ -39,6 +39,10 @@ public Optional<MemberRole> findById(MemberRoleId memberRoleId) {
         return memberRoleRepository.findById(memberRoleId);
     }
 
+    public Optional<MemberRole> findByMemberId(UUID memberId) {
+        return memberRoleRepository.findByMemberId(memberId);
+    }
+
```
pieperm, line 45: I'm not entirely sure why this was added, as it's not being used anywhere. I'm guessing it may have been an alternative approach to calling `roleServices.findUserRoles(id)` in `RolePermissionServicesImpl::findUserPermissions`

server/src/main/java/com/objectcomputing/checkins/services/role/role_permissions/RolePermissionServicesImpl.java
```diff
@@ -53,14 +53,33 @@ public List<RolePermissionsResponseDTO> findAll() {
...
+
+        Set<Role> memberRoles = roleServices.findUserRoles(id);
+
+        return memberRoles.stream().map(role ->
+                findByRoleId(role.getId())
+                    .stream()
+                    .map(RolePermission::getPermission)
+                    .collect(Collectors.toList()))
+            .flatMap(Collection::stream)
+            .collect(Collectors.toList());
```
pieperm, line 83: Right now this method doesn't have any tests associated with it. It may be worth adding a test or two in `CurrentUserControllerTest` to assign a user a role or multiple roles, and ensure the response has the correct permissions.

pieperm (commented):
There are dangling semicolons in the import sections of many files, mostly in the controllers it seems.

pieperm (changes requested):
The migrations are failing when running the gradle build. This is due to `R__Load_testing_data.sql` line 15, which deletes the `permissions` table, which no longer exists due to migration `V96__drop_create_role_permissions.sql`

pieperm:
I also found two pieces of the front-end that break due to selectors:
- The permissions page always displays "You do not have permission to view this page."
  - Caused by `selectHasPermissionAssignmentPermission`
- The "Reports" section does not appear in the Menu even when the user should have permission
  - Caused by `selectHasReportPermission`

server/src/test/java/com/objectcomputing/checkins/services/memberprofile/currentuser/CurrentUserControllerTest.java
```diff
@@ -37,20 +42,15 @@ public class CurrentUserControllerTest {
...
+    @Inject
+    EmbeddedServer embeddedServer;
+
     @BeforeAll
     void setup() {
-        userAttributes.put("name", firstName + ' ' + lastName);
-        userAttributes.put("email", userEmail);
-        userAttributes.put("picture", imageUrl);
 
         MockitoAnnotations.initMocks(this);
```
pieperm, line 55: Since `initMocks` is deprecated, could we change this to `openMocks`?

server/src/test/java/com/objectcomputing/checkins/services/memberprofile/currentuser/CurrentUserControllerTest.java
```diff
@@ -16,17 +21,17 @@
 import org.mockito.MockitoAnnotations;
```
pieperm, line 21: A few unused imports here: `PermissionServices`, `RoleServices`, and `InjectMocks`

server/src/test/java/com/objectcomputing/checkins/services/memberprofile/currentuser/CurrentUserControllerTest.java
```diff
@@ -16,17 +21,17 @@
...
 
 import javax.validation.constraints.NotNull;
-import java.util.HashMap;
-import java.util.Map;
+import java.util.*;
 
 import static com.objectcomputing.checkins.services.memberprofile.MemberProfileTestUtil.mkMemberProfile;
+import static com.objectcomputing.checkins.services.permissions.Permission.*;
```
pieperm, line 27: Unused import

web-ui/src/context/selectors.js
```diff
@@ -16,68 +16,70 @@ export const selectLoading = (state) => state.loading;
...
-        userProfile && userProfile.role && userProfile.permissions.some((p) => p?.permission?.includes("CAN_ASSIGN_ROLE_PERMISSIONS"))
+        userProfile && userProfile.role && userProfile.permissions.some((p) => p?.includes("CAN_ASSIGN_ROLE_PERMISSIONS"))
 );
 
 export const selectHasReportPermission = createSelector(
     selectUserProfile,
-    (userProfile) =>
-        userProfile && userProfile.role && userProfile.permissions.some((p) => p?.permission?.includes("REPORT"))
+    (userProfile) => {
+        console.log(userProfile.permissions);
```
pieperm, line 52: Leftover `console.log`

server/src/test/java/com/objectcomputing/checkins/services/memberprofile/currentuser/CurrentUserControllerTest.java
ZacharyKlein: I'd rather make that change in a separate PR, since we use `initMocks` across many tests and this instance wasn't part of the PR changes.

pieperm (approved):

mkimberlin (approved):

PR #2162, opened 2024-04-01
change birth year to 1900 if user is not admin.

borinquenkid:
Verification: 
use developer tools on browser to verify on profiles page as a regular user the birth year shows as 1900

### DIFF
--- a/server/src/main/java/com/objectcomputing/checkins/services/memberprofile/MemberProfile.java
+++ b/server/src/main/java/com/objectcomputing/checkins/services/memberprofile/MemberProfile.java
@@ -7,10 +7,8 @@ import io.micronaut.data.model.DataType;
 import io.swagger.v3.oas.annotations.media.Schema;
 import io.micronaut.core.annotation.Introspected;
 import io.micronaut.core.annotation.Nullable;
-import javax.persistence.Column;
-import javax.persistence.Entity;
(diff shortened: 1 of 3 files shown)


server/src/main/java/com/objectcomputing/checkins/services/memberprofile/MemberProfile.java
```diff
@@ -343,6 +341,13 @@ public void setExcluded(@Nullable Boolean excluded) {
         this.excluded = excluded;
     }
 
+    @Transient
```
mkimberlin, line 344: Does this need to be marked as `@Transient`? I wouldn't think so since it's not a field, getter, or setter...

server/src/main/java/com/objectcomputing/checkins/services/memberprofile/MemberProfileServices.java
```diff
@@ -20,4 +20,5 @@ Set<MemberProfile> findByValues(String firstName, String lastName, String title,
     List<MemberProfile> findAll();
 
     List<MemberProfile> getSupervisorsForId(UUID id);
+
```
mkimberlin, line 23: Controversial. 🤣

mkimberlin (approved):
I concur with @pieperm...might be worth adding a controller test or two.

PR #2170, opened 2024-04-03
Feature 2111/convert team skills report to use transfer list

### LINKED ISSUE #2111: Convert Team Skills report to use a transfer list
As a user, I want to use a transfer list component to filter and select team members for the "Team Skills" report, so that I can easily identify the skills of various groups within the company.

Acceptance criteria:
 - [ ] The "Team Skills" report page is updated to display a transfer list component instead of the existing team selection options.
The transfer list component allows filtering by the following options:
   - [ ] Team membership (through a multi-select dropdown)
   - [ ] Guild membership (through a multi-select dropdown)
   - [ ] Manager (through a multi-select dropdown)
   - [ ] Skills possessed (through a multi-select dropdown)
   - [ ] Title (through a multi-select dropdown)
   - [ ] Name
   - [ ] Location (through a multi-select dropdown)
 - [ ] Filtering by manager includes two options:
   - [ ] Show all employees reporting to the selected manager (including their direct reports and indirect reports)
   - [ ] Show only the selected manager's direct reports
 - [ ] When filtering by Manager, the user should have the option to only include direct reports
   - [ ] The option to only include direct reports should not be selected by default

pieperm:
Closes #2111 

This feature introduces a new component, `MemberSelector`, which is a way to filter members by several criteria and select them. For this story, the component is being used for selecting members to use in a team skill report.

To test:
- Log in as an admin
- In the menu, go to REPORTS > Team Skills
- Add members via the member selector
  - Search by member name
  - Filter by guild, team, title, location, skill, or manager
  - Filtering by manager has the additional option of choosing only direct reports or all reports
- After selecting members, select skills for the report and click "Run Search"
- All functionality for the report results should be unchanged


web-ui/src/components/member_selector/member_selector_dialog/__snapshots__/MemberSelectorDialog.test.jsx.snap
```diff
@@ -0,0 +1,7 @@
+// Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
+
+exports[`MemberSelectorDialog > renders correctly 1`] = `
+<div
+  aria-hidden="true"
+/>
```
mkimberlin, line 6: I think we can get this to render with an initialState. Take a look at [this test](https://github.com/objectcomputing/check-ins/blob/5035e949fa13fc491ad446e60f83140dea201a90/web-ui/src/components/edit_skills/EditSkillsCard.test.jsx)

pieperm: I managed to get the test to render the component in the snapshot, but it also outputs a warning:
```
Warning: An update to MemberSelectorDialog inside a test was not wrapped in act(...).
When testing, code that causes React state updates should be wrapped into act(...)
```

I tried a few approaches but couldn't figure out how to get rid of the warning.

mkimberlin (approved):

ZacharyKlein (approved):

PR #2176, opened 2024-04-04
Bugfix 2165/member export report not working

### LINKED ISSUE #2165: The member export report is not working
**Describe the bug**
The CSV download of all members is broken and gives a CSV with just `null` in it.

**To Reproduce**
1. Click on your avatar on the top right of the app
2. Click on 'Download Members'
3. Note that the downloaded file does not contain the members

**Expected behavior**
The exported CSV should contain all of the non-terminated members.

**Screenshots**
Screenshot 2024-04-02 at 8 17 06 PM

**Additional context**
I believe this was introduced when the email list export was changed to the backend by @pieperm, but it could have been me with the Vite conversion.

pieperm:
Closes #2165

After testing out the front-end APIs, I determined that the problem originated from transitioning from axios to fetch. In axios, specifying a `responseType` of `blob` would convert the response payload to a Blob object, which is then passed to the file downloader. With fetch, this conversion has to be done manually, but the implementation was only handling JSON and text responses. This modifies the `resolve` method to return a blob in the payload if the "Content-Type" header is "text/csv".

An alternative fix could be to include `responseType` in the request payload and determine the datatype for the response payload based on that, similar to axios.

### DIFF
--- a/web-ui/src/api/api.js
+++ b/web-ui/src/api/api.js
@@ -88,7 +88,9 @@ export const resolve = async (payload) => {
   } else {
     const contentType = resolved.payload.headers.get("Content-Type");
     const contentLength = resolved.payload.headers.get("Content-Length");
-    if(contentLength && contentLength > 0) {
+    if (contentType && contentType.indexOf("text/csv") !== -1) {
+      resolved.payload.data = await resolved.payload.blob();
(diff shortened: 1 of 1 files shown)


S78901 (approved):
Good catch.

PR #2186, opened 2024-04-10
issue#2173 fix query for selecting emails on the send date.

borinquenkid:


server/src/main/resources/db/dev/R__Load_testing_data.sql
```diff
@@ -1090,6 +1090,10 @@ INSERT INTO feedback_requests
 VALUES
 ('ab2da7fc-fac2-11eb-9a03-0242ac130003', '59b790d2-fabc-11eb-9a03-0242ac130003', '2dee821c-de32-4d9c-9ecb-f73e5903d17a', '67dc3a3b-5bfa-4759-997a-fb6bac98dcf3' ,'18ef2032-c264-411e-a8e1-ddda9a714bae', '2021-08-01', '2021-08-05', null, 'pending');
 
+INSERT INTO feedback_requests
+(id, creator_id, requestee_id, recipient_id, template_id, send_date, due_date, submit_date, status) -- requestee: Zach Brown, recipient: Michael Kimberlin
+VALUES
+('ab2da7fc-fac2-11eb-9a03-0242ac130004', '59b790d2-fabc-11eb-9a03-0242ac130003', '43ee8e79-b33d-44cd-b23c-e183894ebfef', '6207b3fd-042d-49aa-9e28-dcc04f537c2d' ,'18ef2032-c264-411e-a8e1-ddda9a714bae', '2024-04-10', '2024-04-11', null, 'pending');
```
pieperm, line 1096: Would it make sense to change this insertion so that the send date is always the current day? That way we always have a piece of test data that could be used to test the cron job.

mkimberlin: something like `CAST(GETDATE() As DATE)` for send_date...the other two dates can be null

server/src/main/java/com/objectcomputing/checkins/services/request_notifications/CheckServicesImpl.java
```diff
@@ -22,7 +26,8 @@ public CheckServicesImpl(FeedbackRequestServicesImpl feedbackRequestServices,
     @Override
     public boolean sendScheduledEmails() {
         LocalDate today = LocalDate.now();
-        List<FeedbackRequest> feedbackRequests = feedbackRequestRepository.findBySendDateBeforeAndStatusEqual(today, "pending");
+        List<FeedbackRequest> feedbackRequests = feedbackRequestRepository.findBySendDateNotAfterAndStatusEqual(today, "pending");
+        LOG.info("About to send emails: " + feedbackRequests.size());
```
mkimberlin, line 30: 👍

server/src/main/resources/db/dev/R__Load_testing_data.sql
borinquenkid: There is also (NOW())::date

borinquenkid: Won't this mean that an email will be sent from dev the day after we deploy to dev?

mkimberlin: yep...that's exactly what it would mean. and that's exactly what I want. :)

mkimberlin: Further, it means that we could force run right after deployment and use it as a smoke test.

mkimberlin: > There is also (NOW())::date

Mein SQL-foo ist scheiẞe.

borinquenkid: okey dokey

mkimberlin (approved):

PR #2192, opened 2024-04-12
Bugfix 2191/fix permission serialization

mkimberlin:


borinquenkid (approved):
Only question is why Permission and PermissionDTO

S78901 (approved):

mkimberlin:
> Only question is why Permission and PermissionDTO

Because that is the fastest path to getting the APIs to serialize the way that is currently needed. Permissions need to be serialized one way in one case and another in another case. I think ultimately it will be cleaner to not use the Permission object in the APIs directly at all, but I'm not going to take that on before the demo. 😉

PR #2199, opened 2024-04-19
Feature 2194/Cacheable Role Permissions Lookup

ZacharyKlein:
- Lookup permissions by role
- Cache permissions per role
- Invalidate permissions cache on change

### DIFF
--- a/server/src/main/java/com/objectcomputing/checkins/services/role/role_permissions/RolePermissionServicesImpl.java
+++ b/server/src/main/java/com/objectcomputing/checkins/services/role/role_permissions/RolePermissionServicesImpl.java
@@ -6,6 +6,9 @@ import com.objectcomputing.checkins.services.permissions.PermissionServices;
 import com.objectcomputing.checkins.services.role.Role;
 import com.objectcomputing.checkins.services.role.RoleServices;
 
+import io.micronaut.cache.annotation.CacheConfig;
+import io.micronaut.cache.annotation.CacheInvalidate;
+import io.micronaut.cache.annotation.Cacheable;
 import jakarta.inject.Singleton;
 
 import javax.validation.constraints.NotBlank;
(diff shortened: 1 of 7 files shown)


server/src/main/java/com/objectcomputing/checkins/services/role/role_permissions/RolePermissionServicesImpl.java
```diff
@@ -70,6 +76,13 @@ public List<RolePermission> findByRoleId(UUID roleId) {
...
 
+
+    @Cacheable
+    @Override
+    public List<RolePermission> findByRole(String role) {
+        return rolePermissionRepository.findByRole(role);
+    }
+
     @Override
     public List<Permission> findUserPermissions(@NotBlank UUID id) {
```
mkimberlin, line 87: Any particular reason not to make this method `@Cacheable` as well?

server/src/main/resources/application.yml
```diff
@@ -14,6 +14,8 @@ micronaut:
       expire-after-write: 300
       heap:
         max-entries: 600
+    role-permission-cache:
+      expire-after-write: 86400
```
mkimberlin, line 18: This is pretty long...but, I guess why not? The service invalidates if we change things, so I guess I'm down. 🤷

ZacharyKlein: I think I just used the same value as the member cache.

server/src/main/java/com/objectcomputing/checkins/services/role/role_permissions/RolePermissionServicesImpl.java
ZacharyKlein: Probably not! It's not the one being used in the annotation lookup, but it is used during login.

mkimberlin (approved):

mjperry91 (approved):

PR #2205, opened 2024-04-22
docs: update macos environment setup

vhscom:
First pass at the docs for macOS. @mjperry91 and I will work to consolidate as needed.

PR #2206, opened 2024-04-22
Adding WindowsOS page with onboarding steps

mjperry91:


vhscom (approved):
There'll be some overlap with #2205 here. Given both are drafts I don't see a strong need to consolidate them just yet until we hear more feedback on setup. I'll loop back later to get these docs in alignment, and perhaps consolidated more.

PR #2216, opened 2024-04-23
potential fix for back button issue

mvolkmann:


web-ui/index.html
```diff
@@ -14,7 +14,7 @@
       manifest.json provides metadata used when your web app is installed on a
       user's mobile device or desktop. See https://developers.google.com/web/fundamentals/web-app-manifest/
     -->
-    <link rel="manifest" href="manifest.json" />
+    <link rel="manifest" href="manifest.json" crossorigin="use-credentials" />
```
mvolkmann, line 17: This seems to fix the issue, but I think we won't know for sure until we try this in the deployed environment.

mkimberlin:
This seems benign enough that I'm going to role the dice and just bring it down.

Issue #2211, opened 2024-04-23
Role-Based Filtering for MemberSelector Component


mkimberlin:
**As a** an administrator configuring reports and functionality
**I want** the `MemberSelector` component to be adaptable to limit selections based on a specified role (e.g., PDL)
**So that** I can create reports and features that target specific user groups within the platform.

 ## Acceptance Criteria:

* The `MemberSelector` component will be extended to accept a configuration parameter to specify a specific role (e.g., using a role name).
* When a role is specified, the member selection process will be restricted to only those users assigned that specific role. 
* The user interface of the component should clearly indicate that a role filter is applied (e.g., by stating that only members of role XXXX can be selected).
* Other search and selection capability should continue to behave as it currently does, with the only change in results being the restriction to only those members that also match the specified role

Issue #2212, opened 2024-04-23
Enhanced Check-Ins Report


mkimberlin:
Depends on #2211

**As an** Admin
**I want** a more informative Check-In report to track the completion of quarterly meetings with PDLs
**So that** I can easily identify PDLs who have completed, scheduled, or not started their Check-Ins.

## Acceptance Criteria:

 - [ ] The existing "Check-Ins" report will be replaced with a new report with the following functionalities:
   - [ ] Utilize the `MemberSelector` component to allow selecting PDLs.
   - [ ] By default, the report will automatically select all PDLs.
 - [ ] The report will list each selected PDL.
 - [ ] Each PDL will be displayed in an expandable section with a summary. This summary will include:
   - [ ] PDL's avatar, name, and title.
   - [ ] A status indicator for their assigned Check-Ins:
     - [ ] "Done": If the PDL has held and completed Check-Ins for all assigned team members in the current quarter.
     - [ ] "In Progress": If the PDL has any Check-Ins scheduled for the current quarter.
     - [ ] "Not Started": If the PDL has not scheduled or completed any Check-Ins for the current quarter.
 - [ ] Users should be able to select the quarter for which they want to run the report.
 - [ ] During the first month of a new quarter, the report will default to showing data for the previous quarter as the "current quarter". It will only switch to the new quarter after a full month has passed.

vhscom:
All ACs now covered and this story can be closed once we've done a little UAT.

vhscom:
Spawns: https://github.com/objectcomputing/check-ins/issues/2425

PR #2240, opened 2024-04-24
feat: add role-based filtering to member selector

vhscom:
closes: #2211

Allows the MemberSelector component to accept an `initialFilter` prop which it passes to MemberSelectorDialog, allowing for the dialog-based member filters to clamp on role based on filter passed like:

```js
initialFilters={[{ type: FilterType.ROLE, value: 'PDL' }]}
```

Demoed experience with Michael to be sure this met the criteria.
Note this change is inert at the moment as it will be built upon in #2212.

To test, add the above `initalFilters` prop to the `MemberSelector` component in `EmailPage` component, where `FilterType.ROLE` is imported from the `MemberSelectorDialog` component.

### DIFF
--- a/web-ui/src/components/member_selector/__snapshots__/MemberSelector.test.jsx.snap
+++ b/web-ui/src/components/member_selector/__snapshots__/MemberSelector.test.jsx.snap
@@ -607,7 +607,9 @@ exports[`MemberSelector > renders correctly with default props 1`] = `
                 <span
                   class="MuiTypography-root MuiTypography-body2 MuiListItemText-primary css-et1ao3-MuiTypography-root"
                 >
-                  No members selected
+                  No 
+                  members
+                   selected
                 </span>
               </div>
             </li>
--- a/web-ui/src/components/member_selector/member_selector_dialog/MemberSelectorDialog.jsx
+++ b/web-ui/src/components/member_selector/member_selector_dialog/MemberSelectorDialog.jsx
@@ -54,7 +54,7 @@ const DialogTransition = React.forwardRef((props, ref) => (
   <Slide direction="up" ref={ref} {...props} />
 ));
 
-const FilterType = Object.freeze({
+export const FilterType = Object.freeze({
   GUILD: 'Guild',
   TEAM: 'Team',
   TITLE: 'Title',
@@ -65,22 +65,45 @@ const FilterType = Object.freeze({
 });
 
 const propTypes = {
+  initialFilters: PropTypes.arrayOf(
+    PropTypes.shape({
+      type: PropTypes.oneOf(Object.values(FilterType)),
+      value: PropTypes.oneOfType([
+        PropTypes.string,
+        PropTypes.number,
+        PropTypes.bool
+      ])
+    })
+  ),
+  memberDescriptor: PropTypes.string,
   open: PropTypes.bool.isRequired,
   selectedMembers: PropTypes.arrayOf(PropTypes.object).isRequired,
   onClose: PropTypes.func,
   onSubmit: PropTypes.func
 };
 
-const MemberSelectorDialog = ({ open, selectedMembers, onClose, onSubmit }) => {
+const MemberSelectorDialog = ({
+  open,
+  memberDescriptor = 'Members',
+  initialFilters = [],
+  selectedMembers,
+  onClose,
+  onSubmit
+}) => {
   const { state, dispatch } = useContext(AppContext);
   const csrf = selectCsrfToken(state);
   const members = selectCurrentMembers(state);
+
+  // Use the first initial filter as the default filter type
+  const [initialFilter] = initialFilters;
 
   // Contains set of ids of checked members for instant add/remove operations
   const [checked, setChecked] = useState(new Set());
 
   const [nameQuery, setNameQuery] = useState('');
-  const [filterType, setFilterType] = useState(FilterType.TEAM);
+  const [filterType, setFilterType] = useState(
+    initialFilter?.type || FilterType.TEAM
+  );
   const [filterOptions, setFilterOptions] = useState(null);
   const [filter, setFilter] = useState(null);
   const [filteredMembers, setFilteredMembers] = useState([]);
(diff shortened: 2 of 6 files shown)


web-ui/src/components/member_selector/member_selector_dialog/MemberSelectorDialog.jsx
```diff
@@ -414,11 +443,11 @@ const MemberSelectorDialog = ({ open, selectedMembers, onClose, onSubmit }) => {
                   {...params}
                   variant="outlined"
                   label="Filter Members"
-                  placeholder={`Search for ${filterType.toLowerCase()}`}
+                  placeholder={`Search for ${filterType?.toLowerCase()}`}
                 />
               )}
               disablePortal
-              disabled={!filterOptions}
+              disabled={!filterOptions || initialFilters.length > 0}
```
vhscom, line 450: This is the `Autocomplete` component, which will throw if it receives `0` as an argument.

web-ui/src/components/member_selector/member_selector_dialog/MemberSelectorDialog.jsx
```diff
@@ -374,7 +403,7 @@ const MemberSelectorDialog = ({ open, selectedMembers, onClose, onSubmit }) => {
           </IconButton>
           <div className="toolbar-title-container">
             <Typography className="toolbar-title" variant="h5">
-              Select Members
+              Select {memberDescriptor ? memberDescriptor : 'Members'}
```
vhscom, line 406: This could probably be tightened up with `||` though I tend to prefer the ternary. Happy to adjust.

web-ui/src/components/member_selector/member_selector_dialog/MemberSelectorDialog.jsx
```diff
@@ -374,7 +403,7 @@ const MemberSelectorDialog = ({ open, selectedMembers, onClose, onSubmit }) => {
           </IconButton>
           <div className="toolbar-title-container">
             <Typography className="toolbar-title" variant="h5">
-              Select Members
+              Select {memberDescriptor ?? 'Members'}
```
mvolkmann, line 406: Maybe 'Members' should be the default value of the `memberDescriptor` prop.

web-ui/src/components/member_selector/member_selector_dialog/MemberSelectorDialog.jsx
```diff
@@ -414,11 +443,11 @@ const MemberSelectorDialog = ({ open, selectedMembers, onClose, onSubmit }) => {
                   {...params}
                   variant="outlined"
                   label="Filter Members"
-                  placeholder={`Search for ${filterType.toLowerCase()}`}
+                  placeholder={`Search for ${filterType?.toLowerCase()}`}
```
mvolkmann, line 446: I suppose `filterType` is always set, but you had to add the `?` to make tooling happy. Right?

web-ui/src/components/member_selector/member_selector_dialog/MemberSelectorDialog.jsx
```diff
@@ -103,6 +126,12 @@ const MemberSelectorDialog = ({ open, selectedMembers, onClose, onSubmit }) => {
       setFilteredMembers([]);
       setDirectReportsOnly(false);
       setSelectableMembers([]);
+    } else {
+      // If the dialog is opened with initial filters, set the initial filter
+      if (initialFilter) {
+        setFilterType(initialFilter.type);
+        setFilter(selectRoles(state).find(role => role.role === 'PDL'));
```
mvolkmann, line 133: Does this make it so only members that are PDLs are displayed?

web-ui/src/components/member_selector/member_selector_dialog/MemberSelectorDialog.jsx
vhscom: Not sure why I didn't think of that. Nice clean-up. Implemented.

web-ui/src/components/member_selector/member_selector_dialog/MemberSelectorDialog.jsx
vhscom: This was unnecessary. Removed. Good eyes!

web-ui/src/components/member_selector/member_selector_dialog/MemberSelectorDialog.jsx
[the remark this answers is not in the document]
vhscom: Logic updated to `!!initialFilters` as it's more terse. The logic is that, if there is an `initialFilter` (an object type) when the component is initialized, we clamp (disable) filter form controls by disabling using `!!initialFilter` which produces the same Boolean result as `Boolean(initialFilter)` and `initialFilters.length > 0`.

web-ui/src/components/member_selector/member_selector_dialog/MemberSelectorDialog.jsx
vhscom: This will set the default of the disabled form control, so it displays the correct default. In OP when we say "clamp" we're referring to disabling (with selected defaults) the form controls in the member select dialog. Since we only support one filter-type for #2211 I've updated this branch logic to be more explicit and look for the ROLE setting.

web-ui/src/components/member_selector/__snapshots__/MemberSelector.test.jsx.snap
```diff
@@ -607,7 +607,9 @@ exports[`MemberSelector > renders correctly with default props 1`] = `
                 <span
                   class="MuiTypography-root MuiTypography-body2 MuiListItemText-primary css-et1ao3-MuiTypography-root"
                 >
-                  No members selected
+                  No
```
mvolkmann, line 610: Why is each word on a separate line?

mvolkmann (approved):

mvolkmann (approved):

mkimberlin (approved):

PR #2247, opened 2024-04-25
fix: parse tsx files with the tseslint parser

vhscom:
Adds support for TSX parsing and documents type used given ESLint now has two config types. Information about the new FlatConfig from Nicholas Zakas can be found here: https://eslint.org/blog/2022/08/new-config-system-part-2/

Ref: https://typescript-eslint.io/packages/typescript-eslint/ on the set-up.

### DIFF
--- a/docs/getting-started/running/index.md
+++ b/docs/getting-started/running/index.md
@@ -66,8 +66,10 @@ yarn --cwd web-ui test
 Or simply `cd` to the `web-ui` directory and run `yarn test`. Tests using Snapshots are likely to fail with component updates. Those which fail to meet their Snapshot will be marked as failed. You can update the Snapshot by running the following command:
 
 ```shell
-yarn --cwd web-ui test -u
+yarn --cwd web-ui test -u # or simply `yarn test` followed by `u`
 ```
+
+Testing Library is installed in the UI project. You can find more information about Testing Library [here](https://testing-library.com/docs/react-testing-library/intro/).
 
 # Running the Server
 
--- a/docs/getting-started/running/index.md
+++ b/docs/getting-started/running/index.md
@@ -63,7 +63,11 @@ To run the UI tests, use the following command:
 yarn --cwd web-ui test
 ```
 
-Or simply `cd` to the `web-ui` directory and run `yarn test`.
+Or simply `cd` to the `web-ui` directory and run `yarn test`. Tests using Snapshots are likely to fail with component updates. Those which fail to meet their Snapshot will be marked as failed. You can update the Snapshot by running the following command:
+
+```shell
+yarn --cwd web-ui test -u
+```
 
 # Running the Server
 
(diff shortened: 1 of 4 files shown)


web-ui/eslint.config.js
```diff
@@ -31,6 +41,13 @@ export default [
       'react-hooks/exhaustive-deps': 'warn',
       ...vitestPlugin.configs.recommended.rules,
       'vitest/expect-expect': 'off',
+      '@typescript-eslint/no-unused-vars': 'off',
```
mvolkmann, line 44: Why do we want to turn these rules off?

web-ui/eslint.config.js
```diff
@@ -31,6 +41,13 @@ export default [
...
+      '@typescript-eslint/ban-ts-comment': 'off'
     }
+  },
+  {
+    // disable type-aware linting on JS and JSX files
```
mvolkmann, line 49: Why do we want to do this?

web-ui/yarn.lock
```diff
@@ -1363,7 +1363,7 @@
   dependencies:
     eslint-visitor-keys "^3.3.0"
 
-"@eslint-community/regexpp@^4.6.1":
+"@eslint-community/regexpp@^4.10.0", "@eslint-community/regexpp@^4.6.1":
```
mvolkmann, line 1366: Are we using two versions of this same package?

web-ui/eslint.config.js
vhscom: That was protoype code. It's been removed.

web-ui/eslint.config.js
vhscom: That was protoype code. It's been removed.

web-ui/yarn.lock
vhscom: We're not, but one of our transient dependencies may be.

docs/getting-started/running/index.md
```diff
@@ -63,10 +63,14 @@ To run the UI tests, use the following command:
 yarn --cwd web-ui test
 ```
 
-Or simply `cd` to the `web-ui` directory and run `yarn test`.
+Or simply `cd` to the `web-ui` directory and run `yarn test`. Tests using Snapshots are likely to fail with component updates. Those which fail to meet their Snapshot will be marked as failed. You can update the Snapshot by running the following command:
+
+```shell
+yarn --cwd web-ui test -u
+```
```
mvolkmann, line 70: I'm thinking you should also mention the option of just pressing the "u" key after failed snapshot tests are reported when the tests are run with "yarn test" (runs in watch mode).

ZacharyKlein (approved):

mvolkmann (approved):

jackkeller (approved):
LGTM 🚀

ZacharyKlein (approved):

mvolkmann (approved):

PR #2248, opened 2024-04-25
Feature 2244 avatar click

mvolkmann:


web-ui/src/components/menu/Menu.jsx
```diff
@@ -20,7 +20,6 @@ import fileDownload from 'js-file-download';
 import { useLocation, Link } from 'react-router-dom';
 import { AppContext } from '../../context/AppContext';
 import { getAvatarURL } from '../../api/api';
-import AvatarMenu from '@mui/material/Menu';
```
mvolkmann, line 23: Clicking the avatar no longer displays a menu.

web-ui/src/components/menu/Menu.jsx
```diff
@@ -183,10 +181,6 @@ function Menu() {
     return links;
   };
 
-  const handleClick = event => {
```
mvolkmann, line 186: This is now handled by a `Link` component.

web-ui/src/components/menu/Menu.jsx
```diff
@@ -304,10 +298,6 @@ function Menu() {
     setFeedbackOpen(false);
   };
 
-  const closeAvatarMenu = () => {
```
mvolkmann, line 307: There's no longer an avatar menu to close.

web-ui/src/components/menu/Menu.jsx
```diff
@@ -430,14 +420,8 @@ function Menu() {
             <MenuIcon />
           </IconButton>
         </Toolbar>
-        <div
-          ref={anchorRef}
-          aria-controls={open ? 'menu-list-grow' : undefined}
-          aria-haspopup="true"
-          onClick={handleToggle}
-        >
+        <Link to={`/profile/${id}`}>
```
mvolkmann, line 423: The avatar is now wrapped in a `Link` that goes to the Profile page.

web-ui/src/components/menu/Menu.jsx
```diff
@@ -430,14 +420,8 @@ function Menu() {
...
           <Avatar
-            onClick={handleClick}
```
mvolkmann, line 440: The click is now handled by the `Link` that wraps this `Avatar` component.

web-ui/src/components/menu/Menu.jsx
```diff
@@ -447,42 +431,7 @@ function Menu() {
               textDecoration: 'none'
             }}
           />
-          <AvatarMenu
```
mvolkmann, line 450: No more avatar menu.

web-ui/src/components/menu/__snapshots__/Menu.test.jsx.snap
```diff
@@ -33,8 +33,8 @@ exports[`<Menu /> > renders correctly 1`] = `
           />
         </button>
       </div>
-      <div
-        aria-haspopup="true"
+      <a
+        href="/profile/undefined"
```
mvolkmann, line 37: I wonder if the reason this contains "undefined" is that in the test we don't have a user id.

web-ui/src/components/menu/__snapshots__/Menu.test.jsx.snap
```diff
@@ -436,8 +436,8 @@ exports[`<Menu /> > renders correctly for admin 1`] = `
...
```
mvolkmann, line 440: See the previous comment.

web-ui/src/components/menu/__snapshots__/Menu.test.jsx.snap
```diff
@@ -911,8 +911,8 @@ exports[`<Menu /> > renders correctly for pdl 1`] = `
...
```
mvolkmann, line 915: See the previous comment.

S78901 (approved):

web-ui/src/components/menu/__snapshots__/Menu.test.jsx.snap
vhscom: In `Menu.test.jsx` currently I see us taking snapshots, but I don't see us making any assertions yet. In this case, the `${id}` is set at runtime based on the user profile. If we want to test various render conditions, and just _that_ it renders, I think we have room to beef up our coverage.

vhscom (dismissed):
:shipit:

web-ui/src/components/menu/__snapshots__/Menu.test.jsx.snap
mvolkmann: I added a test to verify that the Link around the Avatar uses the id value.

web-ui/src/components/menu/Menu.test.jsx
```diff
@@ -84,4 +88,19 @@ describe('<Menu />', () => {
...
+      <AppContextProvider value={initialState}>
+        <BrowserRouter>
+          <Menu />
+        </BrowserRouter>
+      </AppContextProvider>
+    );
+    const link = container.querySelector('header > a');
+    const href = link.getAttribute('href');
+    const lastIndex = href.lastIndexOf('/');
+    const id = href.substring(lastIndex + 1);
```
vhscom, line 103: Another way to grab the `id` could be to split the string into an array and pop off the last element:
```js
const id = href.split('/').pop()
```

vhscom (approved):
Love seeing some examples of how we're component testing. I saw we also have [Testing Library](https://testing-library.com/docs/react-testing-library/intro/), which can be useful as its API surface tends to nudge implementations towards semantic markup and use of ARIA where it matters most (e.g. component suspense states).

S78901 (approved):

PR #2256, opened 2024-04-25
Feature 2254 roles page selection

mvolkmann:


web-ui/src/components/admin/roles/Roles.jsx
```diff
@@ -62,10 +62,30 @@ const Roles = () => {
   memberProfiles?.sort((a, b) => a.name.localeCompare(b.name));
 
   useEffect(() => {
-    if (roles) {
```
mvolkmann, line 65: I don't think we want to get roles from state now. I think we only want to get them from a query parameter.

web-ui/src/components/admin/roles/Roles.jsx
```diff
@@ -62,10 +62,30 @@ const Roles = () => {
...
-      setSelectedRoles(roles.map(roleObj => roleObj.role));
-    }
-  }, [roles]);
+    const url = new URL(location.href);
```
mvolkmann, line 65: This gets the current searchText and selectedRoles to use from query parameters.

web-ui/src/components/admin/roles/Roles.jsx
```diff
@@ -62,10 +62,30 @@ const Roles = () => {
...
+    setSearchText(url.searchParams.get('searchText') ?? '');
+    const selectedRoles = url.searchParams.get('selectedRoles');
+    const roles = selectedRoles?.split(',') ?? [];
+    if (roles.length === 0) roles.push('ADMIN', 'MEMBER', 'PDL');
```
mvolkmann, line 69: I don't think we ever want to start with no roles selected. So if none are present in the query parameter then we make them all selected.

web-ui/src/components/admin/roles/Roles.jsx
```diff
@@ -62,10 +62,30 @@ const Roles = () => {
...
+    setSelectedRoles(roles);
+  }, []);
```
mvolkmann, line 71: This `useEffect` only runs each time the page is loaded.

web-ui/src/components/admin/roles/Roles.jsx
```diff
@@ -62,10 +62,30 @@ const Roles = () => {
...
+
+  useEffect(() => {
```
mvolkmann, line 73: This runs every time the user changes the searchText or selectedRoles.
It updates the query parameters in the URL.

web-ui/src/components/admin/roles/__snapshots__/Roles.test.jsx.snap
```diff
@@ -75,7 +75,7 @@ exports[`renders correctly 1`] = `
               <p
                 class="MuiFormHelperText-root MuiFormHelperText-sizeMedium MuiFormHelperText-contained MuiFormHelperText-filled css-1wc848c-MuiFormHelperText-root"
               >
-                Showing 2/2 roles
+                Showing 3/2 roles
```
mvolkmann, line 78: "3/2" seems wrong. Maybe I need to dig into this test more.

vhscom (approved):

S78901 (approved):

PR #2261, opened 2024-04-26
users page now holds state in query parameters

mvolkmann:


web-ui/src/components/admin/users/Users.jsx
```diff
@@ -69,6 +69,31 @@ const Users = () => {
       ? selectNormalizedMembersAdmin(state, searchText)
       : selectNormalizedMembers(state, searchText);
 
+  useEffect(() => {
```
mvolkmann, line 72: On page load this updates the state based on query parameters.

web-ui/src/components/admin/users/Users.jsx
```diff
@@ -69,6 +69,31 @@ const Users = () => {
...
+    setOpen(addUser === 'true');
+
+    const includeTerminated = url.searchParams.get('includeTerminated');
+    setIncludeTerminated(includeTerminated === 'true');
+
+    const search = url.searchParams.get('search') || '';
+    setSearchText(search);
+  }, []);
+
+  useEffect(() => {
```
mvolkmann, line 85: Every time a piece of the page state changes, this updates the query parameters in the URL.

S78901 (approved):

vhscom (approved):
LGTM. If we find enough URL-controlled state persistence, perhaps we can build towards a util method, and a single-source were we can manage QSPs.

PR #2270, opened 2024-04-26
Feature 2220/stepper review status admin

S78901:
Screenshot 2024-04-26 at 12 58 59 PM


web-ui/src/components/reviews/periods/ReviewPeriodStepper.jsx
```diff
@@ -0,0 +1,24 @@
...
+import Stepper from '@mui/material/Stepper';
+import Step from '@mui/material/Step';
+import StepLabel from '@mui/material/StepLabel';
+
+const steps = [
+  'Review Period Is Open',
+  'Review Period Is Closed'
+];
+
+export default function HorizontalLinearAlternativeLabelStepper({reviewPeriod}) {
```
S78901, line 12: Here we pass in the full reviewPeriod object. If we only need to pass in the status 'open', this could be simplified.

web-ui/src/components/reviews/periods/ReviewPeriodStepper.jsx
```diff
@@ -0,0 +1,24 @@
...
+  return (
+    <Box sx={{ width: '100%' }}>
+      <Stepper activeStep={reviewPeriod?.open === true ? 0 : 1} alternativeLabel>
```
S78901, line 15: Checking for whether the review period status 'open' is true or false. If true, show the activeStep as being option 0 in the steps array, otherwise show option 1 as active.

S78901:
@mkimberlin Please review information and the image in the description to verify that this is what you had in mind.

mkimberlin:
> @mkimberlin Please review information and the image in the description to verify that this is what you had in mind.

Apologies. I gave insufficient AC because I was so deeply mired in looking at all these stories together. I have added an AC that clarifies the status values this needs to support.

S78901:
@mkimberlin Please review the new specs and let me know if this is what you were envisioning.

Screenshot 2024-05-03 at 11 58 22 AM

mkimberlin:
> @mkimberlin Please review the new specs and let me know if this is what you were envisioning.
> 
> Screenshot 2024-05-03 at 11 58 22 AM
> 

Looks perfect.

web-ui/src/components/reviews/periods/ReviewPeriodStepper.jsx
```diff
@@ -0,0 +1,34 @@
+import React, { useEffect, useState } from 'react';
+import Box from '@mui/material/Box';
+import Stepper from '@mui/material/Stepper';
+import Step from '@mui/material/Step';
+import StepLabel from '@mui/material/StepLabel';
+import optionsArr from './reviewStatus.json';
```
mvolkmann, line 6: Is this used in more than one place?

web-ui/src/components/reviews/periods/ReviewPeriodStepper.jsx
```diff
@@ -0,0 +1,34 @@
...
+
+  useEffect(() => {
+    let index = 0;
+    let revPeriod = reviewPeriod?.reviewStatus;
+    optionsArr.map((opt, i) => {
+      if (opt.option === revPeriod) {
+        index = i;
+      }
+    });
+    setActiveIndex(toInteger(index));
```
mvolkmann, line 20: Isn't `index` already an integer?

web-ui/src/components/reviews/periods/ReviewPeriodStepper.jsx
```diff
@@ -0,0 +1,34 @@
...
+import Step from '@mui/material/Step';
+import StepLabel from '@mui/material/StepLabel';
+import optionsArr from './reviewStatus.json';
+import { toInteger } from 'lodash';
+
+export default function ReviewPeriodStepper({ reviewPeriod }) {
+  const [activeIndex, setActiveIndex] = useState(0);
+
+  useEffect(() => {
+    let index = 0;
```
mvolkmann, line 13: I think there's a shorter way to do this. Maybe something like this:

```js
const index = optionsArr.findIndex(opt => opt.option === revPeriod);
```

web-ui/src/components/reviews/periods/ReviewPeriodStepper.jsx
S78901: Not yet, but it may be eventually. Wanted to keep it separate for visibility.

web-ui/src/components/reviews/periods/ReviewPeriodStepper.jsx
S78901: This worked fine, thanks for the suggestion.

web-ui/src/components/reviews/periods/ReviewPeriodStepper.jsx
S78901: Yes, we don't need this anymore.

jackkeller (approved):
🎉 Approved! 🌟

mvolkmann (approved):

mkimberlin (approved):

PR #2271, opened 2024-04-26
Edit Skills pages now gets state from query parameters

mvolkmann:


web-ui/src/pages/EditSkillsPage.jsx
```diff
@@ -20,11 +20,37 @@ const EditSkillsPage = () => {
   const allSkills = selectOrderedSkills(state);
   const pendingSkills = selectPendingSkills(state);
 
+  useEffect(() => {
```
mvolkmann, line 23: On page load this gets all the settings from query parameters.

web-ui/src/pages/EditSkillsPage.jsx
```diff
@@ -20,11 +20,37 @@ const EditSkillsPage = () => {
...
+    setOpen(combining === 'true');
+
+    const search = url.searchParams.get('search') || '';
+    setSearchText(search);
+
+    const showAll = url.searchParams.get('showAll');
+    setShowAllSkills(showAll === 'true');
+  }, []);
+
+  useEffect(() => {
```
mvolkmann, line 36: Any time a setting changes, this updates the query parameters.
I used a slightly different approach this time where query parameters are only present when they have a non-default value. I should probably go back to my earlier PRs that are similar and update them to take this approach. It results in cleaner looking URLs.

S78901 (approved):

Issue #2217, opened 2024-04-23
Replace isOpen with status


mkimberlin:
**As an Administrator**, I want the `isOpen` field for review periods to be converted to an enumeration field named `status` with potential values `Planning`, `Awaiting Approval`, `Open`, and `Closed` **so that** I can easily understand and manage the workflow of each review period.

## Acceptance Criteria
- [ ] The isOpen field on the ReviewPeriod model is removed.
- [ ] A new field named status of type enumeration is added to the ReviewPeriod model with the following allowed values: Planning, Awaiting Approval, Open, and Closed.
- [ ] Existing data is migrated such that true in the isOpen field becomes Open in the new status field and false becomes Closed.

mjperry91:
https://github.com/objectcomputing/check-ins/pull/2273

S78901:
@mjperry91 Please let me know once this is merged so I can finalize Issue #2220.

PR #2273, opened 2024-04-26
Feature 2217/replace isOpen with status

mjperry91:
More Info: https://github.com/objectcomputing/check-ins/issues/2217


mjperry91:
General Testing

Adding a review we can see the boolean open parameter has been replaced in the request with the status enum of OPEN

Adding review with new status field

Archiving a review is successful and we can see the status enum is CLOSED.

Archive review with new status field

server/src/main/java/com/objectcomputing/checkins/services/reviews/ReviewPeriod.java
```diff
@@ -35,9 +35,9 @@ public class ReviewPeriod {
     private String name;
 
     @NotNull
-    @Column(name = "open")
-    @Schema(description = "Whether or not the review period is open")
-    private Boolean open = true;
+    @Column(name = "status")
+    @Schema(description = "The current status of the review period")
+    private String status;
```
ZacharyKlein, line 40: Is there a reason to not use the `ReviewStatus` itself as the type of this column? I think this would simplify the code a little bit and also make the relationship of `ReviewPeriod` and `ReviewStatus` more clear.

ZacharyKlein (changes requested):

server/src/main/java/com/objectcomputing/checkins/services/reviews/ReviewPeriod.java
mjperry91: Nope, just from the initial passthrough. I've updated it to the use them Enum.

ZacharyKlein (approved):

ZacharyKlein (approved):

ZacharyKlein (approved):

web-ui/src/components/reviews/periods/ReviewPeriods.jsx
```diff
@@ -29,27 +29,24 @@ import TextField from '@mui/material/TextField';
 import Tooltip from '@mui/material/Tooltip';
 import Typography from '@mui/material/Typography';
 
-import { styled } from '@mui/material/styles';
+import {styled} from '@mui/material/styles';
```
vhscom, line 32: You'll want to install prettier to take advantage of the default formatter we've configured, and reformat the JSX files. Other than that, looks good to me.

vhscom (dismissed):
Pre-approved under assumption of JSX formatting.

vhscom (approved):

ZacharyKlein (approved):

Issue #2268, opened 2024-04-26
Keep context in URL: View Feedback


mkimberlin:
**As an Administrator**, I want the URL for the view feedback page to include contextual information **so that** I can refresh the page without losing my selection.

## Acceptance Criteria:

 - [x] The URL should include a parameter search with the value of the search string entered in the search bar (if any).
 - [x] The URL should include a parameter with the value the show all toggle.
 - [x] The URL should include a parameter with the value of the "Show requests sent within" dropdown
 - [x] The URL should include a parameter with the value of the "Sort by" dropdown
 - [x] Refreshing the page with any of these parameters in the URL should maintain the previously selected context on the page

mvolkmann:
I accidentally did this work in the feature-2265-team-directory-query-parameters branch.

PR #2276, opened 2024-04-26
Team Results page now gets settings from query parameters

mvolkmann:


web-ui/src/components/team-results/TeamResults.jsx
```diff
@@ -38,7 +38,9 @@ const displayName = 'TeamResults';
 const TeamResults = () => {
   const { state } = useContext(AppContext);
   const loading = selectTeamsLoading(state);
+  const [open, setOpen] = useState(false);
```
mvolkmann, line 41: This tracks whether the "Add Team" dialog is open.

web-ui/src/components/team-results/TeamResults.jsx
```diff
@@ -47,10 +49,38 @@ const TeamResults = () => {
         key={`team-summary-${team.id}`}
         index={index}
         team={team}
+        onTeamSelect={setSelectedTeamId}
+        selectedTeamId={selectedTeamId}
       />
     );
   });
 
+  useEffect(() => {
```
mvolkmann, line 58: On page load this gets settings from query parameters.

web-ui/src/components/team-results/TeamResults.jsx
```diff
@@ -47,10 +49,38 @@ const TeamResults = () => {
...
+    setAddingTeam(addNew === 'true');
+
+    const search = url.searchParams.get('search') || '';
+    setSearchText(search);
+
+    const selectedTeamId = url.searchParams.get('team') || '';
+    setSelectedTeamId(selectedTeamId);
+  }, []);
+
+  useEffect(() => {
```
mvolkmann, line 71: When any setting changes, this update the URL query parameters.

web-ui/src/pages/ViewFeedbackPage.jsx
```diff
@@ -107,6 +107,36 @@ const ViewFeedbackPage = () => {
   const [dateRange, setDateRange] = useState(DateRange.THREE_MONTHS);
```
mvolkmann, line 107: The changes in this file are for the story "Keep context in URL: View Feedback #2268".

web-ui/src/pages/ViewFeedbackPage.jsx
```diff
@@ -107,6 +107,36 @@ const ViewFeedbackPage = () => {
...
   const [includeAll, setIncludeAll] = useState(false);
 
+  useEffect(() => {
```
mvolkmann, line 110: On page load this gets all the settings from query parameters.

web-ui/src/pages/ViewFeedbackPage.jsx
```diff
@@ -107,6 +107,36 @@ const ViewFeedbackPage = () => {
...
+    setSearchText(search);
+
+    const showAll = url.searchParams.get('showAll');
+    setIncludeAll(showAll === 'true');
+
+    const sort = url.searchParams.get('sort') || '';
+    setSortValue(sort);
+  }, []);
+
+  useEffect(() => {
```
mvolkmann, line 126: When any setting changes, this updates the URL query parameters.

web-ui/src/helpers/query-parameters.js
```diff
@@ -0,0 +1,56 @@
...
+ * @property {string} default
+ * @property {string} value
+ * @property {(string) => void} setter - takes query parameter value and updates state
+ * @property {(any) => string} toQP - takes state value and returns query parameter value
+ */
+
+/**
+ * @param {(QPBoolean | QPString[]} qps - query parameters
+ */
+export const queryParameterSetup = qps => {
```
vhscom, line 24: Should we consider this a [custom hook](https://react.dev/learn/reusing-logic-with-custom-hooks), and start the name with `useQueryParameterSetup` to make it clear this function is intended only for use within react components by way of convention? I think this might qualify...

vhscom (dismissed):

web-ui/src/helpers/query-parameters.js
vhscom: I did a little more digging, and custom hooks are intended for encapsulating React state. Here we're not encapsulating state, and so then I think this would not qualify as a custom hook. Though I'm not 100%.

mvolkmann: Interesting idea. I suppose not all hooks encapsulate state. For example, this does not:

```js
useEffect(() => {
  console.log('page was loaded');
}, []);
```

web-ui/src/pages/PermissionsPage.jsx
```diff
@@ -72,21 +73,20 @@ const EditPermissionsPage = () => {
...
+      name: 'role',
+      default: roles[0],
+      value: selectedRole,
+      setter(value) {
+        const role = roles.find(r => r.role === value);
+        setSelectedRole(role);
+      },
+      toQP(selectedRole) {
+        return selectedRole.role;
+      }
```
vhscom, line 87: I believe this is the line causing one of the CI tests to fail. Looks like `selectedRole` is sometimes `undefined` when making this call. I see we're grabbing it from state, so I suspect we need to defer QSP set-up until we have the needed state.

web-ui/src/helpers/query-parameters.js
mvolkmann: I did decide to rename `queryParameterSetup` to `useQueryParameters` since it does call `useEffect`. Great suggestions!

web-ui/src/pages/PermissionsPage.jsx
mvolkmann: Great catch! I wonder how I can implement that delay. It's not obvious to me.
I could use `setInterval` and just check repeatedly.
Did you have another approach in mind?

web-ui/src/helpers/query-parameters.js
```diff
@@ -0,0 +1,66 @@
...
+    const params = {};
+
+    // Add query parameters listed in qps that do not have their default value.
+    for (const qp of qps) {
+      let { toQP, value } = qp;
+      if (toQP) value = toQP(value);
+      if (value && !compare(value, qp.default)) params[qp.name] = value;
+    }
+
+    // Add query parameters that are not listed in qps.
```
S78901, line 53: Great work.

S78901 (dismissed):

S78901 (approved):

web-ui/src/pages/PermissionsPage.jsx
```diff
@@ -83,7 +83,7 @@ const EditPermissionsPage = () => {
         setSelectedRole(role);
       },
       toQP(selectedRole) {
-        return selectedRole ? selectedRole.role : '';
+        return selectedRole?.role ?? '';
```
vhscom, line 86: If this doesn't work, a `?? { role: '' }` may be what you're looking for. It's difficult to tell at eyeshot without types.

vhscom (approved):
Excited to see our links come alive for deep-linking within the app!

web-ui/src/pages/PermissionsPage.jsx
mvolkmann: In this case the return value is used as a query parameter value, so it needs to be a string.

PR #2304, opened 2024-04-30
Feature 2298 check ins report query parameters

mvolkmann:


web-ui/src/helpers/query-parameters.js
```diff
@@ -21,8 +21,21 @@ import { useEffect } from 'react';
 /**
  * @param {(QPBoolean | QPString)[]} qps - query parameters
  */
-export const useQueryParameters = qps => {
+export const useQueryParameters = (
+  qps,
+  requirements = [],
```
mvolkmann, line 26: Some pages require data to be loaded before query parameters can be processed. This page is one of them. We must have an array of all possible PDL objects in order to call `setSelectedPdls` with an array of PDL objects that have ids that match those in the `pdls` query parameter. Support this was extremely difficult and did not increase my love of React. ;-)

`requirements` is an array of data values that must be set in order to process the query parameters. It defaults to an empty array because most pages do not need this.

`processedQPs` is an object created by a call to `useRef` in the calling page. It holds a Boolean that indicates whether we have finished processing the query parameters. This can only happen after all the `requirements` values are present. I'm using a ref because the value needs to be maintained across multiple renders of the page which React loves to do. I'm open to suggestions for other ways to maintain this flag.

web-ui/src/helpers/query-parameters.js
```diff
@@ -21,8 +21,21 @@ import { useEffect } from 'react';
...
+  processedQPs = null
+) => {
+  // This examines all the query parameters and
+  // sets the state value associated with each of them.
   useEffect(() => {
+    if (processedQPs?.current) return;
```
mvolkmann, line 32: If the page supplied a value for `processedQPs` AND the value it holds is `true` then we don't need to run the code that follows again.

web-ui/src/helpers/query-parameters.js
```diff
@@ -21,8 +21,21 @@ import { useEffect } from 'react';
...
+
+    const haveRequirements = requirements.every(req =>
```
mvolkmann, line 34: If `requirements` is an empty array, this will be set to `true` which is what we want.

web-ui/src/helpers/query-parameters.js
```diff
@@ -21,8 +21,21 @@ import { useEffect } from 'react';
...
+      Array.isArray(req) ? req.length > 0 : req !== null && req !== undefined
```
mvolkmann, line 35: When a required value is an array, the requirement is met when the array is not empty.
Otherwise the requirement is met when the value is not `null` or `undefined`.

web-ui/src/helpers/query-parameters.js
```diff
@@ -21,8 +21,21 @@ import { useEffect } from 'react';
...
+    );
+    if (!haveRequirements) return;
```
mvolkmann, line 37: If the requirements are not met then we cannot processed the query parameters yet.

web-ui/src/helpers/query-parameters.js
```diff
@@ -34,13 +47,21 @@ export const useQueryParameters = qps => {
         qp.setter(v || qp.default);
       }
     }
-  }, []);
+    if (processedQPs) processedQPs.current = true;
```
mvolkmann, line 50: If the caller supplied this ref, we need to set its value to to `true` to include that we have processed the query parameters. This prevents use from doing it again later.

web-ui/src/helpers/query-parameters.js
```diff
@@ -34,13 +47,21 @@ export const useQueryParameters = qps => {
...
+  }, requirements);
 
   const dependencies = qps.map(qp => qp.value);
 
+  // This updates the query parameters in the URL
+  // when their associated state values change.
   useEffect(() => {
+    const haveRequirements = requirements.every(req =>
```
mvolkmann, line 58: If the requirements have not been met then we cannot update the query parameters.

web-ui/src/helpers/query-parameters.js
```diff
@@ -55,10 +76,12 @@ export const useQueryParameters = qps => {
       if (!qps.some(qp => qp.name === k)) params[k] = v;
     }
 
-    if (Object.keys(params).length) {
-      newUrl += '?' + new URLSearchParams(params).toString();
+    const search = Object.keys(params).length
+      ? '?' + new URLSearchParams(params).toString()
+      : '';
+    if (search !== url.search) {
```
mvolkmann, line 82: Only update the URL if the query parameters have changed.

web-ui/src/pages/CheckinsReportPage.jsx
```diff
@@ -1,14 +1,16 @@
-import React, { useContext, useEffect, useState } from 'react';
+import React, { useContext, useEffect, useRef, useState } from 'react';
+
+import { TextField } from '@mui/material';
```
mvolkmann, line 3: Minor reordering of imports.

web-ui/src/pages/CheckinsReportPage.jsx
```diff
@@ -24,8 +26,34 @@ const CheckinsReportPage = () => {
...
+          if (isArrayPresent(newPdls)) {
+            const ids = newPdls.map(pdl => pdl.id);
+            return ids.join(',');
+          } else {
+            return [];
+          }
+        }
+      }
+    ],
+    [pdls],
```
mvolkmann, line 53: The requirements for processing query parameters for this page are to have entries in the `pdl` array.

web-ui/src/pages/CheckinsReportPage.jsx
```diff
@@ -24,8 +26,34 @@ const CheckinsReportPage = () => {
     const bPieces = b.name.split(' ').slice(-1);
     return aPieces.toString().localeCompare(bPieces);
   });
+
   const [filteredPdls, setFilteredPdls] = useState(pdls);
 
+  const processedQPs = useRef(false);
+  useQueryParameters(
```
mvolkmann, line 33: This adds support for the `pdl` query parameter which keeps track of the PDLs the user has selected.

vhscom (approved):
:shipit:

mkimberlin (approved):

PR #2321, opened 2024-05-02
Feature 2309 team skills report query parameters

mvolkmann:


web-ui/src/components/member_selector/MemberSelector.jsx
```diff
@@ -80,14 +80,9 @@ const MemberSelector = ({
   className,
   style
 }) => {
-  const isControlled = !!selected && Array.isArray(selected);
```
mvolkmann, line 83: Previously this component maintained its own state of selected members in addition to allowing them to be passed in by a component that also maintains that same state. I changed this to require the selected members to be passed in using the `selected` prop. The function specified by the `onChange` prop is called every time the selection of members changes so the parent component can update its state. This simplifies the code and works much better with our custom hook `useQueryParameters`.

web-ui/src/pages/TeamSkillReportPage.jsx
```diff
@@ -1,4 +1,6 @@
-import React, { useContext, useState } from 'react';
+import React, { useContext, useRef, useState } from 'react';
+
+import { Autocomplete, Button, TextField, Typography } from '@mui/material';
```
mvolkmann, line 3: I combined some MUI imports.

web-ui/src/pages/TeamSkillReportPage.jsx
```diff
@@ -129,7 +163,8 @@ const TeamSkillReportPage = () => {
       <MemberSelector
         className="team-skill-member-selector"
         listHeight={300}
-        onChange={selected => setSelectedMembers(selected)}
+        onChange={setSelectedMembers}
+        selected={selectedMembers}
```
mvolkmann, line 167: The `MemberSelector` component now requires the `selected` prop.

S78901 (approved):

vhscom (approved):

web-ui/src/components/member_selector/MemberSelector.jsx
vhscom: This component is a little tricky I'm noticing, as both the earlier and the current implementations do not seem to update the member selection when the `selected` prop changes (e.g. during runtime as data becomes available from the server) in the controlling parent as I'd expected. Typically with controlled components, changes to the _controlling_ props (like `selected` in this case) should cause a re-render of the child. That's outside the scope here but worth noting.

vhscom: Was a little tricky. Now it works like a dream. ✨

S78901 (approved):

PR #2277, opened 2024-04-29
[2228] Allow planning review period timeline

S78901:

PR #2327, opened 2024-05-03
Bugfix 2227/fixing serialization of local date time

mjperry91:
Fixing LocalDateTime serialization issue that was found here: https://github.com/objectcomputing/check-ins/pull/2277

In the future it is probably simpler to set this format globally using jackson.date-format configuration setting, but that has the potential to cause issues across other APIs and would require more extensive testing. I think this path is good for now to get Syd unblocked, then maybe write up a ticket to look into that.


ZacharyKlein (approved):

S78901 (approved):

S78901:
You may want to check on the Gradle build failure.

mjperry91:
> You may want to check on the Gradle build failure.

Failing tests are from a commit on develop that I mentioned earlier.

PR #2328, opened 2024-05-03
Feature 2279/view review assignment api

### LINKED ISSUE #2279: Create service for viewing review assigments
**As a Manager**, I want to view a list of all my review assignments **so that** I can stay on top of who needs to be reviewed.

## Acceptance Criteria:
 - [x] The service allows a GET request to an endpoint like /reviews/assignments/${reviewPeriodId} with a optional query parameter reviewerId
 - [x] If no reviewerId are provided, the service returns a list of all review assignments for the review period whose ID is provided in the path.
 - [x] If a reviewerId is provided, the service returns a list of assignments where the reviewerID matches.
 - [x] The service can only be called if the current user has the CAN_VIEW_REVIEW_ASSIGNMENTS permission (must be created)
 - [x] If **no assignments exist** for the review period, then a set of default review assignments should be returned:
   - [x] The default review assignments should include every current employee (as reviewee).
   - [x] Each employee should have their supervisor assigned as their default reviewer.
   - [x] The default assignments should not have been persisted, and should not have IDs (some employees may, ultimately, not be included in a review period)
 - [x] Each returned assignment object includes:
   - [x] id: The unique identifier for the assignment (if not providing default assignments).
   - [x] revieweeId: The ID of the employee being reviewed.
   - [x] reviewerId: The ID of the manager conducting the review.
   - [x] reviewPeriodId: The ID of the review period that this assignment relates to.
   - [x] approved: A boolean indicating if the review has been approved (true) or not (false).

ZacharyKlein:
Created API and tests for viewing review assignments

Resolves #2279

### DIFF
--- a/server/src/main/java/com/objectcomputing/checkins/services/permissions/Permission.java
+++ b/server/src/main/java/com/objectcomputing/checkins/services/permissions/Permission.java
@@ -36,7 +36,8 @@ public enum Permission {
   CAN_VIEW_ALL_CHECKINS("View all check-ins", "Check-ins"),
   CAN_UPDATE_ALL_CHECKINS("Update all check-ins, including completed check-ins", "Check-ins"),
   CAN_EDIT_SKILL_CATEGORIES("Edit skill categories", "Skill Categories"),
-  CAN_CREATE_REVIEW_ASSIGNMENTS("Create review assignments", "Reviews");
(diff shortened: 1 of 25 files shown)


mkimberlin (commented):
I feel like this needs tests for not having the necessary permission(s).

mjperry91 (approved):

PR #2330, opened 2024-05-06
chore: remove icon title for idempotent builds

vhscom:
Removes the `title` attribute from the `FontAwesomeIcon` in the GuildSummaryCard component as the presence of this attribute was causing the icon to invalidate its snapshot value with each test run.

### DIFF
--- a/web-ui/src/components/guild-results/GuildSummaryCard.jsx
+++ b/web-ui/src/components/guild-results/GuildSummaryCard.jsx
@@ -143,7 +143,6 @@ const GuildSummaryCard = ({ guild, index, isOpen, onGuildSelect }) => {
         style={{
           color: guild.community ? 'var(--oci-orange)' : 'var(--oci-light-blue)'
         }}
-        title={`This is a ${guild.community ? 'Community' : 'Guild'}.`}
       />
       <CardHeader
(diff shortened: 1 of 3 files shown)


mkimberlin (approved):

ZacharyKlein (approved):

web-ui/src/components/guild-results/GuildSummaryCard.jsx
```diff
@@ -143,7 +143,6 @@ const GuildSummaryCard = ({ guild, index, isOpen, onGuildSelect }) => {
         style={{
           color: guild.community ? 'var(--oci-orange)' : 'var(--oci-light-blue)'
         }}
-        title={`This is a ${guild.community ? 'Community' : 'Guild'}.`}
```
mvolkmann, line 146: Why do you want to remove this?

mvolkmann: There's an AC in the story that triggered adding this. It asks for the icons to have a tooltip.

vhscom: Michael requested we remove this for now to unblock CI.

PR #2333, opened 2024-05-06
chore: introduce app-wide type definitions

vhscom:
In support of #2212 this pull introduces app-wide type definitions to the codebase and specifies the `Member` and `PDL` types. Information on nullable and non-nullable types can be found here: https://jsdoc.app/tags-type

### DIFF
--- a/web-ui/src/app.d.js
+++ b/web-ui/src/app.d.js
@@ -1,0 +1,28 @@
+// Type definitions specified here will be available app-wide
+
+/**
+ * @typedef {Object} MemberProfile
+ * @property {?string} bioText
+ * @property {?Array} birthday
+ * @property {string} employeeId
+ * @property {?string} excluded
+ * @property {string} firstName
+ * @property {!string} id
+ * @property {string} lastName
+ * @property {string} location
+ * @property {?string} middleName
+ * @property {string} name
+ * @property {?string} pdlId
+ * @property {Array} startDate
+ * @property {?string} suffix
+ * @property {?string} supervisorid
+ * @property {?Array} terminationDate
+ * @property {string} title
+ * @property {?string} voluntary
+ * @property {string} workEmail
+ */
+
+/**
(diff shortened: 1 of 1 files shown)


vhscom:
Example usages:

```js
/**
 * Sort PDLS by the last name extracted from the full name.
 * @param {PDL} a - First PDL object
 * @param {PDL} b - Second PDL object
 * @returns {number} - Comparison result for sorting
 */
function sortByLastName(a, b) {
  if (!a.name || !b.name) return;
  const lastNameA = a.name.split(' ').slice(-1)[0];
  const lastNameB = b.name.split(' ').slice(-1)[0];
  return lastNameA.localeCompare(lastNameB);
}
```
```js
const [mappedPDLs, setMappedPDLs] = useState(/** @type {PDL[]} */ ([]));
const [selectedPdls, setSelectedPdls] = useState(/** @type {PDL[]} */ ([]));
```

web-ui/src/app.d.js
```diff
@@ -0,0 +1,28 @@
+// Type definitions specified here will be availalbe app-wide
+
+/**
+ * @typedef {Object} Member
+ * @property {?string} bioText
+ * @property {?Array} birthday
+ * @property {string} empolyeeId
```
mkimberlin, line 7: `employeeId`?

vhscom: The type name has been changed from `Member` to `MemberProfile` to better match the routes we call on the server for this data, in this case `/member-profiles` endpoint. The endpoint returns `employeeId` like:

Screenshot 2024-05-07 at 10 12 55 AM

mkimberlin: @vhscom I'm sorry. I should have been more clear. I was pointing out that it was misspelled.

vhscom: Good eyes. Thank you.

mkimberlin (approved):

PR #2334, opened 2024-05-06
Feature 2233/give check ins theming powers

jackkeller:
- adds _variables.css file to control theming natively
- overwrites a few things in App.jsx where native css variables were not doing the trick

PR #2335, opened 2024-05-07
Feature 2233/theme cherry picks

jackkeller:
- redid commits for this story
- adds _variables.css file to control theming natively
- overwrites a few things in App.jsx where native css variables were not doing the trick
- removes hardcoded color variables throughout the site
- makes a11y happy with contrast and most general things


web-ui/src/App.jsx
```diff
@@ -10,66 +10,79 @@ import { AppContextProvider } from './context/AppContext';
...
+} from '@mui/material/styles';
 
 import './App.css';
 
-const customHistory = createBrowserHistory();
+function getUserColorScheme() {
+  console.log('getting your scheme');
+  if (
+    window.matchMedia &&
+    window.matchMedia('(prefers-color-scheme: dark)').matches
```
mvolkmann, line 25: Does it work to change this to window?.matchMedia(... and remove line 24?
Same on line 30.

web-ui/src/App.jsx
```diff
@@ -10,66 +10,79 @@ import { AppContextProvider } from './context/AppContext';
...
-    },
-    secondary: {
-      light: '#ffe8a2',
-      main: '#feb672',
-      dark: '#c88645',
-      contrastText: '#000'
+const theme = extendTheme({
+  cssVarPrefix: 'checkins',
+  colorSchemes: {
+    light: {
```
mvolkmann, line 41: Do light and dark ever contain properties other than palette?
I'm wondering if that layer of the hierarchy can be removed.

web-ui/src/App.jsx
```diff
@@ -10,66 +10,79 @@ import { AppContextProvider } from './context/AppContext';
...
-    },
-    MuiTextField: {
-      defaultProps: {
-        variant: 'standard'
-      }
     }
   }
 });
 
+const customHistory = createBrowserHistory();
```
mvolkmann, line 72: Is this used?

web-ui/src/styles/_variables.css
```diff
@@ -0,0 +1,100 @@
+:root {
```
mvolkmann, line 1: What's the reason this filename starts with an underscore?

web-ui/src/styles/_variables.css
```diff
@@ -0,0 +1,100 @@
...
+  --primary-link-light: var(--primary-light);
+  --primary-link-light-hover: #6084d1;
+  --primary-link-dark: color-mix(in oklab, var(--oci-dark-blue), var(--white) 35%);
+  --primary-link-dark-hover: #4f5f8c;
+  --action: color-mix(in oklab, var(--black), var(--white) 50%);
+  --action-disabled: color-mix(in oklab, var(--black), var(--white) 5%);
+
+  --checkins-palette-action-disabledOpacity: 0.58;
+
+  /* MUI Prefixes */
```
mvolkmann, line 38: What is the connection between these variables and MUI? Are they only used to style MUI components?

web-ui/src/components/menu/Menu.jsx
```diff
@@ -334,151 +303,157 @@ function Menu() {
...
-        <Toolbar>
-          <IconButton
-            color="inherit"
-            aria-label="open drawer"
-            edge="start"
-            onClick={handleDrawerToggle}
-            className={classes.menuButton}
-            size="large"
+    <div className={classes.root}>
+      <CssVarsProvider>
```
vhscom, line 362: In `App.jsx` [we use this](https://github.com/objectcomputing/check-ins/pull/2335/files#diff-a11cac4bbb0df4570e6a2061d4e78c126ec3bd5e2af9e1533eae376032386e26R78) with `theme`. Is that why we don't need to supply the `theme` here as well, since we already have it from higher up in the component tree?

web-ui/src/styles/_variables.css
jackkeller: Mui provides a prefix and I figured this would be nice for scoping. In many instances omitting a prefix variable allows mui fallbacks.

web-ui/src/components/received_request_card/ReceivedRequestCard.jsx
```diff
@@ -21,21 +22,21 @@ const classes = {
...
+  // },
+  // [`& .${classes.greenTypography}`]: {
+  //   color: '#006400'
+  // },
+  // [`& .${classes.darkGrayTypography}`]: {
+  //   color: '#333333'
+  // },
+  // [`& .${classes.grayTypography}`]: {
+  //   color: 'gray'
+  // }
```
vhscom, line 39: Same question [mark had](https://github.com/objectcomputing/check-ins/pull/2334/files#r1592538204). If we want to keep the styles for now, could we add a comment in the code as to when they're safe to remove?

vhscom (approved):
Looks good to me. Left a couple comments.

web-ui/src/App.jsx
jackkeller: In some cases they are scoped to a component, but for mui to respect the variables then palette must be used, I would have preferred to have had this without palette too but sometimes that's the nature of a design system.

web-ui/src/App.jsx
jackkeller: Appears to be used below in Router `<Router history={customHistory}>`.

web-ui/src/App.jsx
jackkeller: I'll give it a quick try and see if the linting/testing finds it acceptable (it should).

web-ui/src/styles/_variables.css
jackkeller: I think this was just muscle memory from SCSS days, an underscore IIRC told the preprocessor to ignore outputting the file. If preferred then we could change this.

web-ui/src/styles/_variables.css
jackkeller: I believe so, MUI seems to be very driven by naming conventions.

web-ui/src/components/menu/Menu.jsx
jackkeller: Yes, but this was the fix that I found to include the ThemeToggle in Menu, however I could have possibly made that a `props.children` perhaps, I'm so used to `<slot />` I often forget about the children.

web-ui/src/components/received_request_card/ReceivedRequestCard.jsx
jackkeller: I think I just missed removing this, will do in a moment.

web-ui/src/App.jsx
jackkeller: Seems fine, adding to the batch update.

web-ui/src/components/menu/Menu.jsx
jackkeller: refactored this to remove the need for `CssVarsProvider` in Menu.jsx

Issue #2283, opened 2024-04-30
Create a service for deleting review assignments


mkimberlin:
**As a Manager**, I want to delete a review assignment I created in case of an error or if the review is no longer needed **so that** I can keep my review list accurate and avoid clutter.

## Acceptance Criteria:

 - [ ] The service allows a DELETE request to an endpoint like /reviews/assignments/{id} where {id} is the unique identifier for the assignment.
 - [ ] The service deletes the specified review assignment.
 - [ ] The service returns a success message upon successful deletion.

Issue #2281, opened 2024-04-30
Create service for updating a review assignment


mkimberlin:
**As a manager or admin**, I want to update review assignments **so that** I can adjust who is assigned to review an employee.

## Acceptance Criteria:
  - [x] The service allows a PUT request to an endpoint like /reviews/assignments/${id} with a JSON body containing:
    - [x] revieweeId: The ID of the employee being reviewed.
    - [x] reviewerId: The ID of the employee conducting the review.
    - [x] reviewPeriodId: The ID of the review period that the assignment is related to.
    - [x] approved: A boolean indicating if the review has been approved (true) or not (false).
  - [x] The service can only be called if the current user has the CAN_UPDATE_REVIEW_ASSIGNMENTS permission (must be created)
  - [x] The service only allows a review assigment to be adjusted if the reviewee is among the current user's reports, but **NOT** a direct report (i.e. at least one level removed).
  - [x] The service only allows a review assigment to have approved set to true if the current user has the CAN_APPROVE_ALL_REVIEW_ASSIGNMENTS permission (must be created) or if the reviewee is among their reports, but **NOT** a direct report (i.e. at least one level removed).
  - [x] The service returns a JSON response containing the details of the updatted assignment, including:
    - [x] id: The unique identifier for the assignment.
    - [x] revieweeId: The ID of the employee being reviewed.
    - [x] reviewerId: The ID of the manager conducting the review.
    - [x] reviewPeriodId: The ID of the review period that the review assignment relates to.
    - [x] approved: A boolean indicating if the review has been approved (true) or not (false).

PR #2336, opened 2024-05-07
Added delete and update endpoints, persistence, tests - #2281 and #2283

ZacharyKlein:


mjperry91 (approved):

server/src/main/java/com/objectcomputing/checkins/services/reviews/ReviewAssignmentServicesImpl.java
```diff
@@ -66,8 +71,26 @@ public Set<ReviewAssignment> findAllByReviewPeriodIdAndReviewerId(UUID reviewPer
         return reviewAssignments;
     }
 
+    @Override
+    public ReviewAssignment update(ReviewAssignment reviewAssignment) {
+        LOG.warn(String.format("Updating entity %s", reviewAssignment));
```
mvolkmann, line 76: Is `warn` the appropriate method here? That seems to indicate something is wrong.

mvolkmann (approved):

PR #2341, opened 2024-05-08
fix: restore pdl url query on check-ins report

vhscom:
Restores the query processing for MemberSelector default PDL selection on page load on the Check-Ins page.

### DIFF
--- a/web-ui/src/helpers/index.js
+++ b/web-ui/src/helpers/index.js
@@ -1,0 +1,6 @@
+export * from './celebration';
+export * from './checks';
+export * from './datetime';
+export * from './query-parameters';
+export * from './sanitizehtml';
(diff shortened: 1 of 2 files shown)


web-ui/src/helpers/index.js
```diff
@@ -0,0 +1,6 @@
+export * from './celebration';
+export * from './checks';
+export * from './datetime';
+export * from './query-parameters';
+export * from './sanitizehtml';
+export * from './strings';
```
vhscom, line 6: This barrel allows us to import directly from `helpers`, any of the non-test files in the helpers folder.

mvolkmann (approved):

S78901 (approved):

PR #2342, opened 2024-05-08
docs: add db connection info and favicon

vhscom:
Adds information about how to connect to the database and manage test data. Also adds a favicon.ico in the default location [JDT docs config specifies](https://just-the-docs.com/docs/configuration/#site-favicon) for automatic inclusion.

L-R: Before and with favicon
Screenshot 2024-05-08 at 11 54 53 AM


docs/getting-started/running/index.md
```diff
@@ -81,4 +81,24 @@ Testing Library is installed in the UI project. You can find more information ab
...
+
+You can connect to the database on the default port using the following command:
+
+```shell
+psql -h localhost -U postgres -d checkinsdb
+```
+
+The password for the dev database is `postgres`.
+
+Once connected, you can run SQL queries against the database, inspect the schema, and more.
```
mvolkmann, line 98: Is it the case that all the data in the database is encrypted? If so, does this prevent manually inserting and updating data interactively?

vhscom: I see `V40__create_extension.sql` adds the `pgcrypto` extension. I'm not deeply familiar with how it operates, though I have no problems connecting to the database and running SQL queries after migrations run. At runtime, depending on how one connects, the data would be encrypted in transit via SSL per my understanding.

mvolkmann: I think Michael said it doesn't work to manually insert data because it won't be encrypted. When the app tries to access manually inserted data, it will assume it is encrypted, try to unencrypted, and wind up with garbage data. Maybe we should add a note about that in this document.

vhscom: I was not aware of that. I pinged Michael for more information to improve the docs here.

mkimberlin: That is correct, but it is on a per column basis. Not all columns were deemed sensitive. Basically, you have to provide an AES_KEY on startup.

It is possible to manually insert, query, and manipulate data locally by providing the same key in any of your queries...but it's not super easy and none of the automatic data editor views will work for it.

vhscom: This is good information to have on-hand. I've added a new section to the docs in this pull describing the what and how for working with encrypted columns in the database.

mvolkmann (approved):

mjperry91 (approved):

PR #2344, opened 2024-05-08
Bugfix 2282/fix find assignments

ZacharyKlein:
Exclude members w/o supervisors from default review assignments


mjperry91 (approved):

mkimberlin (approved):

server/src/main/java/com/objectcomputing/checkins/services/reviews/ReviewAssignmentServicesImpl.java
```diff
@@ -113,6 +113,10 @@ private Set<ReviewAssignment> defaultReviewAssignments(UUID reviewPeriodId) {
         Set<ReviewAssignment> reviewAssignments = new HashSet<>();
 
         memberProfileRepository.findAll().forEach(memberProfile -> {
+            if(memberProfile.getSupervisorid() == null) {
+                return;
```
mvolkmann, line 117: Should this return an empty set instead of returning nothing?

ZacharyKlein: Should be fine, we're just iterating over the members and conditionally adding review assignments for those w/o supervisors.

PR #2359, opened 2024-05-10
Feature 2212/enhanced checkin warranty

vhscom:
Page now properly selects all PDLs when the user navigates to the enhanced Check-Ins Report

Also updates card treatment to display Latest Activity using a new `--muted` utility compatible with AA conformance in contrast ratio measured using Firefox dev tools, and as close to `4.5:1` contrast as possible using whole numbers.

### DIFF
--- a/web-ui/src/pages/CheckinsReportPage.jsx
+++ b/web-ui/src/pages/CheckinsReportPage.jsx
@@ -97,7 +97,7 @@ const CheckinsReportPage = () => {
           const newPdls = ids
             .map(id => pdls.find(pdl => pdl.id === id))
             .filter(Boolean);
-          setSelectedPdls(newPdls);
+          newPdls.length > 0 && setSelectedPdls(newPdls);
         },
         toQP(newPdls) {
           if (isArrayPresent(newPdls)) {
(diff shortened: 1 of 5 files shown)


web-ui/src/pages/CheckinsReportPage.jsx
```diff
@@ -97,7 +97,7 @@ const CheckinsReportPage = () => {
           const newPdls = ids
             .map(id => pdls.find(pdl => pdl.id === id))
             .filter(Boolean);
-          setSelectedPdls(newPdls);
+          newPdls.length > 0 && setSelectedPdls(newPdls);
```
mvolkmann, line 100: If `newPdls` becomes empty, does it to clear out the array of selected PDLs?

mvolkmann (approved):

web-ui/src/pages/CheckinsReportPage.jsx
vhscom: For `newPdls` to become empty I believe we would have a data issue on our hands. This change is to allow us to short-circuit the logic for changing component state as the default state when no pdls are available is controlled in a `useEffect` queuing off `processedQPs.current`.

mvolkmann: I see what you're saying. This looks good. I actually didn't noticed that this was in code related to the query parameters. Makes sense now.

mvolkmann (approved):

montesmoci (approved):

PR #2362, opened 2024-05-13
fix: include planned with in-progress

vhscom:
Ensure we're capturing planned check-ins when reporting quarterly check-in status by allowing a simple grace period. We also now filter out member who do not have a direct current relationship to the PDL and display the quarter number.

Screenshot 2024-05-13 at 1 42 30 PM

### DIFF
--- a/web-ui/src/helpers/datetime.js
+++ b/web-ui/src/helpers/datetime.js
@@ -9,3 +9,13 @@ export const getQuarterBeginEnd = inputDate => ({
   startOfQuarter: startOfQuarter(inputDate),
   endOfQuarter: endOfQuarter(inputDate)
 });
+
+/**
+ * Return the current quarter number with year.
+ * @param {Date} date - The date to get the quarter number for.
+ * @returns {string} The quarter number with year.
+ */
+export const getQuarterDisplay = date => {
+  const quarter = Math.floor((date.getMonth() + 3) / 3);
+  return `Q${quarter} ${date.getFullYear()}`;
+};
--- a/web-ui/src/helpers/datetime.js
+++ b/web-ui/src/helpers/datetime.js
@@ -1,4 +1,4 @@
-import { startOfQuarter, endOfQuarter } from 'date-fns';
+import { startOfQuarter, endOfQuarter, getQuarter } from 'date-fns';
 
 /**
  * Returns the start and end dates of the quarter that the given date falls in.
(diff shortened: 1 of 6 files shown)


web-ui/src/helpers/datetime.js
```diff
@@ -9,3 +9,13 @@ export const getQuarterBeginEnd = inputDate => ({
...
   endOfQuarter: endOfQuarter(inputDate)
 });
+
+/**
+ * Return the current quarter number with year.
+ * @param {Date} date - The date to get the quarter number for.
+ * @returns {string} The quarter number with year.
+ */
+export const getQuarterDisplay = date => {
+  const quarter = Math.floor((date.getMonth() + 3) / 3);
```
mvolkmann, line 19: The Check-ins project already has a dependency on the date-fns library.
It provides several functions for working with quarters. Maybe those should be used.
See https://date-fns.org/v3.6.0/docs/getQuarter.

web-ui/src/pages/CheckinsReportPage.jsx
```diff
@@ -9,11 +9,11 @@ import {
 
 import {
   Grid,
-  Typography,
```
mvolkmann, line 12: Mark loves alphabetizing!

web-ui/src/pages/CheckinsReportPage.jsx
```diff
@@ -120,12 +121,16 @@ const CheckinsReportPage = () => {
...
 
-  // Set the mapped PDLs to the PDLs with members
+  // Set the mapped PDLs to the PDLs with members attached
+  // filtering out data about check-ins under a different PDL.
   useEffect(() => {
     if (!pdls) return;
-    pdls.forEach(
-      pdl => (pdl.members = selectTeamMembersWithCheckinPDL(state, pdl.id))
-    );
+    pdls.forEach(pdl => {
```
mvolkmann, line 128: Does the code in the function passed to `forEach` cause some kind of side-effect?
Maybe the calls to `selectTeamMembersWithCheckinPDL` are necessary, but I don't understand why the `filter` call is necessary or the `return` keyword.

web-ui/src/helpers/datetime.js
vhscom: Thanks, the `getQuarter` cleans the up the readability of this function.

mvolkmann (approved):

mkimberlin (approved):

PR #2369, opened 2024-05-13
test: adjust tests for new date logic

vhscom:
Forgot to adjust this suite while updating date logic in #2362. Ran tests locally during pull, just not these.


mvolkmann (approved):

jackkeller (approved):
Approved

PR #2374, opened 2024-05-14
Added review permissions to load data script

ZacharyKlein:


mjperry91 (commented):
Might make sense to at least give Michael's account these privileges for testing. I dunno about everyone else but I've just been using that account locally since it's an admin.

ZacharyKlein:
> Might make sense to at least give Michael's account these privileges for testing. I dunno about everyone else but I've just been using that account locally since it's an admin.

My expectation is that since Michael is an admin, he will automatically receive all privileges associated with the admin role.

mjperry91 (approved):

vhscom (approved):

mkimberlin (approved):

PR #2375, opened 2024-05-14
Bugfix 2212/rewire report display

vhscom:
As discovered the previous page selector used in the Check-Ins was filtering out members without Check-Ins activity. Existing report logic is now replaced as expected per ACs in #2212 to report the expected PDL and team members.

Also updates our data load script to provide additional check-ins info to test reporting edge cases.

Finally, adds a keyboard listener to the Check-Ins report page for navigating quarters. This is especially helpful while making comparisons between quarters when the user's viewport has scrolled below the fold.

### DIFF
--- a/web-ui/src/components/reports-section/checkin-report/CheckinReport.jsx
+++ b/web-ui/src/components/reports-section/checkin-report/CheckinReport.jsx
@@ -8,7 +8,10 @@ import { selectFilteredCheckinsForTeamMemberAndPDL } from '../../../context/sele
 import ExpandMore from '../../expand-more/ExpandMore.jsx';
 import HorizontalLinearStepper from './HorizontalLinearStepper.jsx';
 import TeamMemberMap from './TeamMemberMap.jsx';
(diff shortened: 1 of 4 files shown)


jackkeller (approved):
Approved

web-ui/src/components/reports-section/checkin-report/CheckinReport.jsx
```diff
@@ -36,68 +39,51 @@ const CheckinsReport = ({ closed, pdl, planned, reportDate }) => {
...
-        planned
-      );
-      checkins = checkins.filter(checkin =>
-        isCheckinInProgress(checkin, startOfQuarter, endOfQuarter)
-      );
-      return checkins.length > 0;
-    };
-
-    const anyInProgress = members.some(isInProgress);
+    const anyInProgress = members.some(
```
mvolkmann, line 69: I love use of the `some` and `every` `Array` methods!

mvolkmann (approved):

PR #2377, opened 2024-05-14
fix: update in progress status logic

vhscom:
Fixes a bug where completed check-ins during a period where not being considered as part of in-progress indication.


mvolkmann (approved):

jackkeller (approved):
Approved

PR #2397, opened 2024-05-17
[2339] Reducing test run times by preventing migration after each test

mjperry91:
This commit will ensure we only run flyway migration once, and after each test we clear all entities from the db instead of cleaning it completely. As of now this takes our server test runtime from about 12 mins to a little under 2 mins.


sdelamo (approved):
We should not need to truncate every table after each test. The long-term solution would be to identify which tests leak side effects and fix those. 

However, I think we can merge this PR since it reduces test execution time by half: 

Without this PR: 14m https://gradle.com/s/c2hz3kekpipro
With this PR: 4m https://gradle.com/s/piiulfb27odqw


Moreover, I would delete the variable `shouldResetDBAfterEachTest` and both constructors.

mkimberlin (approved):

PR #2401, opened 2024-05-20
fix: use fixed-length reporting grace period

vhscom:
Depends on: https://github.com/objectcomputing/check-ins/pull/2402 ✅ 

The grace period logic used considered a date of May 1st in the Q1 check-in logic. This was caused due to unexpected behavior adding a `1 month` grace period. To address this issue, we instead use 30 days as the grace period.

Adds a test case which expresses the bug:
Q1 check-in date: Feb 12
Q2 check-in date: May 1

Before this update `May 1` was being selected as the check-in for both the Q1 and Q2 reporting periods.
After this update `May 1` is no longer selected as the check-in for the Q1 reporting period (Feb 12 is, as expected).

### DIFF
--- a/web-ui/src/components/reports-section/checkin-report/checkin-utils.test.js
+++ b/web-ui/src/components/reports-section/checkin-report/checkin-utils.test.js
@@ -65,57 +65,92 @@ describe('getLastCheckinDate', () => {
 });
 
 describe('getCheckinDateForPeriod', () => {
-  const mockReportDate = new Date(2024, 3, 15); // April 15, 2024 (example report date)
+  const mockReportDateQ1 = new Date(2024, 2, 31); // March 31, 2024 (end of Q1)
+  const mockReportDateQ2 = new Date(2024, 5, 30); // June 30, 2024 (end of Q2)
 
-  test('returns correct date when check-in is within the reporting period', () => {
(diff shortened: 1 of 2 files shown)


web-ui/src/components/reports-section/checkin-report/checkin-utils.test.js
```diff
@@ -65,58 +65,93 @@ describe('getLastCheckinDate', () => {
...
+
+    expect(result).toBeInstanceOf(Date);
+    expect(result.getFullYear()).toBe(2024);
+    expect(result.getMonth()).toBe(5); // June is 5 (zero-indexed)
+    expect(result.getDate()).toBe(30); // Latest date in the array
+    expect(result.getHours()).toBe(14);
+    expect(result.getMinutes()).toBe(30);
+  });
+
+  test('returns correct date when check-ins are on Feb 12 and May 1 with a report date in Q1', () => {
```
vhscom, line 139: Under previous logic this test would return 5/1 as the Q1 check-in date unexpectedly.

vhscom:
Note test failures currently affecting CI https://github.com/objectcomputing/check-ins/actions/runs/9159926663/job/25181447739

mjperry91 (approved):

Issue #2372, opened 2024-05-13
(Backend) - Refactor Existing Blocking Controllers to Non-Blocking Using Project Reactor


mjperry91:
- We have multiple pre-existing controllers that are implemented in a blocking manner. This story will update those controllers to be non-blocking inline with the vast majority of other check-in controllers.

Blocking Controllers
-  [x] ActionItemController
-  [ ] CheckinNoteController
-  [ ] GuildMemberController
-  [ ] CurrentUserController
-  [ ] MemberRoleController
-  [ ] TeamMemberController

sdelamo:
I think we can closed this via https://github.com/objectcomputing/check-ins/pull/2418

mjperry91:
> I think we can closed this via #2418

Agreed. I coordinated this with @S78901 this morning after seeing your PRs.

mjperry91:
We can move to jdk 21 and use virtual threads for these endpoints, which will give us non-blocking functionality without having to use reactor. This will be address in another issue.

PR #2404, opened 2024-05-20
#2372 Changes

S78901:
In relation to:
https://github.com/objectcomputing/check-ins/issues/2372


mjperry91:
@S78901 Sergio has brought up a good point that It might make more sense to just go ahead and start using virtual threads instead of project reactor. Especially for these endpoints that don't already use it. We can tag up this morning and go over it.

Issue #2355, opened 2024-05-10
Create a service for manipulating settings values


mkimberlin:
**As a System Administrator**, I want REST services to manage setting values, **so that** I can configure the application's behavior through the API.

## Acceptance Criteria:
 - [x] A new table named "settings" should be created in the database schema.
 - [x] The "settings" table should have these columns:
   - [x] `name` (VARCHAR, PRIMARY KEY, UNIQUE): Name of the setting, corresponding to an option contained in the settings enum.
   - [x] `value` (TEXT): The current value for the setting.
 - [x] A migration script should be created to alter the database schema and add the "settings" table.
 - [x] The following services should be implemented:
   - [x] A GET endpoint at `/services/settings` should retrieve all setting name-value pairs from the database.
   - [x] A PUT endpoint at `/services/settings/{name}` should allow updating the value for a specific setting.
     - [x] The request body should contain the new value as JSON data.
 - [x] Error handling should be implemented to return appropriate status codes for successful and unsuccessful requests.
   - [x] Attempting to store values that do not have an associated value in the settings enum should result in an bad request error.
   - [x] A new permission called `CAN_ADMINISTER_SETTINGS` should be created and used for protecting both of these new endpoints.

jackkeller:
@mkimberlin is there an idea of the different settings that we'll want to have control over? Mainly asking to anticipate what we'll be needing on the #2356 side of things.

mkimberlin:
I think there are a few categories of settings we will likely see out of the gate; one could be theming-related settings potentially (logo, primary/secondary/tertiary colors, etc.), another could be enabling and configuring various integrations (mailjet for example), and there might be some other values from our `run.sh` script variables that would be better served as admin configurable.

mjperry91:
@mkimberlin since the PUT endpoint requires us to pass a valid setting object to it, I have left this endpoint as `/services/settings` instead of adding the name path variable as we can just get that from the object and also verify the ID.

mjperry91:
PR: https://github.com/objectcomputing/check-ins/pull/2405

PR #2405, opened 2024-05-20
[2354 & 2355] Updating Existing Setting Controller

mjperry91:
This PR covers both [2354](https://github.com/objectcomputing.com/check-ins/pull/2354) and [#2355]((https://github.com/objectcomputing.com/check-ins/pull/2355) as they were closely related, and [#2355]((https://github.com/objectcomputing.com/check-ins/pull/2355) had already been partially implemented a while back.

-Adding options require name to be valid from the enum
-Removing userId from Setting table as we don't want to tie these to specific users currently
-Adding tests


server/src/main/java/com/objectcomputing/checkins/services/settings/SettingOption.java
```diff
@@ -0,0 +1,57 @@
...
+        for (SettingOption option : values()) {
+            if (option.name().equalsIgnoreCase(name)) {
+                return option;
+            }
+        }
+        throw new UnsupportedOperationException(String.format("Unknown permission: '%s'", name));
+    }
+
+    public enum Category {
+        LOGO
```
mjperry91, line 51: @mkimberlin @jackkeller  This is essentially a filler value for now. Let me know if yall have any ideas as to what you would expect here for an initial take.

server/src/main/java/com/objectcomputing/checkins/services/settings/SettingsServicesImpl.java
```diff
@@ -1,53 +1,51 @@
...
+        if ((settingsRepository.existsByIdAndName(setting.getId(), setting.getName()))) {
             return settingsRepository.update(setting);
         } else {
             throw new BadArgException(String.format("Setting %s does not exist, cannot update", setting.getId()));
         }
     }
 
-    public List<SettingsResponseDTO> findByName(String name) {
-        List<Setting> searchResult = settingsRepository.findByUserId(currentUserServices.getCurrentUser().getId());
+    public List<SettingsResponseDTO> findByName(@Nullable String name) {
```
ZacharyKlein, line 40: I don't know how large of a dataset we expect in this table, but I wonder if it would make sense to test for `name != null` and then either query all the settings or only those with a matching name, based on that result?

ZacharyKlein (approved):
LGTM, made one comment.

sdelamo (dismissed):
I created a PR with improvements: 

https://github.com/objectcomputing/check-ins/pull/2406

server/src/main/java/com/objectcomputing/checkins/services/settings/SettingOption.java
mkimberlin: `Theme`, `Integrations`, `Check-Ins`, `Reviews`

I have some thoughts surrounding settings in each of those areas (and probably more).

mkimberlin (approved):

PR #2415, opened 2024-05-21
(Backend) - Update Google Dependencies to Address Vulnerabilities and General Tech Debt

timyates:
The original plan was to perform a minor upgrade of the Google dependencies, however the latest `postgres-socket-factory` pulls in a transitive major update to v2 of the `google-api-client`

So I went for it and grabbed the latest major version of all Google dependencies.

Also updated:

- Patch Micronaut upgrade to 4.4.2
- Micronaut Gradle plugin patch update
- Selenium to latest
- Latest webdrivermanager
- Latest openrewrite library

Some Google classes have changed package, but I don't believe these are persisted (internal use only), so I think this is ok

### DIFF
--- a/server/src/main/java/com/objectcomputing/checkins/services/file/GoogleDriveAccessor.java
+++ b/server/src/main/java/com/objectcomputing/checkins/services/file/GoogleDriveAccessor.java
@@ -4,7 +4,7 @@ import com.google.api.client.googleapis.javanet.GoogleNetHttpTransport;
 import com.google.api.client.http.HttpRequestInitializer;
 import com.google.api.client.http.javanet.NetHttpTransport;
 import com.google.api.client.json.JsonFactory;
-import com.google.api.client.json.jackson2.JacksonFactory;
+import com.google.api.client.json.gson.GsonFactory;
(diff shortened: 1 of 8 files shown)


mjperry91 (approved):

server/src/main/java/com/objectcomputing/checkins/services/file/GoogleDriveAccessor.java
```diff
@@ -16,7 +16,7 @@
 @Singleton
 public class GoogleDriveAccessor {
 
-    private static final JsonFactory JSON_FACTORY = JacksonFactory.getDefaultInstance();
+    private static final JsonFactory JSON_FACTORY = GsonFactory.getDefaultInstance();
```
mkimberlin, line 19: Any particular driver for this change? Just curious.

timyates: They removed JacksonFactory in favor of GsonFactory 

https://cloud.google.com/java/docs/reference/google-http-client/latest/com.google.api.client.json.jackson2.JacksonFactory

PR #2426, opened 2024-05-22
Restrict review period status transitions

timyates:
Prior to updating a ReviewPeriod, we now validate that the ReviewStatus follows the following rules:

- State can remain the same
- `PLANNING` can progress to `AWAITING_APROVAL`
- `AWAITING_APPROVAL` can progress to `OPEN`
- `OPEN` can progress to `CLOSED`
- `CLOSED` can progress to `OPEN` (for when a case is re-opened)
- `UNKNOWN` can progress to `PLANNING` (`UNKNOWN` was used for period where isOpen was NULL prior to migration)

All other state changes will cause a `BAD_REQUEST` response with an appropriate message

### DIFF
--- a/server/build.gradle
+++ b/server/build.gradle
@@ -99,6 +99,7 @@ dependencies {
     testImplementation("org.testcontainers:postgresql")
     testImplementation("org.testcontainers:testcontainers")
     testImplementation("org.hamcrest:hamcrest")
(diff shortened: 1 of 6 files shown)


ZacharyKlein (approved):
LGTM, tagging @mkimberlin to validate that the state transitions are all correct.

mjperry91 (approved):

Issue #2350, opened 2024-05-09
Implement Pulse Survey report permissions


mkimberlin:
**As a System Administrator**, I want to define new permissions for accessing pulse survey data

## Acceptance Criteria:
 - [x] A new permission named "CAN_VIEW_ALL_PULSE_RESPONSES" should be created.
 - [x] The `PulseResponseController` (and related service) should be altered so that users with this permission are allowed to access pulse survey responses from all team members and others are not.
 - [x] The `PulseResponseController` (and related service) should be altered so that users without this permission are allowed to see their own responses and those of people in their reporting heirarchy.
 - [x] Any existing pulse survey APIs should be reviewed and appropriate permissions added to ensure proper access control

mkimberlin:
This story may need some additional grooming to define the needed permissions.

timyates:
So I have implemented the described permission and added the hierarchy check for reading pulse reports.

This has been done in the Service layer...but I see now that the requirement says it should also be checked in the controller...do we need to check it twice?

Do we have an idea of other permissions we require?  Currently I believe anyone authenticated could create a pulse response with another user's TeamMemberId. I guess we want to prevent this?

My current thoughts are:

1. New permissions `CAN_CREATE_ALL_PULSE_RESPONSES`, `CAN_UPDATE_ALL_PULSE_RESPONSES`, `CAN_DELETE_ALL_PULSE_RESPONSES`
2. Checks when creating, deleting or updating to ensure that the permission is in the current user's role, or the `teamMemberId` field is set to that of the current user, or one of the users in the current user's hierarchy

Do these make sense?

PR #2429, opened 2024-05-22
Implement Pulse Survey report permissions

timyates:
See #2350

Added 1 new permission:

- CAN_VIEW_ALL_PULSE_RESPONSES

create and update check

1. If not, are they the user in the teamMemberId field for the report
2. If not, does the `teamMemberId` in the supervisory tree of the current user?

And view checks the permission exists, and if not then performs the 2 above checks


timyates:
[I removed the service test](https://github.com/objectcomputing/check-ins/pull/2429/commits/3a61986eb425a62f7d2c1449ec5d30e1facdbcbb), the coverage is almost exactly the same, and the mocking was getting excessive.  The controller test does a better job of exercising the paths with real services and actual data (IMO)

mjperry91 (approved):

server/src/main/java/com/objectcomputing/checkins/services/memberprofile/MemberProfileServicesImpl.java
```diff
@@ -150,4 +150,15 @@ public List<MemberProfile> getSupervisorsForId(UUID id) {
         return supervisorsForId;
     }
 
+    @Override
+    @Cacheable
+    public List<MemberProfile> getSubordinatesForId(UUID id) {
+        List<MemberProfile> subordinatesForId = memberProfileRepository.findSubordinatesForId(id);
+        if (!currentUserServices.isAdmin()) {
+            for (MemberProfile memberProfile : subordinatesForId) {
+                memberProfile.clearBirthYear();
```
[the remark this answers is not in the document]
mkimberlin: We don't want to expose folks ages to everyone, so we clear them for those who don't need to know.

server/src/main/java/com/objectcomputing/checkins/services/pulseresponse/PulseResponseServicesImpl.java
```diff
@@ -2,81 +2,123 @@
...
     }
-    
+
     @Override
     public PulseResponse save(PulseResponse pulseResponse) {
+        UUID currentUserId = currentUserServices.getCurrentUser().getId();
         PulseResponse pulseResponseRet = null;
-        if(pulseResponse!=null){
+        if (pulseResponse != null) {
             final UUID memberId = pulseResponse.getTeamMemberId();
```
mkimberlin, line 45: Given my comment about only being able to create/update your own responses. It might be worthwhile to turn the `teammemberid` field into one that is annotated with and populated via `@CreatedBy`. Thoughts?

mkimberlin (approved):

server/src/main/java/com/objectcomputing/checkins/services/pulseresponse/PulseResponseServicesImpl.java
timyates: I had a look, but I am unsure as to how to get the test working with `@CreatedBy` 🤔 

Should I change

```
            } else if (!currentUserId.equals(memberId) && !isSubordinateTo(memberId, currentUserId)) {
```

in save and update to

```
            } else if (!currentUserId.equals(memberId)) {
```

To more closely match your comment?  We can then raise a card to investigate `@CreatedBy` and I can pick it up when I return on the 3rd?

mkimberlin: That works, yes. Sorry for the slow reply. I missed yours completely.

PR #2440, opened 2024-05-24
Allow Ryuk container when testing with Docker Desktop

### LINKED ISSUE #2439: Allow Ryuk container when testing with Docker Desktop
The project is configured to use Podman for container support, and disabled the TestContainers Ryuk container when running tests.

https://github.com/objectcomputing/check-ins/blob/fe118c60b10bd466f827e87a74768cd6b90d2a47/server/build.gradle#L106-L115

If instead you are using Docker Desktop, this causes containers to hang around running and eating resources when running tests or debugging sessions.

We could detect an environment variable which would signify the use of Docker Desktop and skip the podman configuration for the test tasks, then people who want to use Docker Desktop would simply export this env var to re-enable ryuk and skip the podman configuration.

timyates:
Closes #2439


mjperry91 (approved):

docs/getting-started/setup/macos.md
```diff
@@ -267,6 +267,9 @@ You should be using Yarn 1.22.19 or later.
 
 # Install Podman
 
+> If you are using Docker Desktop, you can skip this section, and set an environment variable via `export USE_DOCKER_DESKTOP=1` in your shell profile.
+> This will enable Ryuk for cleaning up old containers under testing.
+
```
mkimberlin, line 272: This seems to affect all environments, so should we add something to the `windowsos.md` as well?

timyates: Good catch! Pushed https://github.com/objectcomputing/check-ins/pull/2440/commits/7102c02f938a1bc5d3d8f4d26b8d9eda2106875b

Issue #2237, opened 2024-04-24
FindByFields in Check-Ins Services have unused branch of code


Samibanley:
There are 4 different service implementations that have FindByFields methods that are searching by `createbyid`. After looking in the web-ui, these particular branches of code are never being used. However, if they were used, they would not allow a current user to search by `createbyid` if they are a PDL participant in the check-in. This is a problem, because it goes against the general viewing rules of check-ins.

If this code is never intended to be used, it should be removed. If it might be used in the future, there should be unit tests for those methods that explicitly test the logic of those methods. Right now, they are only being tested via the controller. 

The methods in question:
CheckinNoteServicesImpl.java > findByFields
AgendaItemServicesImpl.java > findByFields
PrivateNoteServicesImpl.java > findByFields
ActionItemCRUDValidator.java > validatePermissionsFindByFields

PR #2452, opened 2024-05-28
Feature 2237 clean up FindByFields

mjperry91:
closes: https://github.com/objectcomputing/check-ins/issues/2237


ZacharyKlein (approved):

mjperry91:
@mkimberlin the createdById search parameter for these GET endpoints is in fact not used by the UI as the story mentions. This PR just cleans up the code as the logic was difficult to follow. If you think this parameter should not be maintained and can easily remove it as well, but for now preserving the functionality encase we want it some day.

mkimberlin (approved):

Issue #2456, opened 2024-05-29
TestContainer Issue


mjperry91:
When running all tests on develop I encounter some test container errors. Should be noted I do not have a local database running.

```
12:08:51.196 [Test worker] ERROR tc.postgres:11.6 - Could not start container
java.lang.RuntimeException: com.github.dockerjava.zerodep.shaded.org.apache.hc.core5.http.NoHttpResponseException: localhost:2375 failed to respond
```

```
Bean definition [javax.sql.DataSource] could not be loaded: Error instantiating bean of type  [javax.sql.DataSource]

Message: Failed to initialize pool: Container startup failed for image postgres:11.6
Path Taken: DatasourceFactory.dataSource(DatasourceConfiguration datasourceConfiguration)
io.micronaut.context.exceptions.BeanInstantiationException: Bean definition [javax.sql.DataSource] could not be loaded: Error instantiating bean of type  [javax.sql.DataSource]

Message: Failed to initialize pool: Container startup failed for image postgres:11.6
Path Taken: DatasourceFactory.dataSource(DatasourceConfiguration datasourceConfiguration)
	at io.micronaut.context.DefaultBeanContext.initializeContext(DefaultBeanContext.java:1982)
	at io.micronaut.context.DefaultApplicationContext.initializeContext(DefaultApplicationContext.java:294)
	at io.micronaut.context.DefaultBeanContext.readAllBeanDefinitionClasses(DefaultBeanContext.java:3348)
	at io.micronaut.context.DefaultBeanContext.finalizeConfiguration(DefaultBeanContext.java:3692)
	at io.micronaut.context.DefaultBeanContext.start(DefaultBeanContext.java:342)
	at io.micronaut.context.DefaultApplicationContext.start(DefaultApplicationContext.java:202)
	at io.micronaut.test.extensions.AbstractMicronautExtension.startApplicationContext(AbstractMicronautExtension.java:507)
	at io.micronaut.test.extensions.AbstractMicronautExtension.beforeClass(AbstractMicronautExtension.java:346)
	at io.micronaut.test.extensions.junit5.MicronautJunit5Extension.beforeAll(MicronautJunit5Extension.java:84)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
Caused by: io.micronaut.context.exceptions.BeanInstantiationException: Error instantiating bean of type  [javax.sql.DataSource]


```

PR #2457, opened 2024-05-29
[2237] Test Containers Fail To Start For Classes Not Using The TestContainers On Windows

mjperry91:
This may just be a windows issue as it works fine on the CI. One thing I did notice was adding back the @TestContainer and @Container annotations also fixed this issue (as was done in this [draft](https://github.com/objectcomputing/check-ins/pull/2458)). Though despite this causing the tests run faster on my local machine (-40seconds) they become significantly slower on the CI (+3 mins). So this seems to be the preferable solution.

Closes: https://github.com/objectcomputing/check-ins/issues/2456


timyates (approved):
@mjperry91 Cool! I wonder why it wasn't picking up the testcontainers driver and url from config 🤔 

I *really* hope you don't mind, but I pushed some cleanup commits (stop using public on tests, move away from deprecated initMocks and close the handle returned by openMocks)

I also removed the testcontainers datasource config from application-test.yml as I don't think it's used any more 🤔

mkimberlin (approved):

PR #2458, opened 2024-05-29
[2237] Test Containers Fail To Start For Classes Not Using The TestContainersSuite On Windows

mjperry91:
This [PR ](https://github.com/objectcomputing/check-ins/pull/2457) seems to be the preferable solution as it runs slower on my local machine, but significantly faster on the CI. Will leave this draft up for a day or two as a reference.

Closes: https://github.com/objectcomputing/check-ins/issues/2456

### DIFF
--- a/server/src/test/java/com/objectcomputing/checkins/services/TestContainersSuite.java
+++ b/server/src/test/java/com/objectcomputing/checkins/services/TestContainersSuite.java
@@ -1,7 +1,6 @@
 package com.objectcomputing.checkins.services;
 
 import com.objectcomputing.checkins.services.fixture.RepositoryFixture;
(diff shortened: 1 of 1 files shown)


server/src/test/java/com/objectcomputing/checkins/services/TestContainersSuite.java
```diff
@@ -32,6 +32,46 @@ public abstract class TestContainersSuite implements RepositoryFixture, TestProp
 
     public TestContainersSuite() {}
 
+    @BeforeEach
```
mjperry91, line 35: Nothing changed other than the annotations being added back. The diff just shows this because I moved the deleteAllEntities method to the bottom of the class

PR #2462, opened 2024-06-03
Upgrade to Gradle 8.8 and cleanup build

### LINKED ISSUE #2461: Clean up server build
We currently define configurations in the server build file which are no longer used

We also fetch the yarn-build output by explicitly naming the task that generates the output.  This isn't the recommended way any more, and we should use configurations with attributes to model the dependencies

https://docs.gradle.org/current/userguide/cross_project_publications.html#sec:variant-aware-sharing

timyates:
This PR upgrades Gradle to 8.8 and performs some cleanup, namely:

1. Removes unused configurations
2. [Follows current best practices](https://docs.gradle.org/current/userguide/cross_project_publications.html#sec:variant-aware-sharing) to depend on the web-ui yarnBuild output
3. Fixes the publication task (I think)... At least `./gradlew publishToMavenLocal` seems correct now 🤔 
4. Simplifies the main build file by removing subproject configuration
5. Removes the idea plugin which I don't believe is required any more (as import works)
6. The sonar plugin didn't support Gradle 8, so this PR updates it to latest, and configures it to work.

### Running the sonar plugin to test it

- Run sonar locally with `docker run -d --name sonarqube -p 9000:9000 sonarqube` then visit localhost:9000
- When it's running log in as an admin (with admin/admin) and then change your password as requested
- Create a local project and give it a name and a key `check-ins`
- On the next screen, select "Use the global setting" and click "Create project"
- Select "Locally" for the analysis method
- Click "Generate" to create a token, then click "continue"
- Select "gradle", copy the command, and run it locally in a terminal window

Closes #2461


timyates:
Moving back to draft as I think I have broken sonarqube

mkimberlin:
Could I con you into adding something to the docs site for this?

mkimberlin (approved):
The real test will be when the publish github actions workflow gets run. 👀

timyates:
> The real test will be when the publish github actions workflow gets run. 👀

I don't think it was working before 😬 The 0.7.0 jar file just contains a MANIFEST.MF and nothing else 😢 

https://github.com/objectcomputing/check-ins/packages/275981

timyates:
@mkimberlin Docs added!  Checked locally, and they look ok when rendered 😎 

image

mkimberlin:
> > The real test will be when the publish github actions workflow gets run. 👀
> 
> I don't think it was working before 😬 The 0.7.0 jar file just contains a MANIFEST.MF and nothing else 😢
> 
> https://github.com/objectcomputing/check-ins/packages/275981

🤣  oops.

mjperry91 (approved):

vhscom:
Docs were very helpful for getting SonarQube set-up locally.
Screenshot 2024-06-04 at 11 47 16 AM

PR #2463, opened 2024-06-03
Fix assertion with primitive and null

timyates:
Whilst getting sonar working I spotted that we had warnings about comparing a primitive to null in the tests

The lines looked like

```
assertNotNull(response.getContentLength());
```

Which will always pass.

Also saw some places where we had missed the assertion

```
response.equals(opportunitiesResponse);
```

Which I replaced with

```
assertEquals(Set.of(opportunitiesResponse), response.body());
```

### DIFF
--- a/server/src/test/java/com/objectcomputing/checkins/services/opportunities/OpportunitiesControllerTest.java
+++ b/server/src/test/java/com/objectcomputing/checkins/services/opportunities/OpportunitiesControllerTest.java
@@ -22,13 +22,13 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
 import java.util.UUID;
-import java.util.stream.Collectors;
 
 import static com.objectcomputing.checkins.services.role.RoleType.Constants.ADMIN_ROLE;
 import static com.objectcomputing.checkins.services.role.RoleType.Constants.MEMBER_ROLE;
 import static org.junit.jupiter.api.Assertions.assertEquals;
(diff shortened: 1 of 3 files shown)


server/src/test/java/com/objectcomputing/checkins/services/opportunities/OpportunitiesControllerTest.java
```diff
@@ -82,144 +82,126 @@ void testMemberCreateAnOpportunities(){
...
-        final HttpRequest<Object> request = HttpRequest.GET("/").basicAuth(ADMIN_ROLE,ADMIN_ROLE);
-
+        final HttpRequest<Object> request = HttpRequest.GET("/").basicAuth(ADMIN_ROLE, ADMIN_ROLE);
         final HttpResponse<Set<Opportunities>> response = client.toBlocking().exchange(request, Argument.setOf(Opportunities.class));
 
         assertEquals(HttpStatus.OK, response.getStatus());
-        assertNotNull(response.getContentLength());
-        response.equals(opportunitiesResponse);
+        assertTrue(response.getContentLength() > 0, "response.getContentLength() > 0");
+        assertEquals(Set.of(opportunitiesResponse), response.body());
```
timyates, line 174: This is the thing that I changed here (there are 6 instances of the null check)

The rest is formatting... I wish git diff understood formatting a bit better... it never feels big enough to warrant a separate PR, but when put in here it makes it really hard to see what's going on 😿

mjperry91 (approved):

PR #2479, opened 2024-06-05
Fix PulseReportPage test

timyates:
When under test, use a fixed date for the page so that snapshots are consistent


jackkeller (approved):
Makes sense, still not wholly a fan of snapshot tests but this is a good way to handle changing dates to keep the CI happy.

mvolkmann (approved):

PR #2487, opened 2024-06-06
Feature 2363/email refactor

chrismckiernan:


vhscom (approved):

server/src/main/java/com/objectcomputing/checkins/services/guild/GuildServicesImpl.java
```diff
@@ -93,7 +96,18 @@ public GuildResponseDTO save(GuildCreateDTO guildDTO) {
...
         }
 
-        return fromEntity(newGuildEntity, newMembers);
+        GuildResponseDTO guildResponse = fromEntity(newGuildEntity, newMembers);
+
+        Set<String> emailsOfNewGuildLeads = newMembers
+                .stream()
+                .filter(GuildMemberResponseDTO::isLead)
+                .map(lead -> memberProfileServices.getById(lead.getMemberId()).getWorkEmail())
+                .collect(Collectors.toSet());
```
timyates, line 105: We put these into a set, but AFAICS we don't override hashcode and equals in the GuildMemberResponseDTO, so we might end up with duplicates?

chrismckiernan: We are collecting the emails in to as et so that would just be a set of strings in this case and any duplicate emails would be kicked out.

timyates: Ignore me...

I've got Friday afternoon syndrome

PR #2491, opened 2024-06-07
Feature 2448 volunteer activities

mvolkmann:


web-ui/src/components/certifications/CertificationBadges.jsx
```diff
@@ -0,0 +1,51 @@
+import PropTypes from 'prop-types';
```
mvolkmann, line 1: This displays certifications matches on member profile pages.

mvolkmann:
I need to wait to create this PR until the one for certifications is merged. Otherwise the changes for that one and this one will be in the same PR.

PR #2495, opened 2024-06-12
Fix permissions for updating earned certifications

timyates:
Previously we checked the owner of the passed in object, but not the owner of the db object in the case of an update.

This meant that people could effectively change ownership of earned certifications from someone else to themselves.

This commit makes the change so we check both the object in flight, and the object at rest for updates.


server/src/main/java/com/objectcomputing/checkins/services/certification/EarnedCertificationRepository.java
timyates, line 1: I also changed the join from `LEFT JOIN` to `JOIN`, this is a no-op I believe as there cannot be one without the other, however `JOIN` is more correct

mkimberlin (approved):

vhscom (approved):

PR #2499, opened 2024-06-13
Remove the Mono.xxx wrapping from the controllers

timyates:
We have many places where we are wrapping the responses in a Mono in the controllers.

As we declare `@ExecuteOn(TaskExecutors.BLOCKING)` in these controllers, I don't think this is required.

And it probably affects performance, as the work is probably offloaded multiple times to multiple thread pools.

This commit removes (almost) all of the wrapping. The only place it remains is in the Github client (as I'm not currently sure where that's used) 🤔


mkimberlin (approved):

vhscom (approved):
Looks great. This is going to make the codebase much easier to work with.

PR #2502, opened 2024-06-13
Some sonar cleanup (picking the most prevalent or highest priority)

timyates:
Best reviewed commit by commit as I've **tried** to be strong and just fix one thing per commit...

Apologies for when I may have strayed 😢 

This clears up a coupld of hundred Sonar errors and warnings

---

1. [JUnit5 class and method visibility](https://github.com/objectcomputing/check-ins/commit/95af5734022c29a1897b81e79336a7009165a3b0)

    In JUnit 5 it is recommendend to have package-private classes and methods for tests, Sonar flags this as an issue

2. [Logger should be a private static field](https://github.com/objectcomputing/check-ins/commit/69e7fc49bedd59ac30f17a72d4edb6f5c650d7f3)

    Sonar flags this as a naming vilation (as it's all in caps), I think it just needed to be `private final static`

3. [Replace `Collectors.toList()` as appropriate](https://github.com/objectcomputing/check-ins/commit/2e655b34ea934db030f8aee0bec83cf31d39499f)

    Instead of `stream.collect(Collectors.toList())`, it is (since Java 16) recommended to use `stream.toList()`.
    Care must be taken though, as the returned list is then unmodifiable (which is different to the old `Collectors` version)

4. [assertEquals argument order](https://github.com/objectcomputing/check-ins/commit/7df30942d80f06f786342e108168a22cbf7be5f4)

    AssertEquals in tests expects the first argument to be the expected value.
    Having it the wrong way round causes the output to be wrong, and can cause delays in debugging issues.

5. [accidental self assignement](https://github.com/objectcomputing/check-ins/commit/1218b7deee2869e89414fc5c4eb4d60b9ce5edd1)

    In 1 constructor, we were just setting the id to itself (as it wasn't passed as an argument)
    In the second (due to capitalization) we weren't using the passed value for `Id`.
    This wasn't an issue however as I don't believe the constructor is used 🤔 

6. [CSS: `deg` instead of `eg`](https://github.com/objectcomputing/check-ins/commit/53fc2940316bbe8ec96551deaf13161a75293cb5)

    Sonar caught a typo

7. [CSS: duplicated padding member](https://github.com/objectcomputing/check-ins/commit/b256d5df85fe3e0165b5b5dd5f52d05d1d3d195f)

    I don't believe the spec says what to do when it's declared twice (hence the warning in Sonar).
    I kept the second-most padding (which I believe is what most browsers do)

8. [CSS: invalid `box-size` parameter](https://github.com/objectcomputing/check-ins/commit/397492a3cf243a9d2498bea5ca5d8837e1ab7c6b)

    I believe this should be `box-sizing` and I've update it as such

9. [Remove unrequired Mockito `eq()` usage](https://github.com/objectcomputing/check-ins/pull/2502/commits/8759080aabce324e33ccfc62cd8ebde9abbdf001)

    This is not required and makes the tests harder to read

10. [Remove parentheses round single lambda parameters](https://github.com/objectcomputing/check-ins/pull/2502/commits/3ef25cb0f3c521f8053c6cad0d1e3b15eee0ff23)

### DIFF
--- a/server/src/main/java/com/objectcomputing/checkins/notifications/email/MailJetSender.java
+++ b/server/src/main/java/com/objectcomputing/checkins/notifications/email/MailJetSender.java
@@ -17,7 +17,6 @@ import org.slf4j.LoggerFactory;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
-import java.util.stream.Collectors;
 
 @Requires(property = MailJetSender.FROM_ADDRESS)
 @Requires(property = MailJetSender.FROM_NAME)
@@ -58,7 +57,7 @@ public class MailJetSender implements EmailSender {
             // Get only the first n elements limited by MailJet's API
             List<String> limitedRecipients = recipientList.stream()
                     .limit(MAILJET_RECIPIENT_LIMIT)
-                    .collect(Collectors.toList());
+                    .toList();
 
             // Add each recipient to a JSON array to be sent in a MailJet request
             JSONArray recipientArray = new JSONArray();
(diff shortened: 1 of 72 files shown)


vhscom (approved):
Nice clean-up. Re: CSS cascading order, [the spec says](https://www.w3.org/TR/CSS2/cascade.html#cascade):

> Finally, sort by order specified: if two declarations have the same weight, origin and specificity, the latter specified wins. Declarations in imported style sheets are considered to be before any declarations in the style sheet itself.

mkimberlin (approved):

PR #2506, opened 2024-06-17
Remove instance booleans and fix conditionals

timyates:
Where possible this commit switches from `Boolean` class parameters to primitive booleans.

The problem with `Boolean` is that it introduces 3 states, `TRUE`, `FALSE` and `null`.

Where the contract has a Nullable requirement, I have updated the `if` statements, as `if (obj.nullableMember()) {` will throw a null pointer exception if the Boolean result is `null`.  The safe replacement is `if (Boolean.TRUE.equals(obj.nullableMember())) {`


vhscom (approved):

mkimberlin (approved):
I'm not going to lie. This makes me a little sad. I have a personal hatred of primitives in Java.

timyates:
Unexpected NullPointerExceptions are worse imho 😉

PR #2515, opened 2024-06-18
Fixed file upload

mkimberlin:


vhscom (approved):

timyates (approved):
Is it possible to test this locally (without mocks)?  I know AWS has localstack for example which allows you to test S3 integration locally via docker

rodecapd (approved):

mkimberlin:
> Is it possible to test this locally (without mocks)? I know AWS has localstack for example which allows you to test S3 integration locally via docker

This should work locally. I tested it that way.

timyates:
> This should work locally. I tested it that way.

But all the tests were passing both before and after this change 🤔

PR #2517, opened 2024-06-19
Front-end import cleanup

timyates:
This PR clears up 80 warnings issued by Sonar around imports in the React code in 2 commits (might be easier to review each commit separately)

1. [Imports from the same module should be merged](https://github.com/objectcomputing/check-ins/commit/71906208d6693fd5a57823e2e02df6a6b7f11b0d)

    > Having the same module imported multiple times can affect code readability and maintainability.

2. [Unnecessary imports should be removed](https://github.com/objectcomputing/check-ins/commit/82df0520b93db75b286b9beb067fa4b9931c7a04)

    > These imports do not contribute to the functionality of the application and only add extra weight to the JavaScript bundle, leading to potential performance and maintainability issues.

### DIFF
--- a/web-ui/src/components/feedback_recipient_selector/FeedbackRecipientSelector.jsx
+++ b/web-ui/src/components/feedback_recipient_selector/FeedbackRecipientSelector.jsx
@@ -1,19 +1,20 @@
 import React, { useContext, useEffect, useState, useRef } from 'react';
 import { styled } from '@mui/material/styles';
-import './FeedbackRecipientSelector.css';
 import FeedbackRecipientCard from '../feedback_recipient_card/FeedbackRecipientCard';
 import { AppContext } from '../../context/AppContext';
 import {
   selectProfile,
   selectCsrfToken,
+  selectCurrentUser,
   selectNormalizedMembers
 } from '../../context/selectors';
 import { getFeedbackSuggestion } from '../../api/feedback';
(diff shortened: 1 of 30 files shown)


web-ui/src/components/feedback_recipient_selector/FeedbackRecipientSelector.jsx
```diff
@@ -1,20 +1,21 @@
 import React, { useContext, useEffect, useState, useRef } from 'react';
 import { styled } from '@mui/material/styles';
-import './FeedbackRecipientSelector.css';
```
mvolkmann, line 3: I don't think Sonar can really detect whether this import is used. We probably need to restore all CSS imports.

mvolkmann (approved):

web-ui/src/components/feedback_recipient_selector/FeedbackRecipientSelector.jsx
timyates: Yeah, I tried to arrange them all to kind of match each other... In the main the css import comes just prior to the const values

mvolkmann: That matches my preference.

vhscom (approved):

PR #2504, opened 2024-06-14
Addition of Kudos

mkimberlin:

PR #2523, opened 2024-06-24
fixed R__Load_testing_data.sql

mvolkmann:


vhscom:
Looks like these two lines are available in https://github.com/objectcomputing/check-ins/pull/2504/files#diff-2c49ee74aa139dcd59b6c80214bd483d40a22bc99268f24d198fbc41f7c8363bR20-R21, which is yet to be merged.

timyates (changes requested):
I don't think we need this until the Kudos pr is ready (it's already in the Kudos pr)

If the database changes in the pr, then this might end up being wrong won't it?

PR #2527, opened 2024-06-26
Use config class instead of `check-ins.web-address` and `check-ins.application.name`

timyates:
Currently we have a proliferation of `@Property(name = "check-ins.web-address") String webURL` for whenever we need to send an email to the user (or create an absolute URL link back to the server)

This PR replaces these with a single singleton configuration object that holds the web address and can be injected into the services as required.

This is the first step in moving all the `check-ins` configuration into code, to add type safety and reduce duplication of key names

[edit] I also added `check-ins.application.name` to the config object, and replaced it's usage in the code.


vhscom (approved):
LGTM

mkimberlin (approved):

PR #2530, opened 2024-06-27
Fix GoogleServiceConfiguration field names

timyates:
Before this PR;

1. We had non-idiomatic field names in `GoogleServiceConfiguration`
2. We were building JSON via toString()

When you create a service account with Google, they supply you with a JSON document which contains all the fields previously in `GoogleServiceConfiguration`.

Instead of requiring these to be copied out into separate fields for us to then re-create the JSON document on startup, this PR;

1. Switches to a single property. This is a Base64 encoded JSON document (as supplied by Google)
2. Adds a validator to ensure this value (1) is not null (2) is base64 encoded, and (3) represents a json document with the required keys

> [!WARNING]
> We will need to ensure that this `SERVICE_ACCOUNT_CREDENTIALS` environment variable exists where we deploy to.
> I have no visibility of PROD or DEV environments, so cannot confirm this.
>
> This environment variable needs to contain the Base64 encoded JSON document from google for the service account


server/src/main/java/com/objectcomputing/checkins/security/GoogleServiceConfiguration.java
```diff
@@ -1,171 +1,64 @@
...
 
+@Getter
+@Setter
 @ConfigurationProperties("service-account-credentials")
-@Introspected
 public class GoogleServiceConfiguration {
 
     @NotNull
-    public String directory_id;
+    @JsonProperty("directory_id")
```
mkimberlin, line 15: Not sure your plan for this PR after our discussion but, if you don't end up removing these completely, it's worth noting that you converted the `_`s to `-`s in the json.

mkimberlin (approved):
I'm going to go ahead and approve this. It's a positive step regardless of whether you end up eventually taking the step of removing these in favor of the encoded version.

server/src/main/java/com/objectcomputing/checkins/security/GoogleServiceConfiguration.java
timyates: In the *yaml* configuration I converted to use hyphens instead of underscores, but both will work for defining properties in Micronaut (but hyphen is more idiomatic)

In the generated JSON from this object, it will still use underscores (as the `JsonProperty` specifies underscores)

rodecapd (approved):
Discussed with Tim, he's goiing to change the ENV variable name and I'll pick it up in the terraform work.

mkimberlin (approved):

PR #2533, opened 2024-06-28
Graal support

### LINKED ISSUE #2532: Use GraalVM to create a native image
If we use GraalVM to build a native image of the server, we should be able to greatly reduce startup time, and prevent the occasional timeouts we see with a cold start.

We will use the latest GraalVM v21

- [ ] `./gradlew nativeCompile` creates a native image when using a GraalVM JDK
- [ ] `./gradlew nativeTest` runs all the tests against the nativeImage (to shake out errors)
- [ ] The native image is manually checked and is runnable
- [ ] Github actions to build native image are created

timyates:
Closes #2532 

## Instructions for local building

- To compile, install a GraalVM JDK for Java 21 (i.e. `sdk install java 21.0.3-graal`)
- Run `./gradlew nativeCompile`
- Create a `nativeRun.sh` script which is a copy of `run.sh` but with

    ```
    ./server/build/native/nativeCompile/check-ins
    ```

    In place of `./gradlew run` at the end

## Usage on Cloud-run

I propose for cloud-run we make use of `./gradlew dockerBuildNative`.
This PR has updated the Gradle file to configure the image name when executed via Github actions to be the same as the current process...
We should be able to just replace the `Build the Docker image` step in the actions with one to run this task.

## Tests

- When using the GraalVM JDK (above), you can run `./gradlew nativeTest` to run the tests against a native binary.
- 22 tests have been disabled when run this way via the `@DisabledInNativeImage` annotation
- This is because I cannot get the mocking working when run natively (not sure it's even possible)
- Raised https://github.com/objectcomputing/check-ins/issues/2534 to investigate replacing some of this mocking


server/build.gradle
```diff
@@ -29,6 +28,27 @@ micronaut {
...
+    binaries {
+        all {
+            imageName.set('check-ins')
+
+            // See https://github.com/micronaut-projects/micronaut-core/issues/10771
+            buildArgs.add("--initialize-at-build-time=kotlin.coroutines.intrinsics.CoroutineSingletons")
+
+            // When we are building on CI, we do not want to include the dev migrations.
+            def isCI = System.getenv("CI") != null
+            def migrationLocations = isCI ? "db/common" : "db/common,db/dev"
```
mkimberlin, line 42: What if we are building for the dev environment? We run the dev migrations there.

timyates: Good call...

I'm wondering whether we should pivot to simply use `dockerBuildNative` on CI to generate the docker image automatically for us...

I've pushed a commit which does this (and tests the image building works on CI), and also switched to check the `SERVICE_NAME` and var (and disable the dev db load when it exists and equals `checkins-master`)

timyates:
Regular app startup time is 17 seconds

`Startup completed in 17280ms. Server Running: http://localhost:8080"`

Native app startup time is 3.5 seconds

`Startup completed in 3547ms. Server Running: http://localhost:8080"`

I wonder what it's doing for those 3.5 seconds 🤔

timyates:
Tried AOT again.  In my unscientific test, startup time seemed to reduce from 3s to 2s when this commit was applied

https://github.com/objectcomputing/check-ins/pull/2533/commits/51094e53b974305baf161e825ad6ce5d107c7087

I've reverted it though, as I don't want to have too many things in the same PR, and this PR is about building a native image

timyates:
@mkimberlin I have aligned both the `gradle-deploy-develop.yml` and `gradle-deploy-native-develop.yml` actions

They currently both also trigger on pushes to the `'feature-2532/graal'` branch (this branch) for testing, but we probably want to get rid of that....

Should the deploy job depend on the test job passing?

We could just inline the `Run tests with gradle` step if so, but we may already be running enough testing before it gets to the develop branch 🤔 

I tried to pull commonality out into separate actions, and stitch them together as a github workflow-run, but I could not work out how to get the docker auth to work when it was run in a separate action, so I rolled it back and went with this...

It may be worth pulling the URL out into the `env` and referencing it with `${{ env.URL }}` as currently each action references it 3 times, and they all need to be the same

mkimberlin:
> @mkimberlin I have aligned both the `gradle-deploy-develop.yml` and `gradle-deploy-native-develop.yml` actions
> 
> They currently both also trigger on pushes to the `'feature-2532/graal'` branch (this branch) for testing, but we probably want to get rid of that....

Agreed.

> Should the deploy job depend on the test job passing?

Let's leave it as is for the moment. Broken code in dev doesn't worry me too much. Rollback is pretty easy, and this allows us to validate while the test suite is running.

> We could just inline the `Run tests with gradle` step if so, but we may already be running enough testing before it gets to the develop branch 🤔

It was inline, but I pulled it out because of how long the tests take.
 
> I tried to pull commonality out into separate actions, and stitch them together as a github workflow-run, but I could not work out how to get the docker auth to work when it was run in a separate action, so I rolled it back and went with this...

I went down a similar path at one point and just hit eject on that.

> It may be worth pulling the URL out into the `env` and referencing it with `${{ env.URL }}` as currently each action references it 3 times, and they all need to be the same

Agreed on that, too. I thought about that after we got off of stand-up earlier and then neglected to mention it in chat.

timyates:
I have removed the `feature-2532/graal` branch from the deployment actions, so they now only run on `develop` 👍

PR #2537, opened 2024-07-05
Switch to config object from Property for application config

timyates:
Following on from https://github.com/objectcomputing/check-ins/pull/2527 this moves the `check-ins.application.feedback` and `check-ins.application.google-api` configuration into an object instead of using `@Property` to grab the config by string based name.

It also flattens the config, as we used to have `notifications` for the `subject` and `content`, however the `content` property isn't used in the codebase, and having a parent for a single item seemed overkill


server/src/main/resources/application.yml
```diff
@@ -92,13 +92,11 @@ check-ins:
     google-api:
       delegated-user: ${ GSUITE_SUPER_ADMIN }
       scopes:
-        scopeForDriveApi: "https://www.googleapis.com/auth/drive.file"
-        scopeForDirectoryApi: "https://www.googleapis.com/auth/admin.directory.user.readonly"
+        scope-for-drive-api: "https://www.googleapis.com/auth/drive.file"
+        scope-for-directory-api: "https://www.googleapis.com/auth/admin.directory.user.readonly"
```
timyates, line 96: This is a no-op change, but kebab-case configuration is more idiomatic than camelCase

server/src/main/resources/application.yml
```diff
@@ -92,13 +92,11 @@ check-ins:
...
     feedback:
       max-suggestions: 6
-      notifications:
-        subject: "Feedback request"
-        content: "You have received a feedback request. Please go to the "
+      request-subject: "Feedback request"
```
timyates, line 99: `content` was unused so I removed it, and flattened the structure 1 level

server/src/test/resources/application-test.yml
```diff
@@ -38,6 +38,9 @@ flyway:
 ---
 check-ins:
   web-address: "https://checkins.objectcomputing.com"
+  application:
+    google-api:
+      delegated-user: test@objectcomputing.com
```
timyates, line 43: This was added as the property is defined as coming from an env var in application.yml and this would otherwise fail under test

mkimberlin (approved):

PR #2538, opened 2024-07-05
Bugfix sonar next round

timyates:
Few more sonar fixes here, again one by one for ease of reviewing:

1. [Mergeable "if" statements should be combined](https://github.com/objectcomputing/check-ins/commit/164ec68ea535a8ef6bf80fc610c73541cce331e8)
2. [Use `!collection.isEmpty()` instead of `collection.size() > 0` as it expresses intent better](https://github.com/objectcomputing/check-ins/commit/182f337c76984e36ced50d756e8a5114d755825e)
3. [Use new switch expressions instead of if statements for enum/sealed class comparison](https://github.com/objectcomputing/check-ins/commit/8320c213ee45457d1960f765cbe32cd2099d9d9a)
    Where there's a fixed number of possible outcomes, it's recommended to use switch expressions as any un-tested branches when a new value is added will throw a compilation error
4. [Instead of checking a Boolean class is null or false, just use `!Boolean.TRUE.equals(v)`](https://github.com/objectcomputing/check-ins/commit/4d2e96f79a51cd0685b32109d180c7608ae3b868)
5. Reduce cognitive load on two methods; [report](https://github.com/objectcomputing/check-ins/commit/19b26a0985080b8612c815ae8d10f58c4da5d398) and [getPotentiallyQualifyingMembers](https://github.com/objectcomputing/check-ins/commit/d79622be2b7ba1b2b32e74e08393a5893ea7ab64)
    Sometimes methods can end up too busy and become hard to follow in your head. Method extraction can help simplify these.
6. [Use `anyMatch` on stream instead of `filter.findFirst.isPresent`](https://github.com/objectcomputing/check-ins/pull/2538/commits/2a5076b4399dfc90f1a72ffe2652a5c4f5684099)


server/src/main/java/com/objectcomputing/checkins/services/memberprofile/retentionreport/RetentionReportServicesImpl.java
```diff
@@ -163,12 +137,10 @@ private RetentionReportResponseDTO getReportValues(LocalDate periodStartDate, Lo
 
     public float getTurnoverRate(LocalDate dateToStudy, List<MemberProfile> memberProfiles, FrequencyType frequency) {
         LocalDate beginningDate;
-        if (frequency == FrequencyType.DAILY) {
-            beginningDate = dateToStudy.minusDays(1);
-        } else if (frequency == FrequencyType.WEEKLY) {
-            beginningDate = dateToStudy.minusMonths(1);
-        } else {
-            beginningDate = dateToStudy.minusMonths(1);
```
timyates, line 171: WEEKLY is `dateToStudy.minusMonths(1)`

But so is the default (`MONTHLY`)... is this right?

mkimberlin: Yeah, not going to lie...I don't think this bit of logic is right. I think some wires got crossed on this bit of fuctionality. I need to go back and find the original task and create some bug fix stories. Luckily this bit doesn't have a UI yet.

mkimberlin (approved):

Issue #2534, opened 2024-07-03
Investigate reducing the usage of Mockito


timyates:
We have made good use of Mockito in the tests, however it doesn't play well with GraalVM native image compilation. As part of adding Graal support in https://github.com/objectcomputing/check-ins/pull/2533 we had to exclude 22 of these tests from the `nativeTest` execution.

We should investigate how much of this mocking can be replaced with standard classes, or simply using the actual components after the database has been set up in the way we expect for the tests

PR #2547, opened 2024-07-12
Replace Mockito mocked EmailSender in tests

timyates:
As part of investigating removing the mocking from the tests (for https://github.com/objectcomputing/check-ins/issues/2534), this switches from using a `mock(EmailSender)` to creating a bean under test which tracks the calls made to it in a simple list.

If the `replace.mailjet.factory` property is set in a test then the standard `MailJetFactory` is replaced by our factory.

This generates EmailSender instances (for HTML and TEXT) which track the calls to their methods in a list of strings.

We can also remove the `setEmailSender` methods which we had to use to force the mock bean into the running services.

I believe this has uncovered a bug in the Guild Service whereby too many emails are being sent. (see `com.objectcomputing.checkins.services.guild.GuildControllerTest#testEmailSentToGuildLeadWhenGuildMembersAdded`) I would have expected a single email, but we are sending 2... 🤔


server/src/test/java/com/objectcomputing/checkins/services/guild/GuildControllerTest.java
```diff
@@ -76,15 +92,21 @@ void testEmailSentToGuildLeadWhenGuildMembersAdded() {
...
+        assertEquals(2, emailSender.events.size());
+        assertEquals(
+                List.of(
+                        // Email to the Guild lead
+                        List.of("SEND_EMAIL", "null", "null", "Membership changes have been made to the Ninja guild", "<h3>Bill Charles has joined the Ninja guild.</h3><a href=\"https://checkins.objectcomputing.com/guilds\">Click here</a> to view the changes in the Check-Ins app.", memberProfile.getWorkEmail() + "," + memberProfile2.getWorkEmail()),
+                        // Email to both the Guild leads
+                        List.of("SEND_EMAIL", "null", "null", "Membership Changes have been made to the Ninja guild", "<h3>Changes have been made to the Ninja guild.</h3><h4>The following members have been added:</h4><ul><li>Bill Charles</li></ul><a href=\"https://checkins.objectcomputing.com/guilds\">Click here</a> to view the changes in the Check-Ins app.", memberProfile.getWorkEmail() + "," + memberProfile2.getWorkEmail())
+                ),
+                emailSender.events
         );
```
timyates, line 110: I believe this is a bug and we may be sending too many emails

server/src/test/java/com/objectcomputing/checkins/services/guild/GuildControllerTest.java
```diff
@@ -110,13 +132,15 @@ void testEmailSentToGuildLeadWhenGuildMembersRemoved() {
...
-                "<h3>Changes have been made to the Ninja guild.</h3><h4>The following members have been removed:</h4><ul><li>Bill Charles</li></ul><a href=\"https://checkins.objectcomputing.com/guilds\">Click here</a> to view the changes in the Check-Ins app.",
-                "billm@objectcomputing.com"
+        assertEquals(2, emailSender.events.size());
+        assertEquals(List.of(
+                        List.of("SEND_EMAIL", "null", "null", "Membership Changes have been made to the Ninja guild", "<h3>Bill Charles has left the Ninja guild.</h3><a href=\"https://checkins.objectcomputing.com/guilds\">Click here</a> to view the changes in the Check-Ins app.", memberProfile.getWorkEmail()),
+                        List.of("SEND_EMAIL", "null", "null", "Membership Changes have been made to the Ninja guild", "<h3>Changes have been made to the Ninja guild.</h3><h4>The following members have been removed:</h4><ul><li>Bill Charles</li></ul><a href=\"https://checkins.objectcomputing.com/guilds\">Click here</a> to view the changes in the Check-Ins app.", memberProfile.getWorkEmail())
+                ),
+                emailSender.events
         );
```
timyates, line 141: And here... I believe this is too many emails being sent 🤔

server/src/test/java/com/objectcomputing/checkins/services/guild/GuildControllerTest.java
timyates: Raised https://github.com/objectcomputing/check-ins/issues/2549

server/src/test/java/com/objectcomputing/checkins/services/guild/GuildControllerTest.java
timyates: Raised https://github.com/objectcomputing/check-ins/issues/2549

server/src/test/java/com/objectcomputing/checkins/services/guild/GuildControllerTest.java
timyates: potential PR to fix is in https://github.com/objectcomputing/check-ins/pull/2551

If that gets merged, we need to remove one of the Lists in this assert (and reduce the size assertion to 1)

server/src/test/java/com/objectcomputing/checkins/services/guild/GuildControllerTest.java
timyates: potential PR to fix is in https://github.com/objectcomputing/check-ins/pull/2551

If that gets merged, we need to remove one of the Lists in this assert (and reduce the size assertion to 1)

Issue #2553, opened 2024-07-26
Need notification for new Kudos


ocielliottc:
When a user submits a new kudos
  - If not public, send email to receiver (containing the kudos text)
  - If public, send email to any user in admin role (needs approval, link to approval page)
  - Update flow, kudos transitioning from no approval, to approval, send email to receiver of public kudos

ocielliottc:
For some reason, I cannot set my branch in "Development".

https://github.com/objectcomputing/check-ins/tree/feature-2553/kudos-notification

PR #2554, opened 2024-07-29
Feature 2553/kudos notification

ocielliottc:
Implementation for issue #2553.

My implementation for finding admin email addresses seems a bit too brute-force.  Maybe there is a more elegant way to do that?  I couldn't find anything better.


server/src/main/java/com/objectcomputing/checkins/services/kudos/KudosServicesImpl.java
```diff
@@ -269,4 +283,29 @@ private KudosResponseDTO constructKudosResponseDTO(Kudos kudos) {
 
         return kudosResponseDTO;
     }
+
+    void sendNotification(Kudos kudos) {
+        // Only public kudos really need approval, but just in case...
+        if (kudos.getPubliclyVisible()) {
```
mkimberlin, line 289: It looks like this is sending if publicly visible. There needs to be an additional check for publicly visible kudos that confirms that there is an approval date as well. We only want to send public kudos after they have been approved by an admin.

Additionally, if the kudos is not publicly visible, then the notification should be sent.  So, I think this if should really be:

```java
if (!kudos.getPubliclyVisible() || kudos.getDateApproved() != null) {
```

ocielliottc: This method is only called after the Kudos has been approved.  At this point, the Kudos is certain to have an approved date, as far as I can tell.

Private kudos are handled elsewhere.

mkimberlin (approved):

mkimberlin (approved):

PR #2570, opened 2024-08-12
Feature 2569/impersonate user

ocielliottc:
Adds the functionality to switch to a different user (given the correct permissions) and to switch back to the original user.
The UI change to allow this is within the drop-down button on the /admin/users page.

### DIFF
--- a/server/src/main/java/com/objectcomputing/checkins/security/ImpersonationController.java
+++ b/server/src/main/java/com/objectcomputing/checkins/security/ImpersonationController.java
@@ -1,0 +1,152 @@
+package com.objectcomputing.checkins.security;
+
+import com.objectcomputing.checkins.Environments;
+import com.objectcomputing.checkins.services.memberprofile.MemberProfile;
+import com.objectcomputing.checkins.services.memberprofile.currentuser.CurrentUserServices;
+import com.objectcomputing.checkins.services.permissions.Permission;
+import com.objectcomputing.checkins.services.permissions.RequiredPermission;
+import io.micronaut.http.MutableHttpResponse;
+import io.micronaut.http.HttpHeaders;
+import io.micronaut.http.cookie.SameSite;
+import io.micronaut.http.cookie.Cookie;
+import io.micronaut.http.netty.cookies.NettyCookie;
+import io.micronaut.security.utils.SecurityService;
+import io.micronaut.context.annotation.Requires;
+import io.micronaut.context.event.ApplicationEventPublisher;
+import io.micronaut.http.HttpRequest;
+import io.micronaut.http.HttpResponse;
+import io.micronaut.http.HttpStatus;
+import io.micronaut.http.MediaType;
+import io.micronaut.http.annotation.Consumes;
+import io.micronaut.http.annotation.Produces;
+import io.micronaut.http.annotation.Controller;
+import io.micronaut.http.annotation.Get;
+import io.micronaut.http.annotation.Post;
+import io.micronaut.scheduling.TaskExecutors;
+import io.micronaut.scheduling.annotation.ExecuteOn;
+import io.micronaut.security.annotation.Secured;
+import io.micronaut.security.authentication.Authentication;
+import io.micronaut.security.authentication.AuthenticationResponse;
+import io.micronaut.security.authentication.Authenticator;
+import io.micronaut.security.authentication.UsernamePasswordCredentials;
+import io.micronaut.security.event.LoginFailedEvent;
+import io.micronaut.security.event.LoginSuccessfulEvent;
+import io.micronaut.security.handlers.LoginHandler;
+import io.micronaut.security.rules.SecurityRule;
+import reactor.core.publisher.Flux;
+import reactor.core.publisher.Mono;
+
+import java.util.HashMap;
+import java.util.Locale;
+import java.util.Map;
+import java.util.Optional;
+import java.util.HashSet;
+import java.util.Set;
+import java.net.URI;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@Requires(env = Environments.LOCAL)
+@Controller("/impersonation")
+@ExecuteOn(TaskExecutors.BLOCKING)
+@Secured(SecurityRule.IS_AUTHENTICATED)
+public class ImpersonationController {
+    public static final String originalJWT = "OJWT";
+    private static final Logger LOG = LoggerFactory.getLogger(ImpersonationController.class);
+    protected final Authenticator authenticator;
+    protected final LoginHandler loginHandler;
+    protected final ApplicationEventPublisher eventPublisher;
+    private final CurrentUserServices currentUserServices;
+    private final SecurityService securityService;
+
+    /**
+     * @param authenticator       {@link Authenticator} collaborator
+     * @param loginHandler        A collaborator which helps to build HTTP response depending on success or failure.
+     * @param eventPublisher      The application event publisher
+     * @param currentUserServices Current User services
+     */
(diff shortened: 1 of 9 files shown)


server/src/main/java/com/objectcomputing/checkins/security/ImpersonationController.java
```diff
@@ -0,0 +1,151 @@
...
+import java.util.Map;
+import java.util.Optional;
+import java.util.HashSet;
+import java.util.Set;
+import java.net.URI;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@Requires(env = Environments.LOCAL)
```
mkimberlin, line 49: `LOCAL` or `DEV`, I think...

server/src/main/java/com/objectcomputing/checkins/security/ImpersonationController.java
```diff
@@ -0,0 +1,151 @@
...
+                                   CurrentUserServices currentUserServices,
+                                   SecurityService securityService) {
+        this.authenticator = authenticator;
+        this.loginHandler = loginHandler;
+        this.eventPublisher = eventPublisher;
+        this.currentUserServices = currentUserServices;
+        this.securityService = securityService;
+    }
+
+    @ExecuteOn(TaskExecutors.BLOCKING)
```
mkimberlin, line 82: You shouldn't need this here or on the `revert` endpoint. It's at the class level already.

server/src/main/java/com/objectcomputing/checkins/security/ImpersonationController.java
```diff
@@ -0,0 +1,151 @@
...
+    @Consumes({MediaType.APPLICATION_FORM_URLENCODED, MediaType.APPLICATION_JSON})
+    @Post("/begin")
+    @RequiredPermission(Permission.CAN_IMPERSONATE_MEMBERS)
+    public Mono<Object> auth(HttpRequest<?> request, String email) {
```
mkimberlin, line 86: You can get rid of the `Mono` usage here. It shouldn't be necessary any longer (there are only a few endpoints that use these still). I think you want an `HttpResponse<Void>` here for your return object instead...or maybe a `HttpResponse<?>`. Then you can use methods like `HttpResponse.serverError()` and `HttpResponse.ok()` to construct a response with the right response code. Those methods return `MutableHttpResponse` instances. So, you can do things like add cookies, etc. if needed.

At minimum it looks like you probably want to return `HttpResponse.serverError()` in the default cases toward the end.

server/src/main/java/com/objectcomputing/checkins/security/ImpersonationController.java
```diff
@@ -0,0 +1,151 @@
...
+                                       new NettyCookie(originalJWT, jwt.getValue()).path("/").sameSite(SameSite.Strict)
+                                       .maxAge(jwt.getMaxAge()));
+                        } else {
+                            eventPublisher.publishEvent(new LoginFailedEvent(authenticationResponse, null, null, Locale.getDefault()));
+                            return loginHandler.loginFailed(authenticationResponse, request);
+                        }
+                    }).single(Mono.just(HttpResponse.status(HttpStatus.UNAUTHORIZED)));
+                }
+            } else {
+                LOG.error("Attempted impersonation without authentication.");
```
mkimberlin, line 125: `HttpResponse.serverError()` here...

server/src/main/java/com/objectcomputing/checkins/security/ImpersonationController.java
```diff
@@ -0,0 +1,151 @@
...
+            }
+        }
+        return null;
```
mkimberlin, line 128: `HttpResponse.serverError()` here...

server/src/main/java/com/objectcomputing/checkins/security/ImpersonationController.java
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    @ExecuteOn(TaskExecutors.BLOCKING)
```
mkimberlin, line 131: Kill this `@ExecuteOn`

server/src/main/java/com/objectcomputing/checkins/security/ImpersonationController.java
```diff
@@ -0,0 +1,151 @@
...
+    @Produces(MediaType.TEXT_HTML)
+    @Get("/end")
+    public HttpResponse revert(HttpRequest<?> request) {
```
mkimberlin, line 134: I think `HttpResponse<Void>` is generally preferred in cases like this.

server/src/main/java/com/objectcomputing/checkins/security/ImpersonationController.java
```diff
@@ -0,0 +1,151 @@
...
+        final Cookie ojwt = request.getCookies().get(originalJWT);
+        if (ojwt == null) {
+            return HttpResponse.status(HttpStatus.UNAUTHORIZED);
```
mkimberlin, line 137: `HttpResponse.unauthorized()`

server/src/main/java/com/objectcomputing/checkins/security/ImpersonationController.java
ocielliottc: The `Authenticator.authenticate()` returns a Publisher.  I'm thinking Mono was left in the LocalLoginController because of this.  Or maybe I just don't know how to deal with the Publisher such that we can return an HttpResponse?

server/src/main/java/com/objectcomputing/checkins/security/ImpersonationController.java
ocielliottc: I found no reference to a DEV environment anywhere in the code.  Am I missing something?

mkimberlin: You are completely correct. It is specified on deployment though (see line 109):

https://github.com/objectcomputing/check-ins/blob/6d8cdb94c60b61f94f15456ebac2cd5c56887682/.github/workflows/gradle-build-development.yml#L109

You'll need to add it to `Environments` as `Environments.DEVELOP` or some such with a value of "dev"

server/src/main/java/com/objectcomputing/checkins/security/ImpersonationController.java
mkimberlin: I looked at this a little bit. Definitely not worth trying to change.

server/src/main/java/com/objectcomputing/checkins/security/ImpersonationController.java
mkimberlin: You can still do these though...just `return Mono.just(HttpResponse.serverError());`

server/src/main/java/com/objectcomputing/checkins/security/ImpersonationController.java
ocielliottc: I used `Environment.DEVELOPMENT` here from micronaut.  Is that going to do what's required?

server/src/main/java/com/objectcomputing/checkins/security/ImpersonationController.java
```diff
@@ -123,7 +124,7 @@ public Mono<Object> auth(HttpRequest<?> request, String email) {
                 LOG.error("Attempted impersonation without authentication.");
             }
         }
-        return null;
+        return Mono.just(HttpResponse.unauthorized());
```
mkimberlin, line 127: 👍

mkimberlin (approved):

PR #2603, opened 2024-09-24
Upgrades to Member Profile

thelenw:
Upgraded stylesheets for member profile and fixed various display issues including:

* Resolved "Organizations/My Orgs/Events" position
* Resolved button text getting cut off
* Resolved night-mode compatibility for inlaid tables
* Upgraded component and CSS for more efficient handling and faster UI


web-ui/src/components/profile/Profile.jsx
```diff
@@ -112,6 +112,60 @@ const Profile = ({ memberId, pdlId, checkinPdlId }) => {
...
+  const handleSaveNewOrganization = async () => {
+    const res = await resolve({
+      method: 'POST',
+      url: '/services/volunteer/organization',
+      headers: {
+        'X-CSRF-Header': csrf,
+        'Content-Type': 'application/json'
+      },
+      data: newOrganization
+    });
```
mkimberlin, line 124: Code like this we generally pull up into a file in the `web-ui/src/api` directory. I would put this and the one below for events in a file together called `volunteer.js`, if it were me.

web-ui/src/components/volunteer/Organizations.jsx
```diff
@@ -278,20 +282,21 @@ const Organizations = ({ onlyMe = false }) => {
...
+    } else {
       const index = organizations.findIndex(org => org.id === id);
       organizations[index] = newOrg;
-    } else {
-      organizations.push(newOrg);
     }
-    sortOrganizations(organizations);
-    setOrganizations(organizations);
-    setSelectedOrganization(null);
+    setOrganizations([...organizations]);
```
mkimberlin, line 299: Looks like this might be losing sorting and clearing of a selection.

web-ui/src/components/volunteer/VolunteerRelationships.jsx
```diff
@@ -161,6 +163,81 @@ const VolunteerRelationships = ({ forceUpdate = () => {}, onlyMe = false }) => {
...
+    
+    const res = await resolve({
+      method: 'POST',
+      url: organizationBaseUrl,
+      headers: {
+        'X-CSRF-Header': csrf,
+        'Content-Type': 'application/json'
+      },
+      data: newOrganization
+    });
```
mkimberlin, line 181: Pull this up to an API file as well...

web-ui/src/components/volunteer/VolunteerRelationships.jsx
```diff
@@ -428,6 +479,38 @@ const VolunteerRelationships = ({ forceUpdate = () => {}, onlyMe = false }) => {
...
+            value={newOrganization.website}
+            onChange={e => setNewOrganization({ ...newOrganization, website: e.target.value })}
+          />
+        </DialogContent>
+        <DialogActions>
+          <Button onClick={() => setOrganizationDialogOpen(false)}>Cancel</Button>
+          <Button onClick={handleCreateNewOrganization}>Save</Button>
+        </DialogActions>
+      </Dialog>
+
```
mkimberlin, line 513: This dialog exists twice. I wonder if it could be extracted for reuse in both places?

ocielliottc (commented):
1. Creating a volunteer organization requires a website URL (but it's not marked as required).
2. Cannot create a volunteer organization with the same name as a previously deleted organization.

ocielliottc:
The second point may be outside of the scope of this PR.

web-ui/src/components/volunteer/VolunteerTables.css
```diff
@@ -30,31 +47,46 @@
   td,
   th {
     padding: 0.5rem;
+    color: black; /* Ensures text color remains black */
```
mkimberlin, line 50: This is actually not the right color. This should be: `var(--checkins-palette-text-primary)`

web-ui/src/components/volunteer/VolunteerTables.css
```diff
@@ -30,31 +47,46 @@
...
+  }
+
   .warning {
     font-style: italic;
   }
 }
 
+
+.volunteer-tables .MuiIconButton-root {
+  color: black;
```
mkimberlin, line 81: Same here...should be: `var(--checkins-palette-text-primary)`

web-ui/src/components/volunteer/VolunteerTables.css
```diff
@@ -30,31 +47,46 @@
...
+}
+
+@media (prefers-color-scheme: dark) {
+  .volunteer-tables .MuiIconButton-root {
+    color: black;
```
mkimberlin, line 86: And here... `var(--checkins-palette-text-primary)`

web-ui/src/components/volunteer/VolunteerTables.css
ocielliottc: The problem with `var(--checkins-palette-text-primary)` in this case is that the table background is white in both light and dark mode.  And, in dark mode, this changes the text color to white.  Should the table background be something other than white when in dark mode?

mkimberlin:
@thelenw I fixed your yarn.lock issues.

mkimberlin:
@thelenw ok...tried again. Hopefully this one is in good shape.

web-ui/src/components/volunteer/VolunteerTables.css
mkimberlin: Is it not being altered appropriately by the theming?

mkimberlin: @thelenw Ok, I looked at this and yes...the table background is definitely not being altered appropriately. I would see what variables are being used for the other backgrounds and see if there are appropriate values available. My guess is that there is one for the primary background, but maybe not one for the alternating rows. You may need to find a color that works for that and define a new alternate row background variable for use by both themes.

mkimberlin: @thelenw I took a look at this, and I think I have a fix for you. I'll push shortly.

thelenw:
Thank you for yarn.lock fix. Seems to be resolved.

PR #2616, opened 2024-10-07
Label Feedback Responses

thelenw:
* Labled empty feedback responses
* UI adjustments


web-ui/src/components/view_feedback_responses/ViewFeedbackResponses.css
```diff
@@ -24,6 +24,17 @@
...
   padding: 10px;
+  display: flex;
+  align-items: center;
+}
+
+.caution-icon {
+  display: inline-block;
+  width: 16px;
+  height: 16px;
+  background-image: url('path-to-caution-icon.png'); /* Add the correct icon path */
```
mkimberlin, line 35: This looks suspect...

web-ui/src/components/view_feedback_responses/ViewFeedbackResponses.jsx
```diff
@@ -85,7 +86,21 @@ const ViewFeedbackResponses = () => {
           ? requests
           : [requests]
         : [];
-      return await getQuestionsAndAnswers(requests, cookie);
+      const res = await getQuestionsAndAnswers(requests, cookie);
+
+      if (res) {
+        console.log('Retrieved Questions and Answers:', res); // Debugging log
```
mkimberlin, line 92: as a rule, we try not to leave console.log calls in the codebase.

web-ui/src/components/view_feedback_responses/ViewFeedbackResponses.jsx
```diff
@@ -278,6 +284,41 @@ const ViewFeedbackResponses = () => {
...
+              >
+                Section 2: You can provide more detailed context to your review of this team member below. Be as specific as possible!
+              </Box>
+            );
+          }
+
+          // Handle other questions by removing "Q#: "
+          const questionText = question.question.replace(/^Q\d+: /, '');
+          const hasResponses = question.answers.length > 0;
+
```
mkimberlin, line 321: The above changes are problematic. They assume that the template used for reviews is static, but that is not the case.  Doing this creates an even tighter coupling to the data (template questions) than we already have. I would roll this bit back and stick to just addressing the items mentioned in the issue.

thelenw: Done.

web-ui/src/components/view_feedback_responses/ViewFeedbackResponses.jsx
thelenw: Existing console.log calls in pertinent files have been removed.

web-ui/src/components/view_feedback_responses/ViewFeedbackResponses.css
thelenw: Fixed.

thelenw:
* Implemented logic to display "No input" for questions with ```inputType: 'NONE'``` in ```components/view_feedback_responses/```

* For empty answers, system now displays "⚠️ No response submitted"

* Adjusted the filtering logic for search and responders to only display relevant answers while handling fallback cases properly.

* Improved logic for setting answers in ```FeedbackResponseCard``` to cover edge cases and non-string values.

mkimberlin (approved):

PR #2617, opened 2024-10-07
Feature 2596/send email mjml

ocielliottc:
1. Pulled in mjml4j to implement an MJML email sender.
2. Converted all emails sent by the Feedback Request Services to MJML (using provided templates).
3. Switched the Pulse Email over to MJML.
4. Updated tests.


mkimberlin:
Rather than hardcoding those MJML strings, why don't we load them as resources? Here's an example of what that could look like: https://github.com/ghillert/resource-readable-demo

thelenw (approved):
@mkimberlin reworked implementation of key MJML mechanism. @ocielliottc's original upgrades seem sound. Approved.

mkimberlin (approved):

thelenw (approved):

PR #2630, opened 2024-10-14
Send self-review emails at launch, 3 days to close, and 1 day to close.

ocielliottc:
One thing to note is that the fact that these emails are sent is tracked so we do not send duplicates, but this is not persisted.

### DIFF
--- a/server/src/main/java/com/objectcomputing/checkins/services/reviews/ReviewPeriodServicesImpl.java
+++ b/server/src/main/java/com/objectcomputing/checkins/services/reviews/ReviewPeriodServicesImpl.java
@@ -451,20 +451,21 @@ class ReviewPeriodServicesImpl implements ReviewPeriodServices {
                 }
             }
 
-            for(MemberProfile recipient : recipients) {
-                // Customize the email content using the template.
-                String content = String.format(template,
-                                     webAddress,
-                                     reviewPeriodId.toString(),
-                                     recipient.getFirstName(),
-                                     dateAsString(reviewPeriod.get()
-                                                    .getSelfReviewCloseDate()),
-                                     webAddress);
+            List<String> addresses = recipients.stream()
+                                         .map(p -> p.getWorkEmail()).toList();
 
-                // Send out the email to the individual.
-                emailSender.sendEmail(null, null, subject, content,
-                                      recipient.getWorkEmail());
-            }
+            // Customize the email content using the template.
+            String content = String.format(template,
+                                 webAddress,
+                                 reviewPeriodId.toString(),
+                                 dateAsString(reviewPeriod.get()
+                                                .getSelfReviewCloseDate()),
+                                 webAddress);
+
+            // Send out the email to everyone.
+            emailSender.sendEmail(null, null, subject, content,
+                                  recipients.toArray(
+                                      new String[recipients.size()]));
         } catch(Exception ex) {
             LOG.error("Send Self-Review Email: " + ex.toString());
         }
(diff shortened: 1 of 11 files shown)


server/src/main/java/com/objectcomputing/checkins/services/reviews/ReviewPeriodServicesImpl.java
```diff
@@ -39,20 +43,27 @@ class ReviewPeriodServicesImpl implements ReviewPeriodServices {
...
     private final MemberProfileRepository memberProfileRepository;
     private final FeedbackRequestServices feedbackRequestServices;
+    private final FeedbackRequestRepository feedbackRequestRepository;
     private final ReviewStatusTransitionValidator reviewStatusTransitionValidator;
     private EmailSender emailSender;
     private final Environment environment;
     private final String webAddress;
 
+    private enum SelfReviewDate { LAUNCH, THREE_DAYS, ONE_DAY }
+    private final Map<String, Boolean> sent = new HashMap<>();
```
mkimberlin, line 53: I'm worried about this. We have multiple instances. What happens if this gets called once on each instance? Or even more likely, if the instance shuts down between calls? We may need to persist a given reminder has been sent for a given review period.

server/src/main/java/com/objectcomputing/checkins/services/reviews/ReviewPeriodServicesImpl.java
```diff
@@ -346,4 +358,109 @@ private void createReviewRequest(ReviewPeriod period,
...
+                                     reviewPeriodId.toString(),
+                                     recipient.getFirstName(),
+                                     dateAsString(reviewPeriod.get()
+                                                    .getSelfReviewCloseDate()),
+                                     webAddress);
+
+                // Send out the email to the individual.
+                emailSender.sendEmail(null, null, subject, content,
+                                      recipient.getWorkEmail());
+            }
```
mkimberlin, line 461: I think we may run into an issue here eventually as well. There are limits to the number of emails we can send. So, if we have an email template that does not require customization for a specific recipient, we should send it to all of them simultaneously (the sender batches them to 49 recipients per email).

mkimberlin: Oops...my bad. This one does require customization per recipient.

server/src/main/java/com/objectcomputing/checkins/services/reviews/ReviewPeriodServicesImpl.java
ocielliottc: Yes.  I thought this could be a problem.

server/src/main/java/com/objectcomputing/checkins/services/reviews/ReviewPeriodServicesImpl.java
ocielliottc: @mkimberlin If we take out the by-name-greeting, we could send them out in bulk.

server/src/main/java/com/objectcomputing/checkins/services/reviews/ReviewPeriodServicesImpl.java
mkimberlin: @ocielliottc yeah, I think that's a good call.

server/src/main/java/com/objectcomputing/checkins/services/reviews/ReviewPeriodServicesImpl.java
ocielliottc: @mkimberlin No.  The link just has the review period id in it.

mkimberlin (approved):

Issue #2636, opened 2024-10-16
Reviews - Removing & Adding User bug - removed user isn't removed


Luch76:
Have a Review with one user. Goto list to add/remove members. Add a new member, and remove the original user that was there. You'll then see the user that was removed is still there. 

Note the bug doesn't happen when you have multiple users, and simply remove one user

ocielliottc:
So, what happens is that when you remove and add in one shot, it only locally removes the reviewers and does a POST to add the new ones.  What you get back is the full list of reviewers from the server.  This will just require me to add some code to DELETE the reviewers.

PR #2639, opened 2024-10-16
#2636 - Avoid overwriting the list of team members, if we already have a list.

ocielliottc:


Luch76 (approved):

Luch76:
Bug is fixed - thanks!

mkimberlin (approved):

PR #2647, opened 2024-10-18
Volunteer Tracking Bugfixes

thelenw:
**Update**
Allow and properly reflect empty _Start Date_ field when User creates a relationship with an organization thus.

**Details**
- Changes to frontend @ ```src/components/volunteer``` to support Users creating & editing volunteer organization entries with no start date.

- Changes to backend @ ```main/java/com/objectcomputing/checkins/services/volunteering``` to support nullable _Start Date_.

- Changes to migration 107 to allow nullable _Start Date_ in the database.


server/src/main/resources/db/common/V107__create_volunteering_tables.sql
```diff
@@ -16,7 +16,7 @@ CREATE TABLE volunteering_relationship
     relationship_id varchar PRIMARY KEY,
     member_id       varchar REFERENCES member_profile (id),
     organization_id varchar REFERENCES volunteering_organization (organization_id),
-    start_date      timestamp NOT NULL,
```
mkimberlin, line 19: You cannot edit existing migrations. You have to create a new migration (with the next version number) and alter the table instead of changing this one.

thelenw: This will be fixed momentarily.

thelenw (commented):
Make ```Start Date``` nullable:
- [x] Allow empty start_date field
- [x] Make the field nullable for all server endpoints
- [ ] Update UI to reflect
- [ ] Fix tests

server/src/main/resources/db/common/V115__alter_volunteering_relationship.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE volunteering_relationship
```
mkimberlin, line 1: Need to rename this file to increment to 116 before merging. 115 has now been taken in develop

thelenw: Done.

mkimberlin (approved):

PR #2683, opened 2024-10-29
Replaced the skeleton loader with a message that no guilds are available

ocielliottc:


mkimberlin:
@ocielliottc I see where you are coming from on this...but I don't think removing the skeleton loader completely is the way I would address this. I think we should show the skeleton loader when the data is still loading, and display the message only if the response comes back with no guilds.

ocielliottc:
> @ocielliottc I see where you are coming from on this...but I don't think removing the skeleton loader completely is the way I would address this. I think we should show the skeleton loader when the data is still loading, and display the message only if the response comes back with no guilds.

I initially thought that the guilds were requested through an API.  But, they're actually passed into the component.

ocielliottc:
> > @ocielliottc I see where you are coming from on this...but I don't think removing the skeleton loader completely is the way I would address this. I think we should show the skeleton loader when the data is still loading, and display the message only if the response comes back with no guilds.
> 
> I initially thought that the guilds were requested through an API. But, they're actually passed into the component.

Hmm...that's not right.  I'll take a closer look.

ocielliottc:
> > > @ocielliottc I see where you are coming from on this...but I don't think removing the skeleton loader completely is the way I would address this. I think we should show the skeleton loader when the data is still loading, and display the message only if the response comes back with no guilds.
> > 
> > 
> > I initially thought that the guilds were requested through an API. But, they're actually passed into the component.
> 
> Hmm...that's not right. I'll take a closer look.

Ok.  The guilds are actually loaded by the AppContext.  So, I think they're loaded before the guilds page is even loaded.

PR #2723, opened 2024-11-05
Bugfix 2720/homepage crammed

mkimberlin:


mkimberlin:
As is, this fix is rough at a few screen sizes. I may play with it a bit. I'm not sure grid is the approach I'd go for over flex, but if we can make it work, It doesn't really matter to me.

PR #2725, opened 2024-11-05
Changed app context logic to get CSRF value and reuse it if available

mkimberlin:


web-ui/src/context/AppContext.jsx
```diff
@@ -63,14 +74,17 @@ const AppContextProvider = props => {
...
+      if (!csrf && !cookieVal) {
         const res = await fetch(url, {
           responseType: 'text',
           credentials: 'include'
         });
         if (res && res.ok) {
           dispatch({ type: SET_CSRF, payload: await res.text() });
         }
+      } else if (cookieVal) {
+        dispatch({ type: SET_CSRF, payload: cookieVal });
```
ocielliottc, line 87: If you checked for cookieVal before !csrf, could you remove !cookieVal from that if?

```
if (cookieVal) {
} else if (!csrf) {
}
```

ocielliottc (approved):

PR #2770, opened 2024-11-14
Restrict feedback to within the review period

mkimberlin:


server/src/main/java/com/objectcomputing/checkins/services/reports/ReportDataCollation.java
```diff
@@ -260,7 +260,9 @@ private List<Feedback> getFeedbackType(FeedbackType type) {
                         reviewPeriod.getReviewTemplateId());
                 break;
               case FeedbackType.feedback:
-                use = !template.getIsReview();
+                use = !template.getIsReview() && request.getStatus().equalsIgnoreCase("submitted") &&
+                      request.getSendDate() != null && request.getSendDate().isAfter(reviewPeriod.getPeriodStartDate().toLocalDate()) &&
+                      request.getSubmitDate() != null && request.getSubmitDate().isBefore(reviewPeriod.getCloseDate().toLocalDate());
```
ocielliottc, line 265: What happens if someone misses the close date for their feedback?

ocielliottc (approved):

server/src/main/java/com/objectcomputing/checkins/services/reports/ReportDataCollation.java
mkimberlin: It won't be included in the report...but fair point. I will pull that requirement.

PR #2776, opened 2024-11-15
Send an approved public kudos to the #kudos slack channel.

ocielliottc:
This requires two environment variables:

1. SLACK_WEBHOOK_URL - which is the URL for the Slack installed application that is available when the integration app is installed to the kudos channel (already done).  This is on the Installed App Settings tab.
2. SLACK_BOT_TOKEN - this is available on the same tab.  This part allows the application to search for a channel id and user ids.

https://api.slack.com/apps/A081ESV0D5W/install-on-team

The application has already been installed to the #kudos channel.  We just need to deploy this and provide the above environment variables.

This is what the test channel looks like:

image

### DIFF
--- a/server/src/main/java/com/objectcomputing/checkins/notifications/social_media/SlackSearch.java
+++ b/server/src/main/java/com/objectcomputing/checkins/notifications/social_media/SlackSearch.java
@@ -1,5 +1,6 @@
 package com.objectcomputing.checkins.notifications.social_media;
 
+import com.objectcomputing.checkins.configuration.CheckInsConfiguration;
 import com.slack.api.model.block.LayoutBlock;
 import com.slack.api.Slack;
 import com.slack.api.methods.MethodsClient;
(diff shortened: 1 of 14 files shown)


ocielliottc:
I'm looking into adding a test for the SlackPoster...

server/src/main/java/com/objectcomputing/checkins/notifications/social_media/SlackSearch.java
```diff
@@ -0,0 +1,78 @@
...
+                LOG.error("SlackSearch.findChannelId: " + e.toString());
+            } catch(SlackApiException e) {
+                LOG.error("SlackSearch.findChannelId: " + e.toString());
+            }
+        }
+        return null;
+    }
+
+    public String findUserId(String userEmail) {
+        String token = System.getenv(env);
```
mkimberlin, line 58: I missed this...fixing.

PR #2783, opened 2024-11-21
Adjust redirect logic for feedback requests

mkimberlin:


Luch76 (approved):
Yep, that fixes it, thanks!

PR #2792, opened 2024-12-17
Feature 2759/server side markdown generation

ocielliottc:


server/src/test/java/com/objectcomputing/checkins/services/reports/FileServicesImplReplacement.java
```diff
@@ -0,0 +1,59 @@
...
+import java.util.UUID;
+
+import jakarta.inject.Singleton;
+import io.micronaut.context.env.Environment;
+import io.micronaut.context.annotation.Replaces;
+import io.micronaut.context.annotation.Requires;
+
+@Singleton
+@Replaces(FileServicesImpl.class)
+@Requires(env = Environment.TEST)
```
mkimberlin, line 32: That's the magic!

ocielliottc: Yep!  It came to me one morning to do it like we do the email sender!

PR #2797, opened 2024-12-20
Allow for anonymous pulse submission and viewing in the pulse report.

ocielliottc:


web-ui/src/pages/PulsePage.jsx
```diff
@@ -97,22 +97,18 @@ const PulsePage = () => {
...
     if (res.error) return;
 
-    // Refresh browser to show that pulses where already submitted today.
-    history.go(0);
+    if (submitAnonymously) {
+      setSubmittedToday(true);
+      Cookies.set(cookieName, 'true', { expires: 1 });
+    } else {
+      // Refresh browser to show that pulses where already submitted today.
+      history.go(0);
```
mkimberlin, line 110: I know this wasn't you, but this is odd to me. Why are we refreshing the whole page to trigger a state change, I wonder... I blame @mvolkmann. I think he forgot how to write React code. 🤣

mkimberlin: Now that I think about this. Couldn't this whole if/else be turned into just lines 106-107?

web-ui/src/pages/PulsePage.jsx
```diff
@@ -141,9 +137,25 @@ const PulsePage = () => {
...
+            <Button
+              style={{ marginTop: 0 }}
+              onClick={submit}
+              variant="contained">
+              Submit
+            </Button>
+            <div style={{ padding: '.3rem' }}/>
+            <label>
+              <Checkbox
+                disableRipple
```
mkimberlin, line 150: The ripple haunts my dreams now.

web-ui/src/pages/PulsePage.jsx
ocielliottc: Yes.  I think we can.  Currently, the cookie is only used when anonymous.  But, we could do it generically too.  Also, the refresh was weird to me too.

server/src/main/resources/db/common/V119__alter_pulse_response_team_id.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE pulse_response DROP CONSTRAINT pulse_response_teamMemberId_fkey;
```
mkimberlin, line 1: It seems odd to me that we would need to drop a foreign key reference to allow null. Is this in some way related to: https://stackoverflow.com/questions/23325838/postgresql-null-value-in-foreign-key-column ?

ocielliottc: @mkimberlin Honestly, I just assumed that because it was referencing the member_profile, that it could not be null.  I did not test without this.

ocielliottc: I removed that file, it was not necessary and took out the history.go(0) call too.

web-ui/src/pages/PulsePage.jsx
mvolkmann: `history.go(0)` is the tool for when you can't find another way to get it to work and you are getting tired. ;-)

mkimberlin: 🤣🤣🤣

mkimberlin (approved):

PR #2810, opened 2025-01-08
Feature 2805/convert skills data structure

ocielliottc:


server/src/main/resources/db/common/V119__alter_member_skills_table.sql
```diff
@@ -0,0 +1,11 @@
+ALTER TABLE member_skills ADD COLUMN interested boolean DEFAULT false;
```
mkimberlin, line 1: Do you think we should default to null maybe? This could be misconstrued later as folks explicitly saying they were not interested.

ocielliottc: Yeah, we can just remove "DEFAULT false".  That's a good idea.

PR #2820, opened 2025-01-13
Feature 2814/slack pulse survey integration

ocielliottc:
This needs to be deployed to an external server to continue development.

There is an extraneous change on this branch where I renamed `sendPendingEmail` method to `notifyUsers` because I originally thought that we were going to push the Pulse form out to the users.  It is now up to the user to run the slash command: `/pulse` to initiate the Slack pulse flow.


web-ui/src/pages/PulseReportPage.jsx
```diff
@@ -328,7 +328,13 @@ const PulseReportPage = () => {
     });
     if (res.error) return;
 
-    const pulses = res.payload.data;
+    // Get the pulses and filter out invalid data.
+    const pulses = res.payload.data.filter((pulse) => {
+      return pulse.internalScore > 0 && pulse.internalScore <= 5 &&
+             (pulse.externalScore == null ||
+              (pulse.externalScore > 0 && pulse.externalScore <= 5));
+    });
```
mkimberlin, line 336: Do we have bad data somewhere?

ocielliottc: There is nothing that restricts externalScore from being any integer value.  When I was testing, just for fun, I decided to send values outside of that range and it did not deal well.

PR #2828, opened 2025-01-20
Feature 2825/adjust admin menu

ocielliottc:


server/src/main/java/com/objectcomputing/checkins/services/kudos/KudosServicesImpl.java
```diff
@@ -121,10 +122,6 @@ public Kudos save(KudosCreateDTO kudosDTO) {
 
     @Override
     public Kudos approve(Kudos kudos) {
-        if (!currentUserServices.isAdmin()) {
-            throw new PermissionException(NOT_AUTHORIZED_MSG);
-        }
-
         UUID kudosId = kudos.getId();
```
mkimberlin, line 125: Do we not need to replace this with a permission check? I don't see it happening elsewhere in here, but I could be missing it.

server/src/main/java/com/objectcomputing/checkins/services/kudos/KudosServicesImpl.java
```diff
@@ -173,10 +170,6 @@ public KudosResponseDTO getById(UUID id) {
 
     @Override
     public void delete(UUID id) {
-        if (!currentUserServices.isAdmin()) {
-            throw new PermissionException(NOT_AUTHORIZED_MSG);
-        }
-
         Kudos kudos = kudosRepository.findById(id).orElseThrow(() ->
```
mkimberlin, line 173: Same above here. Is there a permission check happening?

server/src/main/java/com/objectcomputing/checkins/services/memberprofile/MemberProfileServicesImpl.java
```diff
@@ -237,6 +217,47 @@ public List<MemberProfile> getSubordinatesForId(UUID id) {
     @Override
     @CacheInvalidate(cacheNames = {"member-cache"})
     public MemberProfile updateProfile(MemberProfile memberProfile) {
+        if (memberProfile.getId() == null) {
+            throw new BadArgException("Null profile id in update");
+        }
+
+        MemberProfile currentUser = currentUserServices.getCurrentUser();
+        boolean isAdmin = currentUserServices.isAdmin();
+        if (!isAdmin && (currentUser == null || !currentUser.getId().equals(memberProfile.getId()))) {
```
mkimberlin, line 226: Was this an intentional or accidental can kicking? 🤣

server/src/main/java/com/objectcomputing/checkins/services/memberprofile/MemberProfileServicesImpl.java
```diff
@@ -237,6 +217,47 @@ public List<MemberProfile> getSubordinatesForId(UUID id) {
...
+                emailAssignment(updatedMemberProfile, false); // Supervisor
+            }
+        }
+
+        return updatedMemberProfile;
+    }
+
+    @Override
+    @CacheInvalidate(cacheNames = {"member-cache"})
+    public MemberProfile updateCurrentUserProfile(MemberProfile memberProfile) {
```
mkimberlin, line 260: This seems to be misleading. This isn't related to the "current user" at all, is it? Am I missing something?

server/src/main/java/com/objectcomputing/checkins/services/kudos/KudosServicesImpl.java
ocielliottc: The controller handles this:

```
    @Put
    @RequiredPermission(Permission.CAN_ADMINISTER_KUDOS)
    public Kudos approve(@Body @Valid Kudos kudos) {
        return kudosServices.approve(kudos);
    }
```

server/src/main/java/com/objectcomputing/checkins/services/kudos/KudosServicesImpl.java
ocielliottc: Same with the delete:

```
    @Delete("/{id}")
    @Status(HttpStatus.NO_CONTENT)
    @RequiredPermission(Permission.CAN_ADMINISTER_KUDOS)
    public void delete(@NotNull UUID id) {
        kudosServices.delete(id);
    }
```

server/src/main/java/com/objectcomputing/checkins/services/memberprofile/MemberProfileServicesImpl.java
ocielliottc: Well, it is only called from the current user controller when updating the "last seen" for the current user.

server/src/main/java/com/objectcomputing/checkins/services/memberprofile/MemberProfileServicesImpl.java
ocielliottc: Well, intentional in that there are no required permissions to update a member profile.  Since every member can update their own profile, I didn't see a way to limit which profile could be updated except by using the existing mechanism of checking for admin.  But, as I'm writing this, I'm thinking that we could add a permission to be able to "update all profiles" and specifically check for it here instead of admin.  Thoughts?

ocielliottc: Yes, I think that would work!

server/src/main/java/com/objectcomputing/checkins/services/memberprofile/MemberProfileServicesImpl.java
ocielliottc: Maybe it should be renamed to something like "directProfileUpdate" or something indicating that there is no security related to it?

server/src/main/java/com/objectcomputing/checkins/services/memberprofile/MemberProfileServicesImpl.java
mkimberlin: 👍 I'm onboard with that.

server/src/main/java/com/objectcomputing/checkins/services/memberprofile/MemberProfileServicesImpl.java
mkimberlin: Maybe we isolate that code to update the "last seen" here and call it something like 'updateLastSeen' and just pass in the user id? that feels less risky than an unsecured profile update API.

mkimberlin (approved):

PR #2851, opened 2025-01-29
Intercept 401 errors, pop up a login window that closes when successful.

ocielliottc:


web-ui/src/api/api.js
```diff
@@ -58,6 +58,24 @@ export const getMyFetch = async () => {
...
+          clearInterval(interval);
+          resolve();
+        }
+      } catch (err) {
+        reject(err);
+      }
+    }, 1000);
+  });
+}
+
```
mkimberlin, line 78: I think this is only workable for localhost, isn't it?

ocielliottc: I think it works elsewhere because the URL doesn't really matter (except for our route match to close the window).  In production, it will see that the JWT has expired (or doesn't exist) and will make the user pick their google account to use.

mkimberlin:
I've never been so excited to see if something works. 🤣

PR #2876, opened 2025-02-11
Feature 2868/pull slack kudos in

ocielliottc:
This adds a scheduled task to read slack messages from a configured channel and import user messages as Check-Ins Kudos upon verification with the poster of the slack message.

1. Moved slack related functionality into a separate `slack` service directory.
2. Added the ability to save pre-approved kudos.
3. Modified `CurrentUserServicesImpl.hasPermission` to return false if there is no current user.
4. Moved Slack related functionality out of the `PulseResponseController`
  - **At some point, we will probably want to change the URI for Slack related traffic.**
5. The configuration requires a `SLACK_KUDOS_CHANNEL_ID` value.
  - For testing, I was using `C0814P2PEV7`

This will need to be deployed to finish testing.

Also, once this is deployed to production, we will need to add the following scopes to the Check-Ins Integration.
- `channels:history`
- `chat:write`
- `groups:history`
- `im:write`

### DIFF
--- a/server/src/main/java/com/objectcomputing/checkins/services/slack/kudos/KudosChannelReader.java
+++ b/server/src/main/java/com/objectcomputing/checkins/services/slack/kudos/KudosChannelReader.java
@@ -52,6 +52,8 @@ public class KudosChannelReader {
             kudosChannelReadTimeStore.save(new KudosChannelReadTime());
         }
 
-        slackKudosCreator.store(messages);
+        if (!messages.isEmpty()) {
+            slackKudosCreator.store(messages);
+        }
     }
 }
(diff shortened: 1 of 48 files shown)


server/src/main/java/com/objectcomputing/checkins/services/slack/kudos/KudosChannelReader.java
```diff
@@ -0,0 +1,57 @@
...
+    @Inject
+    private CheckInsConfiguration configuration;
+
+    @Inject
+    private SlackReader slackReader;
+
+    @Inject
+    private SlackKudosCreator slackKudosCreator;
+
+    @Scheduled(fixedDelay = "10m")
```
mkimberlin, line 36: I don't think we can count on this because sometimes we scale down completely...and we will end up with multiple channel readers when the app is busy as well. We can end up with up to 10 instances running currently, I think.

ocielliottc: @mkimberlin I just checked in a change to remove the scheduled task and tie it to the `CheckServicesController`.

mkimberlin (approved):

PR #2878, opened 2025-02-11
Bugfix 2877/private kudos handling

ocielliottc:
Immediately approve private kudos, as they do not need to go through the approval process.


ocielliottc:
This fix is incorrect.

PR #2881, opened 2025-02-12
Feature 2870/edit kudos

ocielliottc:
This implements editing public/private, message text, and recipients and the ability for a user to delete kudos they have sent.

### DIFF
--- a/server/src/main/java/com/objectcomputing/checkins/services/kudos/KudosServicesImpl.java
+++ b/server/src/main/java/com/objectcomputing/checkins/services/kudos/KudosServicesImpl.java
@@ -150,7 +150,7 @@ class KudosServicesImpl implements KudosServices {
     }
 
     @Override
-    public Kudos update(KudosResponseDTO kudos) {
+    public Kudos update(KudosUpdateDTO kudos) {
         // Find the corresponding kudos and make sure we have permission.
         final UUID kudosId = kudos.getId();
         final Kudos existingKudos =
@@ -171,6 +171,7 @@ class KudosServicesImpl implements KudosServices {
         if (existingPublic && !proposedPublic) {
             // TODO: Somehow find and remove the Slack Kudos that the Check-Ins
             //       Integration posted.
+            existingKudos.setDateApproved(LocalDate.now());
         } else if (!existingPublic && proposedPublic) {
             // Clear the date approved when going from private to public.
             existingKudos.setDateApproved(null);
--- a/server/src/main/java/com/objectcomputing/checkins/services/kudos/KudosServicesImpl.java
+++ b/server/src/main/java/com/objectcomputing/checkins/services/kudos/KudosServicesImpl.java
@@ -170,6 +170,7 @@ class KudosServicesImpl implements KudosServices {
         }
 
         // Begin modifying the existing kudos to reflect desired changes.
+        final String originalMessage = existingKudos.getMessage();
         existingKudos.setMessage(kudos.getMessage());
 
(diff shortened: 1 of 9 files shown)


server/src/main/java/com/objectcomputing/checkins/services/kudos/KudosServicesImpl.java
```diff
@@ -118,6 +121,12 @@ public Kudos save(KudosCreateDTO kudosDTO) {
             kudosRecipientServices.save(kudosRecipient);
         }
 
+        if (!savedKudos.getPubliclyVisible()) {
+            // Private kudos do not need to be approved by another party.
+            savedKudos.setDateApproved(LocalDate.now());
```
mkimberlin, line 126: Previously we didn't require any approved date for non-public kudos, so we didn't set one. I'm guessing something has shifted somewhere in the logic such that we need this date now? If so, then we might need to make a migration to add these dates to existing private kudos.

mkimberlin:
@ocielliottc I think we may also want to "unapprove" (remove the approval date) a public kudos that has the text edited.

server/src/main/java/com/objectcomputing/checkins/services/kudos/KudosServicesImpl.java
ocielliottc: I will have to look to see if anything changed anywhere to see why private kudos weren't being displayed.

web-ui/src/components/kudos_card/KudosCard.jsx
```diff
@@ -208,7 +285,7 @@ const KudosCard = ({ kudos, includeActions, onKudosAction }) => {
         <DialogTitle>Delete Kudos</DialogTitle>
         <DialogContent>
           <DialogContentText>
-            Are you sure you want to deny approval for these kudos? The kudos
+            Are you sure you want to complete this action? The kudos
             will be deleted.
           </DialogContentText>
           <TextField
```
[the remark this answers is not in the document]
ocielliottc: I'm all for removing the text field.

server/src/main/java/com/objectcomputing/checkins/services/kudos/KudosServicesImpl.java
ocielliottc: I found the root of the problem.

```
    private List<KudosResponseDTO> findAllToMember(UUID memberId) {
        UUID currentUserId = currentUserServices.getCurrentUser().getId();

        if (!currentUserId.equals(memberId) &&
            !hasAdministerKudosPermission()) {
            throw new PermissionException("You are not authorized to retrieve the kudos another user has received");
        }

        List<KudosRecipient> kudosRecipients = kudosRecipientRepository.findByMemberId(memberId);

        List<KudosResponseDTO> kudosList = new ArrayList<>();

        kudosRecipients.forEach(kudosRecipient -> {
            UUID kudosId = kudosRecipient.getKudosId();
            Kudos relatedKudos = kudosRepository.findById(kudosId).orElseThrow(() ->
                    new NotFoundException(KUDOS_DOES_NOT_EXIST_MSG.formatted(kudosId)));

            if (relatedKudos.getDateApproved() != null) {
                kudosList.add(constructKudosResponseDTO(relatedKudos));
            }
        });

        return kudosList;
    }
```

When we retrieve all of the received kudos, we are requiring an approved date.  According to git, it's been like this for 3 years.  I will undo my date changes and fix this method (and some display stuff too).

mkimberlin (approved):

